PR #151, opened 2014-08-16
[WIP] Refactor HTTP status codes

ixti:
* [ ] Default well-known status codes should be auto-generated
* ~~Rename `HTTP::Response::Status` -> `HTTP::Status`~~
* [ ] Replace `Status::*` consts with registry singleton that will allow add/override own status codes
* [ ] Make `HTTP::Status` an ordinary class, coercible to Fixnums, Symbols, Strings and so on


ixti:
Implemented status auto-generation, other tasks decided to avoid at this time.
Gonna come with a stronger PR before 1.0.0, but this one should not hold us back anymore.

PR #173, opened 2015-01-03
Fail upon invalid HTTP header field name

ixti:
Disallows to pass invalid header field names like:

- `"Foo Bar"` (using spaces)
- `"  Foo-Bar"` (leading and/or tailing spaces)
- `"Foo:Bar"` (using separator field)

See http://tools.ietf.org/html/rfc7230#section-3.2 and specifically `field-name` syntax for details.

### DIFF
--- a/lib/http/errors.rb
+++ b/lib/http/errors.rb
@@ -8,6 +8,9 @@ module HTTP
   # Generic Response error
   class ResponseError < Error; end
 
-  # Request to do something when we're in the wrong state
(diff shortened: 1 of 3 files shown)


ixti:
Do not merge yet.

tarcieri:
One note on this: There is the traditional wisdom of "be liberal in what you accept and strict in what you send".

That said, I'm not sure that "wisdom" makes sense if we care about security. Maybe we should start favoring correctness.

PR #176, opened 2015-01-18
Use at_exit instead of overriding trap for interrupt to shutdown ProxyServer, AuthenticatedProxyServer

olegkovalenko:
Call to trap replaces handler for specified signal,
so ProxyServer.shutdown would have never been called.
at_exit registers handler for execution when the programm
exits, handlers are executed in revers order of registration.

### DIFF
--- a/spec/lib/http_spec.rb
+++ b/spec/lib/http_spec.rb
@@ -41,35 +41,49 @@ RSpec.describe HTTP do
     end
   end
 
-  context "with http proxy address and port" do
-    it "proxies the request" do
-      response = HTTP.via("127.0.0.1", 8080).get dummy.endpoint
-      expect(response.headers["X-Proxied"]).to eq "true"
-    end
-  end
+  context "with http proxy" do
+    let(:handler) { proc { |_, res| res["X-PROXIED"] = true } }
 
-  context "with http proxy address, port username and password" do
-    it "proxies the request" do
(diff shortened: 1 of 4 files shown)


tarcieri:
Hmm, I think the best solution here is to factor ProxyServer startup/shutdown into before/after blocks...

ixti:
Although this patch really makes things a bit better, it doesn't fixes same issue with example_server, and I agree with @tarcieri it's better to start/stop it with before/after blocks.

ixti:
I have re-factored how we start `ExampleServer` (also renamed it to `DummyServer`) in specs.
You can check new implementation and kind of follow same way if you want, or I will jump on this later.

ixti:
Ouch. My last commit broke your last commit :D

Please, don't put before blocks into `spec_helper`.
Instead add `RSpec.configure do |config|` to the end of `proxy_server` with your code.
And remove `ExampleServer` start out of there - I have completely refactored it.

olegkovalenko:
No worries ;)

tarcieri:
Typically I start and stop these sorts of things in conjunction with the tests that depend on them, rather than globally in the suite.

ixti:
Thanks for your PR, but I have implemented it in a bit different way :D
In any case thank you very much.

olegkovalenko:
(handshake)
Thanks you for time and suggestions ;)

Issue #77, opened 2014-02-03
Caching support (was: Conditional Gets)


Asmod4n:
Would be nice to have some form of Response caching here, with a simple Hash Store and Memcache API Store, seeing how Github itself doesn't count successfull Conditional Gets towards the API limit.

Asmod4n:
Could someone else take https://github.com/Asmod4n/http/compare/caching over? I am afraid i don't use http.rb anymore and haven't found fixes for some pitfalls for caching (attackers can allocate any amount of ram they want with certain conditions).

ixti:
@Asmod4n I don't think I will be able to jump anytime soon on this issue, but I would like to keep it open if anybody would like to jump on. I have merged your `caching` branch of your repo into `wip/caching-support` on this repo.

Asmod4n:
@ixti thanks a lot, i will try to take a look at it after i have read the new HTTP rfcs regarding caching.

ixti:
Closed by #177 which is in master and was released as 0.8.0.pre

sferik:
@ixti This issue should also be moved into the 0.8.0 milestone. Are you able to do that? I’m still getting an error.

ixti:
@sferik  Same here.

PR #177, opened 2015-02-03
Caching

pezra:
Implementation of HTTP caching using [rake-cache](http://rtomayko.github.io/rack-cache/)'s persistance layer. Based on @Asmod4n's initial pass (<https://github.com/Asmod4n/http/compare/caching>) but much has changed. :) Addresses issue #77.


ixti:
Huge! Some cleanups and refactorings and would be good to go. :D

Rakefile
```diff
@@ -23,7 +23,7 @@ end
 require "yardstick/rake/verify"
 Yardstick::Rake::Verify.new do |verify|
   verify.require_exact_threshold = false
-  verify.threshold = 58
+  verify.threshold = 58.1
```
tarcieri: does this not support `>= 50` or something like that?

tarcieri:
Yeah, I'll second huge... looks interesting though.

Rakefile
ixti: It is. And `verify.threshold = 58.1` in conjunction with `verify.require_exact_threshold = false` (above) means `>= 58.1`. We bump it up to try to enforce ourselves to "keep moving forward" to the 100% API documentation coverage :D

pezra:
It is pretty big but mostly self contained. Hopefully that will make it easier to understand.  I'm fixing many of the RuboCop atm. (Pretty sure it hates me as much as i despise it.:) ) Those issues are why the build is broken (the tests already pass) so it should go green with that commit.

I'll integrate any feed back asap so comment away.

Asmod4n:
\o/ I would replace the locking InMemoryStore with a https://github.com/ruby-concurrency/thread_safe collection, and maybe constantize the most use Strings for a performance boost.

Thanks for taking this over :)

pezra:
@Asmod4n, no problem. Thanks for getting it started.

.rubocop.yml
```diff
@@ -59,3 +59,6 @@ Style/StringLiterals:
 
 Style/TrivialAccessors:
   Enabled: false
+
+Style/AlignParameters:
+  Enabled: false
```
ixti: Please do not disable this cop.

ixti:
@pezra Can you please rename:

```
HTTP::Cache::RequestWithCacheBehavior -> HTTP::Request::Cached
HTTP::Cache::ResponseWithCacheBehavior -> HTTP::Response::Cached

HTTP::Cache::CacheControl -> HTTP::Cache::Control
```

Also, IMO `CacheControl` should be named `Cache::Headers` actually and should be a decorator/facade of `HTTP::Headers` instead, thus initializer will accept headers structure directly. I'll try to provide my proposal as commit. :))

.rubocop.yml
pezra: Done. I think this rule encourages harder to read code but it is your house. :)

ixti:
@pezra I have pushed my proposals into [caching branch](https://github.com/httprb/http.rb/tree/caching) (based on top of yours). Will highly appreciate if you will merge it into your one.

.rubocop.yml
ixti: It encourages consistency of alignment :stuck_out_tongue_winking_eye:

pezra:
@ixti, i merged your changes. Overall i like them. I'm not sure i like the `Cached` name. Many of the responses, and all of the requests, that get wrapped by those objects will not actually be cached. Maybe `Caching` or `Cacheable` would be a better choice? I'm happy to change that unless you object.

ixti:
@pezra `HTTP::{Response,Request}::Cacheable` sound good to me. No objections about renaming.
I had not looked deeper actually yet. Will try to jump on better review during a weekend.

pezra:
@ixti, i ended up going with `Caching` because both of the classes have a `#cacheable?` method. Having both `#cacheable` and `#cacheable?` methods with such radically different behavior is a recipe for confusion. Also, for requests `Caching` is much better description of what is being accomplished (and it is not worse for responses).

ixti:
Hm. I agree about `#cacheable` and `#cacheable?`, but `Caching` still seems bad to me ;)) Anyway, let's keep it as it is now. I still want to review them all more precisly.

pezra:
I'm open to a better name (or names) if we think of them.

PS: Anybody know what's up with the jruby build?

ixti:
jRuby is suffering from my improvement to servers init in specs

pezra:
OK, I think I'm done now. :) (Except, of course, for handling any feedback.)

PS: re naming: is either `HTTP::{Request,Response}::ForCache` or `...::ForCaching` better?

pezra:
I ran into a couple of issues when i tried using this in anger. These latest commits address those issues. This PR's build passes and i have tested this feature in another project. I think it is really ready for review now.

pezra:
@ixti, @tarcieri, is there anything i can do to help this along?

tarcieri:
@pezra I've been both on vacation and now that I'm back I have a cold, however @ixti has largely taken ownership of this project so if he can review it, that's sufficient for me.

ixti:
Well I was on vacation as well actually. I thought I will have a chance to review this PR, but life had different plans on my spare time... I will try to review this PR ASAP.

pezra:
No worries. I understand that reviewing things like this take time.

pezra:
Ping

tarcieri:
Haven't had time to review this, sorry. It's pretty big and I've been busy with other things.

http.gemspec
```diff
@@ -24,6 +24,7 @@ Gem::Specification.new do |gem|
 
   gem.add_runtime_dependency "http_parser.rb", "~> 0.6.0"
   gem.add_runtime_dependency "http-form_data", "~> 1.0.0"
+  gem.add_runtime_dependency "rack-cache",     "~> 1.2"
```
tarcieri: This makes rack a hard dependency of http.rb which I don't think I really like. Perhaps you could make `http/cache` an optional require that pulls in `rack-cache` when loaded?

tarcieri:
This looks fine to me with the caveat that it makes http.rb depend on rack and rack-cache even when caching isn't used.

I would prefer caching be optional and loaded when a user does `require 'http/cache'`

ixti:
I had some business issues myself. Just starting to catching up on time to spend on projects I love :D so will hope to finally completely review this PR this week for sure (max on weekend).

pezra:
@tarcieri, thanks for reviewing!

I change the file loading behavior so that rack, etc is not required until caching is enabled. For example, `HTTP.with_cache(metastore: ...)` causes the caching code, rack-cache, etc to be loaded. That deviates a little from your suggestion but seems to accomplish the goal while being a little easier on user.

tarcieri:
Seems good, thanks! Will wait to hear what @ixti says

Gemfile
```diff
@@ -26,6 +26,7 @@ group :test do
   gem "rspec-its"
   gem "rubocop"
   gem "yardstick"
+  gem "fakefs", :require => "fakefs/safe"
```
ixti: Seems like not used (and thus useless) dependency

spec/lib/http/client_spec.rb
```diff
@@ -71,6 +71,28 @@ def simple_response(body, status = 200)
     end
   end
 
+  describe "caching" do
+    before { require "http/cache" }
```
ixti: require should go into the top of `client_spec`.

spec/lib/http/options/merge_spec.rb
```diff
@@ -39,11 +40,12 @@
       :form      => {:bar => "bar"},
       :body      => "body-bar",
       :json      => {:bar => "bar"},
-      :headers   => {"Accept"  => "xml", "Foo" => "foo", "Bar" => "bar"},
+      :headers   => HTTP::Headers.coerce("Foo" => "foo", "Accept"  => "xml", "Bar" => "bar"),
```
ixti: Why does this change was needed?

spec/lib/http/response/caching_spec.rb
```diff
@@ -0,0 +1,202 @@
...
+
+  # Background
+  let(:cache_control) { "" }
+  let(:headers) { {"cache-control" => cache_control} }
+  let(:response) { HTTP::Response.new(200, "http/1.1", headers, "") }
+
+  describe "#caching" do
+    subject(:caching_response) { response.caching }
+    it { is_expected.to be caching_response }
+  end
```
ixti: I assume that I have suggested this change, but this is something like `expect(true).to be true`.

ixti:
Please, move `let` statements in specs to the top of blocks they belong to.

Gemfile
pezra: It is left over from a failed approach to testing the caching stuff. Removed.

spec/lib/http/client_spec.rb
pezra: done

spec/lib/http/options/merge_spec.rb
pezra: Also left over from a refactor that didn't pan out. Removed.

pezra:
> Please, move let statements in specs to the top of blocks they belong to.

Where *you* think they belong. :) I prefer to put the important part of the tests front and center and leave the less interesting support structures below the fold.

It's your code base, though, so i'll do that.

ixti:
@pezra I mean following:

``` ruby
describe :a do
  let(:b) { 1 }
  describe :c do
    let(:d) { 2 }
  end
end
```

`:b` belongs to block `:a`.
`:c` belongs to `:a` (declared inside `:a`) so it has access to `:b` as well.
`:d` belongs to block `:c`.

I just prefer to read specs from top to the bottom. And it's a bit inconvenient to me to scroll down to understand "initial" data. :D

ixti:
In other words. `let` goes first things in the block. Then goes `subject` if it is needed. Then before/after hooks. Then `it` examples. Then nested `context/describe`s.

spec/lib/http/response/caching_spec.rb
pezra: i think it is a bit more like `it "returns itself"`. I changed it though, because that is an implementation detail not an essential part of the behavior.

pezra:
> I just prefer to read specs from top to the bottom. 

Of course. However, if a spec's purpose cannot be understood without referring to the let definitions i consider that a naming failure. One better fixed by finding highly indicative names rather than rearranging things so that the glossary comes first.

pezra:
@ixti, i think all the issues you noted are sorted out now.

lib/http/request.rb
```diff
@@ -80,6 +82,8 @@ def initialize(verb, uri, headers = {}, proxy = {}, body = nil, version = "1.1")
 
       @headers["Host"]        ||= default_host
       @headers["User-Agent"]  ||= USER_AGENT
+      now = Time.now
+      @headers["Date"] = now.httpdate
```
ixti: > Clients SHOULD only send a Date header field in messages that include an entity-body, as in the case of the PUT and POST requests, and even then it is optional. A client without a clock MUST NOT send a Date header field in a request.

> The HTTP-date sent in a Date header SHOULD NOT represent a date and time subsequent to the generation of the message. It SHOULD represent the best available approximation of the date and time of message generation, unless the implementation has no means of generating a reasonably accurate date and time. In theory, the date ought to represent the moment just before the entity is generated. In practice, the date can be generated at any time during the message origination without affecting its semantic value. 

http://www.w3.org/Protocols/rfc2616/rfc2616-sec14.html#sec14.18

So, looks like this should go to `Request::Caching` instead. And also we MUST not override headers if they were explicitly given by user.

ixti: Actually I'm not sure if this still needed?

pezra: Good catch.

ixti:
I have merged PR into master. Thanks!

pezra:
Yay! Thanks @ixti.

PR #178, opened 2015-03-02
This should address issue 112

hiromipaw:
After a quick chat on Friday with @ixti, thought to address issue 112: https://github.com/httprb/http.rb/issues/112
Please let me know if I haven't understood the specs correctly and if anything is missing.

PR #179, opened 2015-03-16
Refactor redirector logic

### LINKED ISSUE #112: Enhance Redirector logic
Redirector should allow to act in two modes: `strict` *(default)* and `non-strict`:

``` ruby
# strict mode with 5 max hops
HTTP.follow(true)
HTTP.follow(:strict => true)
HTTP.follow(:max_hops => 5, :strict => true)

# strict mode with unlimited hops
HTTP.follow(:max_hops => false)
HTTP.follow(:max_hops => false, :strict => true)

# non-strict mode with 5 max hops
HTTP.follow(:strict => false)

# non-strict mode with unlimited hops
HTTP.follow(:max_hops => false, :strict => false)
```

##### Strict mode VS non-strict

Thus in strict mode if original request was neither GET nor HEAD, and it received redirect 300, 301 or 302, a `StateError` exception will be raised. In non-strict mode, request will be redirected with `GET` verb. Notice that redirects with 307 and 308 codes will work same in both modes - redirect preserving verb of original request.

Notice, that such behavior is a mix of current [RFC](https://tools.ietf.org/html/rfc2616#section-10.3) and [proposed (1)](https://tools.ietf.org/html/draft-ietf-httpbis-p2-semantics-26#section-6.4) [changes (2)](https://tools.ietf.org/html/draft-reschke-http-status-308-07) to that RFC.

RFC2616 says that upon any request but GET or HEAD, if response is redirect 300, 301, 302 or 307, user agent SHOULD NOT continue. Extension to that RFC says that old clients actually will change verb to GET in this case, but it also says, that correct behavior is to follow redirect with original verb if that verb is defined as "safe" (all methods are safe by default except PUT, DELETE and POST), otherwise user agent should halt. Experimental RFC that introduces redirect 308 says that 301 and 302 redirects should be followed with GET verb while 307 and 308 with original verb.

Also, notice that, main difference between 301 VS 308 (permanent redirect) and 302 VS 307 (temporary redirect) is the way how user agent should handle caching (this comes from original RFC).

##### List of references to keep in mind:

- https://tools.ietf.org/html/rfc2616#section-10.3
- https://tools.ietf.org/html/draft-reschke-http-status-308-07
- https://tools.ietf.org/html/draft-ietf-httpbis-p2-semantics-26#section-6.4

Previous issue: #94

ixti:
Resolves #112
Overtakes #178

### DIFF
--- a/lib/http/chainable.rb
+++ b/lib/http/chainable.rb
@@ -106,7 +106,7 @@ module HTTP
     # @param opts
     # @return [HTTP::Client]
     # @see Redirector#initialize
-    def follow(opts = true)
+    def follow(opts = {})
       branch default_options.with_follow opts
     end
 
--- a/lib/http/client.rb
+++ b/lib/http/client.rb
@@ -39,13 +39,11 @@ module HTTP
       req = HTTP::Request.new(verb, uri, headers, proxy, body)
       res = perform req, opts
 
-      if opts.follow
-        res = Redirector.new(opts.follow).perform req, res do |request|
-          perform request, opts
-        end
-      end
+      return res unless opts.follow
 
-      res
+      Redirector.new(opts.follow).perform req, res do |request|
+        perform request, opts
+      end
     end
 
     # Perform a single (no follow) HTTP request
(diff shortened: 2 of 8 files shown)


tarcieri:
Test failures?

PR #182, opened 2015-03-25
[DO NOT MERGE] Work on conn reuse

zanker:
Very much WIP, @tarcieri rough idea based on what you were doing previously. Thoughts before I go through and clean up and actually make it all work?


zanker:
Branch messed up a bit, going to recreate the PR

PR #183, opened 2015-03-25
Backport #180 (CVE-2015-1828) in 0-6-stable branch

nicoolas25:
Here is a very naive backport of @zanker's SSL host verification. I would like to have this in a 0.6.x version then add this patched version to https://github.com/rubysec/ruby-advisory-db/blob/master/gems/http/CVE-2015-1828.yml too.

Tell me if I can do anything else about this.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -15,7 +15,7 @@ MethodLength:
 
 ClassLength:
   CountComments: false
-  Max: 100
+  Max: 110
 
 CyclomaticComplexity:
   Max: 13 # TODO: lower to 6
--- a/Gemfile
+++ b/Gemfile
@@ -26,6 +26,7 @@ group :test do
   gem 'simplecov', '>= 0.9'
   gem 'yardstick'
   gem 'certificate_authority'
+  gem 'activemodel', '~> 3.0'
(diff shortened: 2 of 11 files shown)


ixti:
Can you explain how exactly `activemodel` and `i18n` adds ruby 1.8 compatibility?
I'm not against them - just want to understand if they are really needed, or if we should better
add some shims into: https://github.com/httprb/http.rb/blob/0-6-stable/lib/http/backports.rb

nicoolas25:
I explained it [here](https://github.com/httprb/http.rb/pull/183#commitcomment-10391460). Everything is coming from `certificate_authority`. I don't think the code in `http.rb` needs it.

ixti:
Oh. Indeed (re certificate_authority). All good then :D Thanks for explanation.

ixti:
Thanks for your help. Patch released as part of `0.6.4`.
Update for CVE proposed: https://github.com/rubysec/ruby-advisory-db/pull/139

nicoolas25:
Many thanks to you and @zanker! I'll soon remove my `--ignore` flag from `bundle-audit` and that feels good.

PR #184, opened 2015-03-26
Added connection reuse

zanker:
Adds a connection reuse API to http.rb, via `HTTP.persistent("http://www.google.com")`.

```ruby
client = HTTP.persistent("http://www.google.ocm")
client.get("http://www.google.com/foo")
client.get("http://www.google.com/bar")
```

Results in reusing a connection. I went with the redundant host being passed because it keeps the API consistent. I could make it just `/foo` or `/bar` calls too.


zanker:
cc @tarcieri

zanker:
Last bit of work in and it should go green now.

I've added as many state checks for things like recovering from errors that I can think of. Since body's are lazily loaded, I've added sequence IDs. If you make multiple requests, you need to either have read out a body before making the next one, or you can't use the body later without it erroring.

ixti:
There was a discussion about connection re-use: https://github.com/httprb/http.rb/issues/121
Just want to get attention that in future we would like to have connections pool. I think of API like:


``` ruby
# @param [Hash] opts
# @option opts [Array<#to_s>] :hosts ([]) Makes sure connections persist for
#   the given list of hosts only. Any host if empty.
# @option opts [#to_i] :pool_size (5) Amount of simultaneous open connections. 
```

zanker:
My concern with using a connection pool is that it adds a lot of additional complexity and I'm not sure that call patterns are worth it.

The `connection_pool` gem works great, and I'm not sure how beneficial it is to really reuse a client across multiple hosts.

tarcieri:
Yeah, I like that this is doing the simple thing, and agree that if pools are needed, `connection_pool` can be used.

ixti:
Then I must admit I was thinking about some absolutely different API :D
But I agree that this version is pretty neat :D 
After all advanced pooling can be added later as part of core or as a separate extension gem.

sferik:
As the person who initially requested this feature, I’m :+1: on this pull request. The API is essentially identical to what I proposed in https://github.com/httprb/http.rb/issues/121. Nice work @zanker! Thank you for implementing this!

zanker:
@sferik Hah no problem. We're replacing our use of Net::HTTP::Persistent and the only gems with proper conn reuse are either JRuby or libcurl based :(

zanker:
Made a quick change. On any exception we reset the conn, rather than a specific subset of network related ones.

Mostly it's a safety measure to be sure we don't have mixed requests/response.

ixti:
:+1:

PR #186, opened 2015-03-26
First attempt at supporting https connections through proxy

Connorhd:
This is pretty rough but I wanted to see if this looks like I'm at all going in the right direction. I added a failing test which now passes, so I think this is at least functional. Any feedback on what I need to do to get this into a mergable state would be great.

Connecting to https servers through a http proxy requires sending a CONNECT request to the proxy to setup a tunnel, then performing a normal https request through that tunnel. This PR adds the code to send that CONNECT request before continuing with the normal request.

### DIFF
--- a/lib/http/client.rb
+++ b/lib/http/client.rb
@@ -64,8 +64,11 @@ module HTTP
       @state = :dirty
 
       @connection ||= HTTP::Connection.new(req, options)
-      @connection.send_request(req)
-      @connection.read_headers!
+
+      unless @connection.failed_proxy_connect?
+        @connection.send_request(req)
+        @connection.read_headers!
+      end
 
       res = Response.new(
         @connection.status_code,
--- a/lib/http/connection.rb
+++ b/lib/http/connection.rb
@@ -6,6 +6,7 @@ module HTTP
   # A connection to the HTTP server
   class Connection
     extend Forwardable
+
     # Attempt to read this much data
     BUFFER_SIZE = 16_384
 
@@ -18,16 +19,18 @@ module HTTP
     # @param [HTTP::Request] req
     # @param [HTTP::Options] options
     def initialize(req, options)
-      @persistent         = options.persistent?
-      @keep_alive_timeout = options[:keep_alive_timeout].to_f
-      @pending_request    = false
-      @pending_response   = false
+      @persistent           = options.persistent?
+      @keep_alive_timeout   = options[:keep_alive_timeout].to_f
+      @pending_request      = false
+      @pending_response     = false
+      @failed_proxy_connect = false
 
       @parser = Response::Parser.new
 
       @socket = options[:timeout_class].new(options[:timeout_options])
       @socket.connect(options[:socket_class], req.socket_host, req.socket_port)
 
+      send_proxy_connect_request(req)
       start_tls(req, options)
       reset_timer
     end
@@ -40,6 +43,11 @@ module HTTP
 
     # @see (HTTP::Response::Parser#headers)
     def_delegator :@parser, :headers
+
+    # @return [Boolean] whenever proxy connect failed
+    def failed_proxy_connect?
+      @failed_proxy_connect
+    end
 
     # Send a request to the server
     #
@@ -129,7 +137,7 @@ module HTTP
     # @param (see #initialize)
     # @return [void]
     def start_tls(req, options)
-      return unless req.uri.https? && !req.using_proxy?
+      return unless req.uri.https? && !failed_proxy_connect?
 
       ssl_context = options[:ssl_context]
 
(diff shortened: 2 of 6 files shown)


lib/http/request/writer.rb
```diff
@@ -23,12 +23,32 @@ def add_headers
         end
       end
 
+      # Adds headers to be sent to the proxy
+      def add_proxy_connect_headers
+        @headers.each do |field, value|
+          next unless ['Host', 'User-Agent'].include? field
```
ixti: It's better to move this array into a constant and make it a set:

``` ruby
PROXY_HEADERS_BLACKLIST = %w(Host User-Agent).to_set.freeze
```

ixti:
@Connorhd I have rebased and cleaned up a bit your PR. By some reason 3 examples are failing with "IOError: problem making HTTP request: end of file reached" now. Please, see branch:
[feature/https-through-proxy-support](https://github.com/httprb/http.rb/tree/feature/https-through-proxy-support).

/cc @zanker

Connorhd:
@ixti I've fixed the tests on your rebase, one of the condition changes got lost in the rebase, plus the ssl tests need the ssl client to be configured now.

tarcieri:
:thumbsup: from me

lib/http/client.rb
```diff
@@ -64,8 +64,11 @@ def make_request(req, options)
       @state = :dirty
 
       @connection ||= HTTP::Connection.new(req, options)
-      @connection.send_request(req)
-      @connection.read_headers!
+
+      unless @connection.failed_proxy_connect?
```
zanker: A failed proxy connection should be an error, rather than soft failing I think. It's the equivalent of a socket error in that you're saying to connect to X and it failed.

If the concern is around accessing the response for further debugging, we can do something like what `Net::HTTP` does and include `response` as part of the error.

Connorhd: It's unclear to me what the correct behavior is here. This case includes the 407 response when proxy authentication fails (which we have a test case for), its also consistent with the current behavior for http proxies.

i.e.
HTTP.get('http://not-real-sldjflkdjlkds.com') => exception
HTTP.via(proxy.addr, proxy.port).get('http://not-real-sldjflkdjlkds.com') => 500 response

zanker: Okie-doke. If it's consistent with what we have then :+1:

lib/http/connection.rb
```diff
@@ -129,7 +137,7 @@ def expired?
     # @param (see #initialize)
     # @return [void]
     def start_tls(req, options)
-      return unless req.uri.https? && !req.using_proxy?
+      return unless req.uri.https? && !@failed_proxy_connect
```
zanker: Would use the `failed_proxy_connect?` method here

lib/http/connection.rb
```diff
@@ -141,6 +149,20 @@ def start_tls(req, options)
       @socket.start_tls(req.uri.host, options[:ssl_socket_class], ssl_context)
     end
 
+    # Open tunnel through proxy
+    def send_proxy_connect_request(req)
+      return unless req.uri.https? && req.using_proxy?
+
+      req.connect_using_proxy @socket
```
zanker: Can we set the pending flags for this like below? I can't think of a case where it's strictly necessary, but when it comes to sockets and `Timeout.timeout` I'd like to be 100% sure we're protected since these issues can be subtle and nasty if they do popup.

```ruby
def send_proxy_connect_request(req)
  return unless req.uri.https? && req.using_proxy?
  
  @pending_request = true

  req.connect_using_proxy @socket

  @pending_request = false
  @pending_response = true

  read_headers!

  if @parser.status_code == 200
    @parser.reset
    @pending_response = false
    return
  end

  @failed_proxy_connect = true
end
```

zanker:
Besides that looks good to me. If you could fix those two things can merge. Thanks for working through this!

Connorhd:
@zanker Updated with those changes

ixti:
Awesome! Thanks@

timrwilliams:
Hi guys, trying to test this out in irb using version 0.8.5 and it hangs whenever I try and access an HTTPS endpoint. 

   HTTP.via("176.31.170.159", 8080).get("https://www.google.com/")

Has anybody got this working against a real proxy server? It works fine from curl on the command line.

    curl --proxy http://176.31.170.159:8080 -k https://www.google.co.uk/

(Sample HTTP proxies taken from http://proxylist.hidemyass.com/search-1297445#listable)

ixti:
Thanks for report! I confirm bug exists.

ixti:
@timrwilliams Fixed in master. Will cut new release today or tomorrow.

timrwilliams:
Nice, I've tested master and it works like a charm now. We've got an issue over on the Twitter gem (https://github.com/sferik/twitter/pull/663) that needs this so cutting a new release would be really helpful.

ixti:
@timrwilliams OK.

ixti:
@timrwilliams Released as 0.8.8

Issue #160, opened 2014-10-02
timeout


ghost:
Is there a timeout option available?
thank you

PR #187, opened 2015-03-26
Added configurable connection timeouts

zanker:
This is mildly crazy. Basically this adds three sets of timeouts

Null, which is just what there is today (nothing)
Per operation, which is what you see in things like Net::HTTP, Excon where the timeout is per operation
Global, which sets an allocation of how much time can be used and uses it up as the request goes. Ensuring a request can never run longer than the set time.

Getting DNS to have a customizable timeout is awful though, which is why there's a lot of extra code for doing `connect` (mostly cribbed from Tony's code). My inclination, would be to just wrap it in `Timeout.timeout` since that can be done more safely than relying on all the Ruby internals to handle the DNS/open timeouts properly. And it's safer, any thoughts on that?

### DIFF
--- a/lib/http.rb
+++ b/lib/http.rb
@@ -1,6 +1,9 @@
 require "http/parser"
 
 require "http/errors"
+require "http/timeout/null"
+require "http/timeout/per_operation"
+require "http/timeout/global"
 require "http/chainable"
 require "http/client"
 require "http/connection"
--- a/lib/http/client.rb
+++ b/lib/http/client.rb
@@ -17,6 +17,7 @@ module HTTP
 
     def initialize(default_options = {})
       @default_options = HTTP::Options.new(default_options)
+      @connection = nil
     end
 
     # Make an HTTP request
--- a/lib/http/connection.rb
+++ b/lib/http/connection.rb
@@ -16,14 +16,22 @@ module HTTP
 
       @parser = Response::Parser.new
 
-      @socket = options[:socket_class].open(req.socket_host, req.socket_port)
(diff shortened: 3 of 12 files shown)


ixti:
I guess this will close #160

zanker:
Yup! This is kind of a complicated set of changes though. See any issues with it?

sferik:
@zanker It looks like the “HTTP::Client timeouts with a per operation timeout connection of 0 times out” specs are failing in CI.

zanker:
Haha awesome. I can make it go green locally, but I expect the CI servers are running fast enough that it immediately connects, or maybe they do something strange.

I don't have a good way of testing connection latency. Since it's "connected" before I call accept on a server, and if the servers not up it just fails immediately. Let me try setting a float timeout and see if that hacks it into working.

tarcieri:
Timeouts, especially ones related to I/O so they can't simply be mocked, have been rather troublesome for me to test in things like Celluloid. Short of using something like rspec-retry and rerunning the test until it passes, I'm not sure I have any good advice.

On Celluloid we use a "quantum" of 50ms for all timer tests.

zanker:
:(

I stripped that test then

tarcieri:
Seems there are still 3 failing tests

zanker:
Think something got a bit messed up from merge conflicts. Should go green now

ixti:
Awesome!!! :clap:

tarcieri:
Released as 0.8.0.pre4

PR #188, opened 2015-03-27
Add a close API and us that instead for the client

zanker:
For our tests, we also want to close the connection to get a clean slate, but also since it's a pattern that's repeated, rather just make it a method.

### DIFF
--- a/lib/http/client.rb
+++ b/lib/http/client.rb
@@ -75,12 +75,14 @@ module HTTP
     # On any exception we reset the conn. This is a safety measure, to ensure
     # we don't have conns in a bad state resulting in mixed requests/responses
     rescue
-      if default_options.persistent? && @connection
-        @connection.close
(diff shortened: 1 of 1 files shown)


ixti:
IMO the name of method is not the best. It might be only me, but having `#close` makes me think that htere'should be `#open`.

Probably it should be named `#close_connections` or `#reset!`. Can't find any better names...

tarcieri:
I'm ok with `#close`

sferik:
IMO, `#close` is a :+1: name. We could alias `#persistent` to `#open` for the sake of parallelism.

ixti:
Aliasing `#persistent` to `#open` maeks sense to me. :+1:

tarcieri:
Gonna go ahead and merge this then. We can loop back on `persistent/open` in a separate issue/PR

PR #189, opened 2015-03-27
Alias HTTP::Chainable#persistent to #open

sferik:
Also, alias `HTTP::Options#persistent?` to `#open?`.

Idea came out of the discussion in https://github.com/httprb/http.rb/pull/188.

We can discuss it further here.

### DIFF
--- a/lib/http/chainable.rb
+++ b/lib/http/chainable.rb
@@ -78,6 +78,7 @@ module HTTP
     def persistent(host)
       branch default_options.with_persistent host
     end
+    alias_method :open, :persistent
 
     # Make a request through an HTTP proxy
     # @param [Array] proxy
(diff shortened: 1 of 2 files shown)


sferik:
Arguably, `#open` should be the canonical name and `#persistent` should be the alias. It doesn’t really matter, so I don’t have a very strong feeling either way, but `#open` is shorter, easier to spell, and offers a nice parallel to `#close`, which was just merged.

zanker:
Seems fine to me, but as a note. `close` is actually closing a socket, while `open` is just marking the client as being dedicated to a host. Not sure if that influences your decision on this.

ixti:
I'm sorry to be pain in the neck :D but in the light of:

> open is just marking the client

I think that we might better `alias :reset :close`. Or not adding aliases at all...

sferik:
> `close` is actually closing a socket, while `open` is just marking the client as being dedicated to a host.

I suppose `open` is a pretty confusing name for that. I’m going to close this.

PR #192, opened 2015-03-29
Added SSL config via the ssl param instead of requiring context passing

zanker:
Makes it a bit easier to configure things for SSL. As an advantage, this gives people more secure SSL configuration by default via the `DEFAULT_PARAMS` in https://github.com/ruby/openssl/blob/master/lib/openssl/ssl.rb#L25.

Personally, I'm not sure how much sense allowing `ssl_context` makes, since I wasn't aware there was really competition for other SSL gems in Ruby. Although I defer to @tarcieri on that.

### DIFF
--- a/lib/http/connection.rb
+++ b/lib/http/connection.rb
@@ -22,7 +22,7 @@ module HTTP
       @socket.start_tls(
         req.uri.host,
         options[:ssl_socket_class],
-        options[:ssl_context]
+        options[:ssl_context] || setup_ssl_context(options[:ssl])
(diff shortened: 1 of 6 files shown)


tarcieri:
I like the hypothetical idea of abstracting away `ssl_context` because it makes it easier to move to potential ponies and unicorns SSL libraries in the future.

But there are probably people using the `ssl_context`-based API already, so providing both for now seems ok to me.

tarcieri:
Released as 0.8.0.pre5

zanker:
Makes sense to me. Thanks!

PR #193, opened 2015-03-29
Include JRuby 1.7 as a supported version

zanker:
Thoughts? I'm going to end up having to maintain this for JRuby incidentally anyway, so might as well include it as an officially supported language.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,6 +10,7 @@ rvm:
   - 2.2
   - jruby
   - jruby-head
+  - jruby-19mode
   - ruby-head
   - rbx-2
 matrix:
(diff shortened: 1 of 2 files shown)


ixti:
If you are going to support it, then it worth to add to `.travis.yml`:

``` yaml
# ...
rvm:
  - 1.9.3
  # ...
  - jruby-19mode
#...
```

zanker:
Good point, done

ixti:
Please, squash PR into a single commit.
@tarcieri LGTM

ixti:
Oh I just figured out that `-jruby` and `-jruby-19mode` are the same... Sorry. Seems like only first of your commits was needed.

ixti:
Sorry, 3:48am.. I'm a bit slow :D
Closed via https://github.com/httprb/http.rb/commit/cb3c9b1d3c45e81b71258b35cbca949bc508f42e

PR #194, opened 2015-03-29
Add an extra set of state tracking for Timeout.timeout edge cases

zanker:
Thoughts? This isn't strictly necessary, as the chance of this happening is very very small. It is still a chance though.

Normally, `Timeout.timeout` raises an exception and kills the block and everything is handy dandy. In rare situations, it will raise and then kill execution while the `rescue` block executes (see https://github.com/mperham/connection_pool/pull/67 for more info).

This adds an extra layer of protection, where if we run into that, we reset the connection anyway.

### DIFF
--- a/lib/http/client.rb
+++ b/lib/http/client.rb
@@ -19,6 +19,7 @@ module HTTP
     def initialize(default_options = {})
       @default_options = HTTP::Options.new(default_options)
       @connection = nil
+      @state = :clean
     end
 
     # Make an HTTP request
@@ -58,6 +59,8 @@ module HTTP
     def make_request(req, options)
       verify_connection!(req.uri)
 
+      @state = :dirty
+
       @connection ||= HTTP::Connection.new(req, options)
       @connection.send_request(req)
       @connection.read_headers!
@@ -71,6 +74,7 @@ module HTTP
       )
 
       @connection.finish_response if req.verb == :head
+      @state = :clean
 
       res
 
@@ -85,6 +89,7 @@ module HTTP
     def close
       @connection.close if @connection
       @connection = nil
+      @state = :clean
     end
 
     private
@@ -93,10 +98,13 @@ module HTTP
     def verify_connection!(uri)
       if default_options.persistent? && base_host(uri) != default_options.persistent
         fail StateError, "Persistence is enabled for #{default_options.persistent}, but we got #{base_host(uri)}"
-
       # We re-create the connection object because we want to let prior requests
       # lazily load the body as long as possible, and this mimics prior functionality.
       elsif @connection && (!@connection.keep_alive? || @connection.expired?)
+        close
+      # If we get into a bad state (eg, Timeout.timeout ensure being killed)
+      # close the connection to prevent potential for mixed responses.
+      elsif @state == :dirty
         close
       end
     end
--- a/spec/support/http_handling_shared.rb
+++ b/spec/support/http_handling_shared.rb
@@ -95,13 +95,13 @@ RSpec.shared_context "HTTP handling" do |ssl = false|
       it "errors if connecting takes too long" do
         socket = Socket.new(Socket::AF_INET, Socket::SOCK_STREAM, 0)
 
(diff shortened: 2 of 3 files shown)


lib/http/client.rb
```diff
@@ -98,6 +103,11 @@ def verify_connection!(uri)
       # lazily load the body as long as possible, and this mimics prior functionality.
       elsif @connection && (!@connection.keep_alive? || @connection.expired?)
         close
+
```
ixti: Why do you add this extra line? :D

ixti: I mean, can you try to avoid it? :D

spec/support/http_handling_shared.rb
```diff
@@ -148,6 +148,18 @@
         expect(sockets_used.uniq.length).to eq(1)
       end
 
+      context "on a mixed state" do
+        it "re-opens the connection" do
+          first_socket = client.get("#{server.endpoint}/socket/1").body.to_s
```
ixti: Varname is a bit confusing IMO

spec/support/http_handling_shared.rb
```diff
@@ -148,6 +148,18 @@
...
+
+          client.instance_variable_set(:@state, :dirty)
```
ixti: I wish there were a way to test it without patching internal state. Like enforce timeout... :D

lib/http/client.rb
zanker: I find it makes it harder to read personally, but I don't really care either way.

spec/support/http_handling_shared.rb
zanker: That's the same one used for other tests currently.

tarcieri:
Seems ugly but so does Timeout :stuck_out_tongue_winking_eye: In absence of a fully asynchronous timeout scheme this seems ok to me.

zanker:
Yea it's not great unfortunately but such is life :(

spec/support/http_handling_shared.rb
ixti: That does not excuse neither this case nor the others ;))
I mean `first_socket` is not a socket at all. It's a string. I would understand name `first_response` or `first_socket_response`...

lib/http/client.rb
```diff
@@ -93,11 +98,14 @@ def close
     def verify_connection!(uri)
       if default_options.persistent? && base_host(uri) != default_options.persistent
         fail StateError, "Persistence is enabled for #{default_options.persistent}, but we got #{base_host(uri)}"
-
```
ixti: Thanks! :bow:

spec/support/http_handling_shared.rb
zanker: Given the context of it being a body response I think it's fine personally, response isn't really any clearer. I did `first/last_socket_id` instead as that's more clear.

ixti: With all respect but how `first_socket_id = <...>.body.to_s` makes it more clear? You are saving first socket's response body into a variable named `first_socket_id`. That makes as much sense as naming it `first_socket`. IMO even naming variables `a` and `b` will be less confusing in this case.

zanker: Huh? This is the socket ID you made the request with,
https://github.com/httprb/http.rb/blob/master/spec/support/dummy_server/servlet.rb#L61.
The response body is the socket ID.

ixti: Yeah, just figured out that... Now that really makes more sense! :D But still dislike the names :D Although I can't come with better ones anyway :D

ixti: Sorry for that useless noise. Had not enough of sleep. Will try to avoid commenting while in a grumpy old man mode next time. :D Or at least will try to get into details prior making any judgements. :D Sorry once again.

zanker: No problem, your confusion makes more sense now.

PR #195, opened 2015-03-30
Use Timeout.timeout for TCP connection

zanker:
Unfortunately, getting a pure solution to timeouts is essentially impossible without rewriting the entire stack yourself. Which I might actually do for fun later, but for the time being. This is the simplest way of getting a connect timeout :(

### DIFF
--- a/lib/http/timeout/per_operation.rb
+++ b/lib/http/timeout/per_operation.rb
@@ -1,6 +1,3 @@
-# rubocop:disable Lint/HandleExceptions
-require "resolv"
-
 module HTTP
   module Timeout
     class PerOperation < Null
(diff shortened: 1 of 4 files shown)


zanker:
@httprb/owners fyi if someone can :+1: quickly

lib/http/timeout/per_operation.rb
```diff
@@ -20,40 +17,22 @@ def initialize(*args)
...
 
-        case addr
-        when Resolv::IPv4
-          family = Socket::AF_INET
-        when Resolv::IPv6
-          family = Socket::AF_INET6
-        else fail ArgumentError, "unsupported address class: #{addr.class}"
+      def connect_ssl
+        socket.connect_nonblock
+      rescue IO::WaitReadable
```
tarcieri: I guess I missed this on the last pass, but as the SSL/TLS handshake makes many roundtrips, I believe it's possible for IO::WaitWritable to be raised here too.

tarcieri:
This looks fine to me

lib/http/timeout/per_operation.rb
zanker: Oops I broke Excon then!!

ixti:
:+1: from me as well :D

PR #197, opened 2015-03-31
Add support for non-ascii URIs

### LINKED ISSUE #196: Dealing with non-ascii URIs (IRIs) in redirects
I'm running into problems when getting a URI that happens to redirect to a URI with non-ascii characters in it (an [IRI](http://en.wikipedia.org/wiki/Internationalized_resource_identifier) from what I understand). A simple, repeatable case (in this case using http.rb 0.7.3 and Ruby 2.2.0) is:

```ruby
require 'http'

REDIRECTS_TO_UMLAUT_URL = 'http://www.amazon.co.uk/dp/B00C6Q3SEQ'
http_client = HTTP.follow(true)
http_client.get(REDIRECTS_TO_UMLAUT_URL)
```

Which results in this error:
```
/Users/jamesrucker/.rbenv/versions/2.2.0/lib/ruby/2.2.0/uri/generic.rb:1100:in `rescue in merge': URI must be ascii only "http://www.amazon.co.uk/Bianka-Minte-K\xC3\xB6nig-Emilia-Sch\xC3\xBCle/e/B001K1JKKY/digital/ref=ntt_mp3_rdr/276-0649186-0513969?_encoding=UTF8&sn=d" (URI::InvalidURIError)
	from /Users/jamesrucker/.rbenv/versions/2.2.0/lib/ruby/2.2.0/uri/generic.rb:1097:in `merge'
	from /Users/jamesrucker/.rbenv/versions/2.2.0/lib/ruby/gems/2.2.0/gems/http-0.7.3/lib/http/request.rb:87:in `redirect'
	from /Users/jamesrucker/.rbenv/versions/2.2.0/lib/ruby/gems/2.2.0/gems/http-0.7.3/lib/http/redirector.rb:47:in `follow'
	from /Users/jamesrucker/.rbenv/versions/2.2.0/lib/ruby/gems/2.2.0/gems/http-0.7.3/lib/http/redirector.rb:22:in `perform'
	from /Users/jamesrucker/.rbenv/versions/2.2.0/lib/ruby/gems/2.2.0/gems/http-0.7.3/lib/http/client.rb:35:in `request'
	from /Users/jamesrucker/.rbenv/versions/2.2.0/lib/ruby/gems/2.2.0/gems/http-0.7.3/lib/http/chainable.rb:16:in `get'
	from ./fetch_umlaut_url.rb:16:in `<main>'
```

From doing [a bit of googling](http://stackoverflow.com/questions/10519591/how-to-use-umlauts-with-rubys-uri-library), it sounds like two possible solutions are to call URI.encode before URI.parse (which isn't foolproof...it will encode hash fragments for instance) or to use the [addressable gem](https://github.com/sporkmonger/addressable), neither of which seemed like great options for a PR.  Is there a more reasonable solution?

ixti:
Resolves #196

### DIFF
--- a/lib/http/client.rb
+++ b/lib/http/client.rb
@@ -1,9 +1,11 @@
 require "cgi"
 require "uri"
+
 require "http/form_data"
 require "http/options"
 require "http/connection"
 require "http/redirector"
+require "http/uri"
 
 module HTTP
   # Clients make requests and receive responses
@@ -13,6 +15,8 @@ module HTTP
     CONNECTION         = "Connection".freeze
     KEEP_ALIVE         = "Keep-Alive".freeze
     CLOSE              = "close".freeze
+
+    HTTP_OR_HTTPS_RE   = %r{^https?://}i
 
     attr_reader :default_options
 
@@ -109,33 +113,25 @@ module HTTP
 
     # Strips out query/path to give us a consistent way of comparing hosts
     def base_host(uri)
-      base = uri.dup
-      base.query = nil
-      base.path = ""
-      base.to_s
+      uri.omit(:query, :path).to_s
     end
 
     # Merges query params if needed
-    def make_request_uri(uri, options)
-      uri = normalize_uri uri
-
-      if options.params && !options.params.empty?
-        params    = CGI.parse(uri.query.to_s).merge(options.params || {})
-        uri.query = URI.encode_www_form params
-      end
-
-      uri
-    end
-
-    # Normalize URI
     #
     # @param [#to_s] uri
     # @return [URI]
-    def normalize_uri(uri)
-      if default_options.persistent? && uri !~ /^http|https/
-        uri = URI("#{default_options.persistent}#{uri}")
-      else
-        uri = URI(uri.to_s)
+    def make_request_uri(uri, options)
+      uri = uri.to_s
+
+      if default_options.persistent? && uri !~ HTTP_OR_HTTPS_RE
+        uri = "#{default_options.persistent}#{uri}"
+      end
+
+      uri = HTTP::URI.parse uri
+
+      if options.params && !options.params.empty?
+        params    = CGI.parse(uri.query.to_s).merge(options.params || {})
+        uri.query = ::URI.encode_www_form params
       end
 
       # Some proxies (seen on WEBRick) fail if URL has
--- a/http.gemspec
+++ b/http.gemspec
@@ -23,7 +23,8 @@ Gem::Specification.new do |gem|
   gem.version       = HTTP::VERSION
 
   gem.add_runtime_dependency "http_parser.rb", "~> 0.6.0"
-  gem.add_runtime_dependency "http-form_data", "~> 1.0.0"
+  gem.add_runtime_dependency "http-form_data", "~> 1.0.1"
+  gem.add_runtime_dependency "addressable",    "~> 2.3"
 
   gem.add_development_dependency "bundler", "~> 1.0"
 end
--- a/lib/http/connection.rb
+++ b/lib/http/connection.rb
@@ -129,7 +129,7 @@ module HTTP
     # @param (see #initialize)
     # @return [void]
     def start_tls(req, options)
-      return unless req.uri.is_a?(URI::HTTPS) && !req.using_proxy?
+      return unless req.uri.https? && !req.using_proxy?
 
       ssl_context = options[:ssl_context]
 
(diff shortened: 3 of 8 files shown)


lib/http/uri_normalizer.rb
```diff
@@ -0,0 +1,25 @@
...
+
+module HTTP
+  module URINormalizer
+    module_function
+
+    # Normalize URI
+    #
+    # @param [#to_s] uri
+    # @return [::URI]
+    def normalize_uri(uri)
```
zanker: Can we just use `addressable`? I see in the comments on the issue it's mentioned, but I don't see a problem with adding a dependency on something that supports URIs more closely to the RFC. This workaround seems very ugly :(

tarcieri: I'm fine with a hard dependency on addressable

ixti: Well, I'm 100% agree that this hack is ugly as <put any word you think more appropriate>.
And I'm OK with Addressable, but not as hard dependency.

tarcieri: Why not?

zanker: If we're going to support non-ASCII, we should fully support them imo. Eg, `http://🚚.la` doesn't work for URI, while it does for Addressable.

ixti: Although Addressable is absolutely awesome gem, it's about 1.7x slower on parsing URIs.

ixti: Anyway, I guess I'm just trying to reject any changes today.. :D Need to go sleep! :D Will replace this patch with addressable based solution ;))

ixti:
@httprb/owners thoughts? I would sneak this into 0.8.0 release.

tarcieri:
This seems ok to me

digitalextremist:
The only thing I could see adding is a test with non-ascii characters in a URI.

sferik:
:+1: I think `Addressable::URI` trying to be a drop-in replacement for the `URI` standard library (similar to how `FasterCSV` replaced `CSV` in the standard library).

I made the decision to replace `URI` with `Addressable::URI` in the `twitter` gem [over a year ago](https://github.com/sferik/twitter/commit/7ea2f5390dc7456950f55c90aa4e48f29dcd4604) prompted by a [similar issue](https://github.com/sferik/twitter/issues/487).

sferik:
FWIW, the relevant specs in the `twitter` gem look like this: https://github.com/sferik/twitter/blob/6750ece8a9c44482c073958814810b6cf27e1d1a/spec/twitter/tweet_spec.rb#L354-L368.

ixti:
@digitalextremist Isn't this is enough?

https://github.com/httprb/http.rb/pull/197/files#diff-3ef70b884bdcb4e458a8230f6a2b4fb8R81
https://github.com/httprb/http.rb/pull/197/files#diff-3ef70b884bdcb4e458a8230f6a2b4fb8R171

I'm ok to add more examples though.

ixti:
We can add following URLs (real ones) to tests if you think that worth it:

- https://github.com/httprb/http.rb/pull/197/könig
- http://git.io/jSxl

Both of above will lead to this page :D

digitalextremist:
@ixti I think it's worth it!

ixti:
@digitalextremist OK. Will add later today.

digitalextremist:
Extra points for incorporating the `無` symbol @ixti :)

ixti:
@digitalextremist :+1:

lib/http/client.rb
```diff
@@ -109,33 +113,23 @@ def verify_connection!(uri)
 
     # Strips out query/path to give us a consistent way of comparing hosts
     def base_host(uri)
-      base = uri.dup
-      base.query = nil
-      base.path = ""
-      base.to_s
+      uri.omit(:query, :path).to_s
```
zanker: I already like Addressable more than URI!

ixti: :D

ixti:
That was a good idea to test behavior on a real connection...
/me still tries to debug -- connection just hangs...

ixti:
Ouch. found the issue :D

gmile:
TIL `addressable` gem consumes quite a bit of memory: https://github.com/zquestz/omniauth-google-oauth2/pull/193#issuecomment-170447701. Would a PR to move back to stdlib's very own URI implementation be feasible?

zanker:
Ackkkk :(.

I'm sure it's feasible, although I'm not sure how much it's reinventing the wheel ultimately.

ixti:
Moving to pure stdlib will require fair amount of work. But definitely possible.

PR #199, opened 2015-03-31
Make sure we catch WaitWritable for SSL

zanker:
@tarcieri going to merge on green just as a fyi

Looking at Excon's code, they only time they don't raise is on `read will block` and not on writes (until my change) where I technically broke it.

I'm still going to add this as a just in case, since the Ruby docs mention it.


tarcieri:
LGTM

PR #201, opened 2015-04-02
Allow call `#persistent` with block

### LINKED ISSUE #200: Persistent connection API
Right now basic usage looks like this:

``` ruby
pclient = HTTP.persistent "https://github.com"
pclient.get("https://github.com/httprb/http.rb").to_s # => http.rb homepage
pclient.get("/httprb/http.rb").to_s # => http.rb homepage, same as above
```

After we have done all our calls we need to close underlying connections:

``` ruby
pclient.close
```

Also, if we try to call another host, e.g.:

``` ruby
pclient.get("https://example.com").to_s
```

It will fail (as expected by our design):

```
HTTP::StateError: Persistence is enabled for https://github.com,
  but we got https://example.com
```

I propose to have different API design, that will provide 2 variants of usage. First one - is pretty simple:

``` ruby
pclient = HTTP.persistent

pclient.get("https://github.com/httprb").to_s
# opens persistent connection to https://github.com
# and retrieves /httprb

pclient.get("https://github.com/explore").to_s
# retrieves /explore

pclient.get("http://example.com").to_s
# closes persistent connection to https://github.com
# opens persistent connection to http://example.com
# retrieves /

pclient.close
# closes persistent connection to http://example.com
```

Second, will provide a better API when you have to deal with a single host only:

``` ruby
HTTP.persistent "https://github.com" do |gh_client|
  gh_client.get("/foo").to_s
  gh_client.get("/bar").to_s
end
# connection is auto-closed
```

ixti:
Example Usage:
==============

``` ruby
def keys(users)
  HTTP.persistent("https://github.com") do |http|
    users.map { |u| http.get("/#{u}.keys").to_s }
  end
end
```

Resolves #200

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,6 +1,6 @@
 ## master (unreleased)
 
-* _the future is unwritten_
+* Add block-form `#persistent` calls. See #200, #201. (@ixti)
 
 
 ## 0.8.0 (2015-04-01)
--- a/lib/http/chainable.rb
+++ b/lib/http/chainable.rb
@@ -72,11 +72,41 @@ module HTTP
       branch(options).request verb, uri
     end
 
-    # Flag as persistent
-    # @param [String] host
-    # @raise [Request::Error] if Host is invalid
+    # @overload persistent(host)
+    #   Flags as persistent
+    #   @param [String] host
+    #   @raise [Request::Error] if Host is invalid
+    #   @return [HTTP::Client] Persistent client
+    # @overload persistent(host, &block)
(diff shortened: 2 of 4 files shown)


CHANGES.md
```diff
@@ -1,6 +1,6 @@
 ## master (unreleased)
 
-* _the future is unwritten_
```
tarcieri: LOL Joe Strummer

tarcieri:
Looks good :+1:

CHANGES.md
ixti: :D

PR #210, opened 2015-04-16
Automatically use non-exception NIOs for improved performance

zanker:
Here's a fun one. Apparently, in 2.1.x (and as such, JRuby 9k), Ruby added an undocumented arg to the NIO methods that allows you to use them without exceptions and instead symbol returns of the results. This is only available for write/read and not connect.

Quick performance test of `read_nonblock` with and without exceptions. On MRI, without exceptions is 50% faster, on JRuby it's around 85% faster. This is about inline with what we expect, exceptions in Java/JRuby are much more costly than Ruby, which are already costly.

The problem like everything, is testing it is a bitch :(

cc @tarcieri @nerdrew

### DIFF
--- a/lib/http/timeout/global.rb
+++ b/lib/http/timeout/global.rb
@@ -35,29 +35,65 @@ module HTTP
         end
       end
 
-      # Read from the socket
-      def readpartial(size)
-        reset_timer
+      # NIO with exceptions
+      if RUBY_VERSION < "2.1.0"
+        # Read from the socket
+        def readpartial(size)
+          reset_timer
 
-        begin
-          socket.read_nonblock(size)
-        rescue IO::WaitReadable
-          IO.select([socket], nil, nil, time_left)
-          log_time
-          retry
+          begin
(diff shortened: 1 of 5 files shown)


tarcieri:
Rubbocopped!

Otherwise this is incredibly cool

lib/http/timeout/global.rb
```diff
@@ -35,29 +35,65 @@ def connect_ssl
         end
       end
 
-      # Read from the socket
-      def readpartial(size)
-        reset_timer
+      # NIO with exceptions
+      if RUBY_VERSION < "2.1.0"
```
tarcieri: Perhaps these could be factored into different backends?

zanker: I was thinking about it but I'd rather just do Socketry really. I was starting work on it last night but things blew up :(

Need to do a bit more testing on a lot of this though

tarcieri: Cool, this seems OK for now

PR #216, opened 2015-05-08
Updated documentation on timeouts

zanker:
@ixti the docs on timeout aren't quite right (namely they need the type of timeout). Celluloid::IO works again, since the `respond_to? :sync_close=` change was made. Although @tarcieri should tag a new celluloid-io that has that fix in!


ixti:
> namely they need the type of timeout

That's my defect :D It was supposed to be `:per_operation` if not specified - but I have mistyped it ;))

sferik:
:+1:

ixti:
Just FYI. I have fixed that defect (and covered with specs - why I didn't wrote them before - have no clue). And updated README section about that ;))

PR #218, opened 2015-05-09
Make sure port number isn't blank for proxy connect

### LINKED ISSUE #217: Using proxy hangs...
``` ruby
HTTP.via("176.31.170.159", 8080).get("https://github.com/ixti.keys")
```

hangs for some time and then fails with:

```
IOError: problem making HTTP request: Connection reset by peer
from /home/ixti/Projects/httprb/http.rb/lib/http/connection.rb:99:in `rescue in read_headers!'
```

See original description: https://github.com/httprb/http.rb/pull/186#issuecomment-100322375

Connorhd:
I think this fixes #217 the problem seemed to be that @uri.port was blank so a port wasn't being specified if the default port was used. Seems like this code should be using the helper methods anyway.

### DIFF
--- a/lib/http/request.rb
+++ b/lib/http/request.rb
@@ -132,7 +132,7 @@ module HTTP
 
     # Compute HTTP request header SSL proxy connection
     def proxy_connect_header
-      "CONNECT #{@uri.host}:#{@uri.port} HTTP/#{version}"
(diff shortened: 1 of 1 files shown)


ixti:
Good catch!

ixti:
Thanks!

PR #219, opened 2015-05-09
Force encoding to be consistent.

Connorhd:
Loading binary data through http.rb in jruby returns a utf-8 encoded string right now, which then errors if invalid utf-8 sequences are included.

Not sure if this is the right solution, but it fixes my problem and makes the encodings of returned data consistent across ruby versions.

### DIFF
--- a/lib/http/response/body.rb
+++ b/lib/http/response/body.rb
@@ -36,9 +36,9 @@ module HTTP
 
         begin
           @streaming = false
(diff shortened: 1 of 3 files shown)


ixti:
Closed via 1ab394f06da8b0ce5b9e26b21e07460351677277

ixti:
Thanks!

PR #221, opened 2015-05-11
Add cookies support

ixti:
@httprb/owners if no objections - I would like to merge and cut release.

### DIFF
--- a/lib/http/options.rb
+++ b/lib/http/options.rb
@@ -42,23 +42,23 @@ module HTTP
     end
 
     def initialize(options = {})
-      defaults = {:response =>         :auto,
-                  :proxy =>            {},
-                  :timeout_class =>    self.class.default_timeout_class,
-                  :timeout_options =>  {},
-                  :socket_class =>     self.class.default_socket_class,
-                  :ssl_socket_class => self.class.default_ssl_socket_class,
-                  :ssl =>              {},
-                  :cache =>            self.class.default_cache,
-                  :keep_alive_timeout  => 5,
-                  :headers =>          {}}
+      defaults = {
+        :response =>         :auto,
+        :proxy =>            {},
+        :timeout_class =>    self.class.default_timeout_class,
+        :timeout_options =>  {},
+        :socket_class =>     self.class.default_socket_class,
+        :ssl_socket_class => self.class.default_ssl_socket_class,
+        :ssl =>              {},
+        :cache =>            self.class.default_cache,
+        :keep_alive_timeout  => 5,
+        :headers =>          {},
+        :cookies =>          {}
+      }
 
       opts_w_defaults = defaults.merge(options)
       opts_w_defaults[:headers] = HTTP::Headers.coerce(opts_w_defaults[:headers])
-
-      opts_w_defaults.each do |(opt_name, opt_val)|
-        self[opt_name] = opt_val
-      end
+      opts_w_defaults.each { |(k, v)| self[k] = v }
     end
 
     def_option :headers do |headers|
@@ -145,3 +145,6 @@ module HTTP
     end
   end
 end
+
+# require cookies options
+require "http/options/cookies"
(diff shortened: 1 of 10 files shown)


sferik:
Looks like this changes the header format, which is causing tests to fail?

```diff
-:headers => {"Foo"=>"foo", "Accept"=>"xml", "Bar"=>"bar"},
+:headers => [["Foo", "foo"], ["Accept", "xml"], ["Bar", "bar"]],
```

Is this a necessary/desirable change?

lib/http/options.rb
```diff
@@ -145,3 +145,6 @@ def argument_error!(message)
     end
   end
 end
+
+# require cookies options
+require "http/options/cookies"
```
zanker: This seems like it should be in `lib/http.rb` with all the other requires?

zanker:
@sferik when the options tests fail, they return as an array instead of a hash. It's really weird, but yea.

This looks good to me besides the `require` bit (also the failing tests)

ixti:
> Looks like this changes the header format, which is causing tests to fail?

That's not really true. It's just how RSpec formats it. `HTTP::Headers` is neither Hash nor Array. But can be presented as both of them. Internally its an array of `(key, value)` pairs. Specs failed due to I forgot to "update" specs with new default param value...

>  This seems like it should be in lib/http.rb with all the other requires?

Not really. This is an options "internal" require. Options will be eventually completely refactored.

sferik:
> @sferik when the options tests fail, they return as an array instead of a hash. It's really weird, but yea.

@myronmarston Any idea why RSpec is doing this? Seems like an RSpec bug. Shall I open an issue?

ixti:
@zanker re require. I follow rule that each file requires only what it requires. Thus `http` requires `http/options` and `http/options` requires `http/options/cookies` :D But as I said earlier, I'm gonna refactor whole `Options` thingy anyway - so this will become less *wtf*-ish :D. 

@myronmarston @sferik if I can somehow help in debugging this - ping me - will be happy to chat on IRC. That weirdly failing bit (in fact it's not failing it's just highlights it improperly - in that spec only one field was actually mismatching (`:cookies`)) happens with instance of: https://github.com/httprb/http.rb/blob/master/lib/http/headers.rb

myronmarston:
> @myronmarston Any idea why RSpec is doing this? Seems like an RSpec bug. Shall I open an issue?

According to @ixti, the headers object is not a hash, and given that RSpec doesn't know it can present it as one.  [It is enumerable](https://github.com/httprb/http.rb/blob/3a06852cad71d8611d2a17fdfca8da3cacf84ed5/lib/http/headers.rb#L10), though, so RSpec enumerates it and presents it as an array (in this case, an array of tuples).

I'd say that RSpec is working as designed given that it's not actually a hash but is enumerable.  If you've got a suggestion for how RSpec could improve the output here please do open an issue.

ixti:
@myronmarston @sferik After I have tried to prepare a visual explanation of what I'm concerned about - I have figured out that somehow it seems like it's us who weirdly modifies headers... Following formats diff as expected:

``` ruby
RSpec.describe "test issue with formatting" do
  let(:h) { HTTP::Headers.coerce "Accept" => "*/*" }
  let(:a) { { :foo => 1, :bar => h.dup } }
  let(:b) { { :foo => 2, :bar => h.dup } }

  specify { expect(h).to eq h.dup }
  specify { expect(a).to eq b }
end
```

ixti:
@myronmarston Thanks for looking at this. It's not an RSpec issue indeed.

zanker:
@ixti ah oops, you're just loading it in. That's kind of an odd way to organize it, generally you would use an `include` or something of that nature instead, since this is basically monkeypatching our own code.

If you plan to refactor it soon, can we tag it with a `TODO: fix when we refactor options (###)` please?

ixti:
@zanker not sure how include is better in this place, but I have no objections against tagging it.

PR #234, opened 2015-05-31
Encoding updates

### LINKED ISSUE #224: Determine string encoding based on Content-Type
See discussion: https://github.com/httprb/http.rb/commit/1ab394f06da8b0ce5b9e26b21e07460351677277#commitcomment-11136186

Connorhd:
Should resolve #224

Adds an encoding option and sets the body encoding based on this order of priorities:
1. encoding option
2. charset in content-type
3. UTF-8 if mime-type starts with 'text/' (I believe the spec says ISO-8859-1, but I think UTF-8 is more commonly assumed)
4. Binary

### DIFF
--- a/lib/http/response.rb
+++ b/lib/http/response.rb
@@ -31,12 +31,18 @@ module HTTP
     # @return [URI, nil]
     attr_reader :uri
 
-    def initialize(status, version, headers, body, uri = nil) # rubocop:disable ParameterLists
-      @version = version
-      @body    = body
-      @uri     = uri && HTTP::URI.parse(uri)
-      @status  = HTTP::Response::Status.new status
-      @headers = HTTP::Headers.coerce(headers || {})
+    def initialize(status, version, headers, connection_or_body, encoding = nil, uri = nil) # rubocop:disable ParameterLists
+      @version  = version
+      @uri      = uri && HTTP::URI.parse(uri)
+      @status   = HTTP::Response::Status.new status
+      @headers  = HTTP::Headers.coerce(headers || {})
+      @encoding = encoding
+
+      if connection_or_body.is_a? HTTP::Connection
+        @body = Response::Body.new(connection_or_body, resolved_encoding)
+      else
+        @body = connection_or_body
+      end
     end
 
     # @!method reason
@@ -119,6 +125,16 @@ module HTTP
     # @return [HTTP::Response::Caching]
     def caching
       Caching.new self
+    end
+
+    private
+
+    # Work out what encoding to assume for the body
+    def resolved_encoding
+      return @encoding if @encoding
+      return charset if charset
+      return Encoding::UTF_8 if mime_type && mime_type.start_with?("text/")
+      Encoding::BINARY
     end
   end
 end
--- a/lib/http/response.rb
+++ b/lib/http/response.rb
@@ -142,7 +142,6 @@ module HTTP
     def resolved_encoding
       return @encoding if @encoding
       return charset if charset
-      return Encoding::UTF_8 if mime_type && mime_type.start_with?("text/")
       Encoding::BINARY
     end
   end
--- a/lib/http/response.rb
+++ b/lib/http/response.rb
@@ -31,17 +31,26 @@ module HTTP
     # @return [URI, nil]
     attr_reader :uri
 
-    def initialize(status, version, headers, connection_or_body, encoding = nil, uri = nil) # rubocop:disable ParameterLists
-      @version  = version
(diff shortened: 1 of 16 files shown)


lib/http/response.rb
```diff
@@ -31,12 +31,18 @@ class Response
...
-      @status  = HTTP::Response::Status.new status
-      @headers = HTTP::Headers.coerce(headers || {})
+    def initialize(status, version, headers, connection_or_body, encoding = nil, uri = nil) # rubocop:disable ParameterLists
+      @version  = version
+      @uri      = uri && HTTP::URI.parse(uri)
+      @status   = HTTP::Response::Status.new status
+      @headers  = HTTP::Headers.coerce(headers || {})
+      @encoding = encoding
+
+      if connection_or_body.is_a? HTTP::Connection
```
Connorhd: This didn't seem great, but I couldn't see a cleaner way to do it nicely.

lib/http/response.rb
```diff
@@ -31,12 +31,18 @@ class Response
     # @return [URI, nil]
     attr_reader :uri
 
-    def initialize(status, version, headers, body, uri = nil) # rubocop:disable ParameterLists
-      @version = version
-      @body    = body
-      @uri     = uri && HTTP::URI.parse(uri)
-      @status  = HTTP::Response::Status.new status
-      @headers = HTTP::Headers.coerce(headers || {})
+    def initialize(status, version, headers, connection_or_body, encoding = nil, uri = nil) # rubocop:disable ParameterLists
```
zanker: Should move this to be a hash, getting a bit long for just straight args.

lib/http/response.rb
zanker: To the above, moving this to a hash means just setting it based on key rather than having to do content type detection.

zanker:
This seems fine. But assumption 3 seems risky, but then again the entire HTTP spec is a mess. Someone else want to weigh in?

ixti:
Although HTTP spec is a mess, if it has declared defaults we should stick
with it. So i would like to see #3 being configurable with default value
set to what RFC says (if it does).

Connorhd:
I think http://www.w3.org/Protocols/rfc2616/rfc2616-sec3.html#sec3.7.1 is the relevant RFC, from what I can google I'm not sure anyone actually honors this though. I think most clients do some kind of character detection (which could also be added to http.rb, but I wanted to avoid it in this change if possible).

zanker:
Looking at Net::HTTP, it appears they just read directly to the string via `socket.read` unless there's compression then it forces ASCII 8BIT.

Connorhd:
Well assuming its binary seems like a good last resort, but I think we can do a bit better. At least step 1 and 2 above seem pretty solid to me.

I could just remove 3 if that makes everyone more comfortable, or I'm happy to make it use ISO-8859-1 if thats preferred.

tarcieri:
I'd be fine with #1 and #2. Hopefully good UTF-8 citizens declare they're using such.

Maybe leave #3 out for now and we can see how it works in practice and follow up on it if it seems necessary?

Connorhd:
Ok, I've removed 3 for now, I think its a good point that utf-8 things are probably more likely to declare their encoding anyway.

lib/http/response.rb
```diff
@@ -31,12 +31,27 @@ class Response
...
+    # @option opts [String] :body
+    # @option opts [String] :uri
+    def initialize(opts)
+      @version  = opts.fetch(:version)
+      @uri      = opts.include?(:uri) && HTTP::URI.parse(opts.fetch(:uri))
+      @status   = HTTP::Response::Status.new opts.fetch(:status)
+      @headers  = HTTP::Headers.coerce(opts.fetch(:headers, {}))
+      @encoding = opts.fetch(:encoding, nil)
+
+      if opts.include?(:connection)
```
ixti: I don't like this magic. Response should receive only `:body`. And this `Body` streamer should be left where it is now.

ixti:
LGTM (with fixing response class)

lib/http/response.rb
Connorhd: I would have done that but to set the encoding in the body I need to have the response object. Do you have a better suggestion for how to organize that?

tarcieri:
@Connorhd looks like this needs a rebase. Want to finish it up before we release 1.0?

ixti:
I'm working on this already :D

ixti:
In other words this *will be* in 1.0 for sure :D

ixti:
After thinking a bit more, all further cleanups can be made in future patch releases, so I'm gonna merge #273 (this rebased PR) as is.

PR #235, opened 2015-06-15
Enable exceptionless NIO when available

zanker:
@tarcieri 

Fixed broken tests


tarcieri:
Finally got around to looking at this. So there's the build error, but I can't deduce what exactly it's failing on... looks like it might be minutia?

zanker:
Yea I've been trying to look at it when I have time and it's weird. Works fine locally :/

tarcieri:
@zanker can you rebase this? I bet that will make the tests pass

PR #237, opened 2015-06-24
a DRY change, addressing code duplication in http/timeout/per_operation

rthbound:
CodeClimate highlighted some duplication in the Timeout::PerOperation class. This change eliminates the code duplication.

### DIFF
--- a/lib/http/timeout/null.rb
+++ b/lib/http/timeout/null.rb
@@ -66,7 +66,7 @@ module HTTP
         if IO.select(nil, [socket], nil, write_timeout)
           retry
         else
-          raise TimeoutError, "Read timed out after #{write_timeout} seconds"
+          raise TimeoutError, "Write timed out after #{write_timeout} seconds"
         end
       end
     end
--- a/lib/http/timeout/null.rb
+++ b/lib/http/timeout/null.rb
@@ -45,9 +45,11 @@ module HTTP
       def write(data)
         @socket << data
       end
+      alias_method :<<, :write
 
+      private
       # Retry reading
-      def __rescue_readable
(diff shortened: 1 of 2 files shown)


lib/http/timeout/null.rb
```diff
@@ -46,6 +46,28 @@ def write(data)
         @socket << data
       end
 
+      # Retry reading
+      def __rescue_readable(&block)
```
zanker: If I recall, it's better to not explicitly reference `&block` since it's less performant than implicit. @tarcieri do you remember? I can't find the article on it anymore.

rthbound: Since the block is a required argument (the only argument, and it's required), I would recommend including it in the method signature.  That said, it'd work just fine as you describe, @zanker .

@tarcieri @zanker : Let me know what you'd prefer and I'll change/not-change accordingly :)

Thanks for this library existing! :+1:

tarcieri: Yes, capturing a block as an object allocates memory and is therefore slower. Not sure if it matters here...

rthbound: @zanker I have made the appropriate changes.

ixti:
@rthbound Can you please:

- remove `__` prefix from method names
- make those methods private -- they are not part of public API

ixti:
@zanker most likely you were trying to find this one: https://www.omniref.com/ruby/2.2.0/symbols/Proc/yield?#annotation=4087638&line=711

rthbound:
@zanker: I would be glad to. Done.

ixti:
Thanks!

PR #247, opened 2015-08-18
Fixed exceptionless NIO handling for EOF

zanker:
:(

So in true Ruby tradition. Exceptionless NIO has no real documentation, I misunderstood a patch I was reading originally and thought it returned `:eof` on EOF. Apparently it returns nil (super obvious right?)

I just looked again and found: https://bugs.ruby-lang.org/issues/11358 and reading _that_ patch shows returning a nil is correct.

This won't happen in general use, but it's still not great.


tarcieri:
:thumbsup:

Issue #251, opened 2015-09-03
Responses with more than the buffer break


zanker:
If we get more than the buffer size of data (16,384), it appears HTTP freezes and just time outs rather than returning everything. It was easy enough to test with a server echoing what you sent it:

`HTTP.post("http://foobar.com", "A" * 17_000)` and it'll time out (when enabled). Need to dig into this more

tarcieri:
Fixed in #251

PR #252, opened 2015-09-06
Fixed chunk writing if the data is bugger than our buffer (#251)

zanker:


zanker:
https://github.com/httprb/http/issues/251

cc @tarcieri @ixti @sferik 

The failure mode for this is it either times out, or hangs forever. So apparently nobody has ran into this or has and hasn't reported it.

tarcieri:
One related-looking test failure?

Would be nice if we had a better strategy for write buffering in general. I'm pretty sure we run afoul of the pathological case for Nagle where do a small write immediately followed by another write.

zanker:
:( socket mocking. I want to redo the test suite somepoint and pull in the one we use for our internal stuff, since that has a lot of edge case handling already too.

Issue should be fixed

tarcieri:
LGTM :+1:

lib/http/request/writer.rb
```diff
@@ -65,24 +68,35 @@ def send_request_header
...
         end
       end
 
       private
 
+      def write(data)
+        while data.present?
+          length = @socket.write(data)
+          if data.length > length
+            data = data[length..-1]
```
tarcieri: This will allocate every time. I wonder if it would be possible to use `#slice!` here to reduce allocations. Or maybe we just need real `ByteBuffers` :confused:

zanker: We call this using `@body` as is, so we probably can't use `slice!`, unless we call `dup` beforehand.

Would need to benchmark to figure out if the optimization works out.

tarcieri: Since Ruby strings are CoW anyway, this usage is probably efficient

PR #255, opened 2015-09-07
remove #present? method that’s not on String

presidentJFK:
it is available on String when running the test suite, because
  active support is a dependency of a test env only gem
  “certificate_authority”


presidentJFK:
Oops! Looks like someone beat me to it. Closing this as duplicate of #254

zanker:
Ack sorry. Blanked on `present?` not being part of core, didn't realize we have a gem pulling in active_support already :(

PR #261, opened 2015-10-09
Hide implementation details

kwando:
Dont leak implementation details error from the connection class.

```SocketError``` and ```SystemCallError``` is rescued and reraised as a ```HTTP::Error``` instead.

Maybe there should be a ```HTTP::ConnectionError < HTTP::Error``` class.

### DIFF
--- a/lib/http/connection.rb
+++ b/lib/http/connection.rb
@@ -35,6 +35,9 @@ module HTTP
       send_proxy_connect_request(req)
       start_tls(req, options)
       reset_timer
+
+    rescue SocketError, SystemCallError => ex
+      raise HTTP::Error.new(ex)
     end
 
     # @see (HTTP::Response::Parser#status_code)
--- a/lib/http/errors.rb
+++ b/lib/http/errors.rb
@@ -16,4 +16,7 @@ module HTTP
 
   # Header name is invalid
   class InvalidHeaderNameError < Error; end
+
+  # Something unexpected happened, probably a bug
+  class UnexpectedError < Error;end
 end
--- a/lib/http/response/status.rb
+++ b/lib/http/response/status.rb
@@ -111,6 +111,7 @@ module HTTP
       end
 
       def __setobj__(obj)
+        raise HTTP::UnexpectedError.new('expected object to respond_to #to_i') unless obj.respond_to?(:to_i)
         @code = obj.to_i
       end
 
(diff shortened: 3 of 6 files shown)

### CODE REVIEW

**Comment by zanker** [comment]
Could you give some context on why you want to wrap some of the lower level errors? I'm not necessarily against it, but having the original error class can be helpful for various error programs. Since a ECONNREFUSED can be different from the other various ECONN type errors.

**Comment by kwando** [comment]
If I wanna catch all errors that ```http``` is throwing I essentially have to rescue from ```Exception```, which is a bad for a number of reasons. Or I need to read the source code to figure out which exceptions this lib might throw, since the documentation is not telling me.

Have a typo in a domain or a host that is just not reachable is pretty well inside the core domain of a HTTP lib hence I think those exceptions should be rescued and reraised to something decending from ```HTTP::Error```.

**Comment by tarcieri** [comment]
What error are you observing that isn't descended from `StandardError`?

In general `rescue => ex` (sans `Exception`) should cover all cases.

Are you seeing a `Timeout::Error` or something?

**Comment by kwando** [comment]
HTTP.get("http://127.0.0.1:3456") 

Where there is nothing responding on 3456 on localhost

**Comment by tarcieri** [comment]
```
[1] pry(main)> require 'http'
=> true
[2] pry(main)> begin
[2] pry(main)*   HTTP.get("http://127.0.0.1:3456")
[2] pry(main)* rescue => ex
[2] pry(main)*   puts "Uhoh, got a #{ex.class}: #{ex}"
[2] pry(main)* end
Uhoh, got a Errno::ECONNREFUSED: Connection refused - connect(2) for "127.0.0.1" port 3456
=> nil
[3] pry(main)> Errno::ECONNREFUSED.ancestors
=> [Errno::ECONNREFUSED,
 SystemCallError,
 StandardError,
 Exception,
 Object,
 JSON::Ext::Generator::GeneratorMethods::Object,
 PP::ObjectMixin,
 Kernel,
 BasicObject]
```

So, no?

kwando:
Oh, sorry my bad for not getting my facts straight. I believed Errno::* errors decended from the Exception.

So what do you think about catching both of those errors and reraise them as a HTTP::ConnectionError?

tarcieri:
Having one class that wraps up all of the network-related errors might be nice.

ixti:
I like the idea of `HTTP::ConnectionError < HTTP::Error`

ixti:
I have rebased and fixed that branch of @kwando 
Will open a new PR.

PR #262, opened 2015-10-09
rspec raise_error wants to know which type of error it is expecting.

kwando:
Also added a HTTP::UnexpectedError which can be used to indicate that something went wrong.

### DIFF
--- a/lib/http/errors.rb
+++ b/lib/http/errors.rb
@@ -16,4 +16,7 @@ module HTTP
 
   # Header name is invalid
   class InvalidHeaderNameError < Error; end
+
+  # Something unexpected happened, probably a bug
+  class UnexpectedError < Error;end
 end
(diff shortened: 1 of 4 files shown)


tarcieri:
Can't say I'm really wild about the name `UnexpectedError`

kwando:
Do you have a better name? ImplementationError?

tarcieri:
I think I'd rather have meaningful exceptions

kwando:
DoesNotRespondToToIntError? Anyway, I do not feel very strongly about this. It was just a way for me to make rspec stop complaining.

kwando:
AFAIK that error would only be occur if there is a bug in HTTP.

ixti:
This can be done a bit different way. Thanks for bringing this on. I'll take a look.

PR #268, opened 2015-11-25
Remove Celluloid::IO docs

tarcieri:
The timeout backend is broken/nonexistent. Until it's implemented, we shouldn't advertise Celluloid::IO support.

Users of Celluloid who want parallel requests can use a thread pool.

### DIFF
--- a/README.md
+++ b/README.md
@@ -276,30 +276,6 @@ HTTP.persistent('http://en.wikipedia.org') do |http|
 end
 ```
 
-### Celluloid::IO Support
-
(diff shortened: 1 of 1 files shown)


vasilakisfil:
So should that also be removed? https://github.com/httprb/http/blob/master/examples/parallel_requests_with_celluloid.rb

ixti:
@vasilakisfil seems so. But I would wait @tarcieri on this.

tarcieri:
I removed it

PR #271, opened 2015-12-12
Hide implementation details

ixti:
This is renewed #261

### DIFF
--- a/lib/http/connection.rb
+++ b/lib/http/connection.rb
@@ -35,6 +35,9 @@ module HTTP
       send_proxy_connect_request(req)
       start_tls(req, options)
       reset_timer
+
+    rescue SocketError, SystemCallError => ex
+      raise HTTP::Error.new(ex)
     end
 
     # @see (HTTP::Response::Parser#status_code)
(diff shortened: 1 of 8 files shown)


ixti:
@zanker @tarcieri personally [I have no strong feelings one way or the other](
https://www.youtube.com/watch?v=ussCHoQttyQ), so it's up to you guys if we want this PR to be merged in or not. This is a breaking changes so if we want this to be in - we should do this like now.

sferik:
:+1:

zanker:
I'm a fan, but I'm concerned about breaking changes. Can we leave this for a 2.0?

sferik:
> I'm a fan, but I'm concerned about breaking changes. Can we leave this for a 2.0?

IMHO, if we all agree this is better, we should make this change now, before 1.0 is released, rather than kicking the can down the road and causing breakages in the future when (presumably, if everything goes according to plan) this library has even more users.

zanker:
Whoops, I forgot we only had a pre not an actual 1.0. I'm good with making the change now then.

tarcieri:
This looks good to me, although requires a bit of `#cause` spelunking now if you want to figure out where the original exception occurred (well, I guess we had some rethrows before with different classes, but...)

PR #274, opened 2015-12-18
add socket_factory option

nerdrew:
Add an option for custom socket creation, e.g.

```
HTTP::Client.new(socket_factory: lambda { |host, port|
  sock = TCPSocket.open(host, port)
  sock.setsockopt(Socket::Option.int(:INET, Socket::IPPROTO_TCP, Socket::TCP_NODELAY, 1))
  sock
})
```

### DIFF
--- a/lib/http/connection.rb
+++ b/lib/http/connection.rb
@@ -30,7 +30,11 @@ module HTTP
       @parser = Response::Parser.new
 
       @socket = options.timeout_class.new(options.timeout_options)
-      @socket.connect(options.socket_class, req.socket_host, req.socket_port)
+      if options.socket_factory
+        @socket.factory_connect(options.socket_factory, req.socket_host, req.socket_port)
+      else
+        @socket.connect(options.socket_class, req.socket_host, req.socket_port)
+      end
 
       send_proxy_connect_request(req)
       start_tls(req, options)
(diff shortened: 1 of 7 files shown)


tarcieri:
Rubocop isn't happy :(

This seems ok, although it might be nice to have a separate API specifically for disabling Nagle, and one that uses `HTTP::Options`, e.g.

`HTTP.nodelay(true)...`

or

`HTTP.socketopts(nodelay: true)...`

ixti:
:+1: from me for API. But I see it a bit different:

``` ruby
def nodelay
  socket_options :delay => false
end

def socket_options(options)
  # ...
end
```

ixti:
In fact I think we should have only `socket_options()` API (no sugar).

tarcieri:
@ixti so I was talking to @nerdrew about that, and asking what other socket options we'd actually like to support, and the answer is really: none.

I have a new API suggestion:

```
HTTP.nodelay
```

Done!

zanker:
My name is Zach Anker, and I approve of Tony's message.

zanker:
My name is Zach Anker, and I approve of Tony's message.

ixti:
@tarcieri if there's no other socket options we would like to support, then `nodelay` works for me as well! :D

nerdrew:
Ok, alternative `nodelay` PR: https://github.com/httprb/http/pull/275/files

PR #275, opened 2015-12-18
add nodelay option

nerdrew:


tarcieri:
:+1: 

What's RuboCop complaining about w.r.t. ParameterLists? You can probably just disable that in `.rubocop.yml`

ixti:
https://github.com/httprb/http/blob/master/.rubocop.yml#L27 -- let's bump up this to `5` with comment that we should try to lower that down later down the road

PR #276, opened 2015-12-18
Set default HTTP::Response::Body encoding to Encoding::BINARY

tarcieri:
Otherwise we break Webmock


tarcieri:
Arguably this could be `Encoding::UTF_8`, but `Encoding::BINARY` seems like a safer default

PR #283, opened 2015-12-26
Use io/wait on platforms where it's available

tarcieri:
IO.select is an expensive call, because it has to build an fd_set (large bitfield) from an array each time we invoke it.

We're just interested in IO readiness, so on Ruby 2.0+ we can use the io/wait API instead, which adds wait_readable/wait_writable to IO objects.

cc @zanker @nerdrew

### DIFF
--- a/lib/http/timeout/per_operation.rb
+++ b/lib/http/timeout/per_operation.rb
@@ -59,14 +59,17 @@ module HTTP
         # Read data from the socket
         def readpartial(size)
           loop do
-            result = socket.read_nonblock(size, :exception => false)
+            result = rescue_readable do
+              socket.read_nonblock(size, :exception => false)
+            end
+
             if result.nil?
               return :eof
             elsif result != :wait_readable
               return result
             end
 
-            unless IO.select([socket], nil, nil, read_timeout)
+            unless socket.to_io.wait_readable(read_timeout)
               fail TimeoutError, "Read timed out after #{read_timeout} seconds"
             end
           end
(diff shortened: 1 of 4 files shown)


lib/http/timeout/per_operation.rb
```diff
@@ -59,14 +59,17 @@ def write(data)
         # Read data from the socket
         def readpartial(size)
           loop do
-            result = socket.read_nonblock(size, :exception => false)
+            result = rescue_readable do
```
zanker: This doesn't make sense, there's nothing to rescue since it's not using exceptions?

lib/http/timeout/per_operation.rb
```diff
@@ -75,10 +78,13 @@ def readpartial(size)
         # Write data to the socket
         def write(data)
           loop do
-            result = socket.write_nonblock(data, :exception => false)
+            result = rescue_writable do
```
zanker: Same here on rescue

zanker:
LGs besides those minor things

lib/http/timeout/per_operation.rb
tarcieri: If you recall from before, JRuby silently discards `:exception => false` on SSL sockets and still throws exceptions. I can add a comment to that effect.

tarcieri: Added some clarifying comments in fbb57b5

zanker: Ohh right fsss. Yea if you can put a comment please.

PR #284, opened 2015-12-28
Updated links reference the right issues / PRs

kwando:
Some of the links in the changelog was not referencing the correct issues.


sferik:
Thanks!

PR #287, opened 2015-12-29
Document HTTP.persistent without a block in README

vassilevsky:
Hello 👻

It wasn't until I dug deep into the source code that I realised there was a way to fire multiple requests over the same connection *outside of a block*. I thought that this deserves a mention in `README`.

I've used the `c` variable name because:
- it actually refers to an instance of `Client`
- semantically it is best to call it a connection
- the actual name is irrelevant and can be anything and I wanted it to get out of the way

Hope this helps!

### DIFF
--- a/README.md
+++ b/README.md
@@ -270,14 +270,36 @@ given resource.
 
 ### Reuse HTTP connection: HTTP Keep-Alive
 
-If you have many successive requests against the same host, you better want to
-reuse the same connection again and again:
+If you need to make many successive requests against the same host, you can create
+one persistent connection to the host:
+
+```ruby
+c = HTTP.persistent('http://api.icndb.com')
+```
+
+Then you can issue multiple requests on this connection:
+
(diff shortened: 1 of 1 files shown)


tarcieri:
Maybe `http` as the variable name instead of `c`?

ook:
May I suggest to put the `c` / `http` `close` into a `ensure` block?

ixti:
I'm improving this PR atm (adding my patch on top of it).

ixti:
Re `c` variable name. I don't find it a good idea.

1. it's easier to read code with more meaningful variable name in this particular case 
2. i disagree that semantically it's best to call it connection

With that being said, I would agree on `client` variable name, but not `c`.
After all I have changed that to `http` as it makes a bit more sense IMO (at least in those code snippets).

tarcieri:
@ixti :+1:

PR #288, opened 2015-12-31
Add Spaces Around Brackets

mehdi-farsi:
a small step for a man a giant leap for mankind


zanker:
Sorry, but I don't think this is a very helpful change and mostly just adds noise.

mehdi-farsi:
OK ;)

ixti:
TBH I prefer spaces within brackets myself as well :D
In any case it worth changing only when we will be revising our code-style rules in `.rubocop.yml` IMO

PR #294, opened 2016-01-10
Write chunks of Enumerable bodies in a single system call

tarcieri:
Concatenate the header and body prior to the write call. I believe most Rubies' String implementation will do this efficiently (paricularly with <<) and without excess copying.

Fewer system calls should be more efficient.


zanker:
LGs

PR #296, opened 2016-01-16
Update to RuboCop 0.36

tarcieri:
Includes a lot more cops for frozen string literals and constants which seem nice, particularly on Ruby 2.3+

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -61,17 +61,8 @@ Style/HashSyntax:
 Style/Lambda:
   Enabled: false
 
-Style/SingleSpaceBeforeFirstArg:
-  Enabled: false
-
 Style/SpaceAroundOperators:
-  MultiSpaceAllowedForOperators:
-    - "="
-    - "=>"
-    - "||"
-    - "||="
-    - "&&"
-    - "&&="
+  AllowForAlignment: true
 
 Style/SpaceInsideHashLiteralBraces:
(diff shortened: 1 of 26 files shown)


.rubocop.yml
```diff
@@ -61,17 +61,8 @@ Style/HashSyntax:
 Style/Lambda:
   Enabled: false
 
-Style/SingleSpaceBeforeFirstArg:
```
tarcieri: This doesn't exist anymore?

lib/http/chainable.rb
```diff
@@ -206,7 +205,7 @@ def basic_auth(opts)
       user = opts.fetch :user
       pass = opts.fetch :pass
 
-      auth("Basic " << Base64.strict_encode64("#{user}:#{pass}"))
+      auth("Basic " + Base64.strict_encode64("#{user}:#{pass}"))
```
tarcieri: "Basic" is now frozen here (see `# frozen_string_literal: true` at the top)

.rubocop.yml
ixti: It's now called: `Style/SpaceBeforeFirstArg`

ixti:
:+1:

.rubocop.yml
ixti: In any case I believe it's fine to remove it...

PR #301, opened 2016-02-01
Remove exception wrappers from per_operation.rb

tarcieri:
I started out reverting the changes to this file back to the version
@ixti bisected as being the last good one.

I left in the *actual* code that replaced IO.select with io/wait. And
I seem to have come upon the same thing @ixti did here:

https://github.com/httprb/http/commit/9e1e100d89e365600010859ea8c87400a766c29b

I'll have to check the behavior on JRuby, and am uncertain what sequence
of events causes the rescue_readable call to be an issue, but I think this
both lets us keep io/wait and at least has the same semantics as before.

### DIFF
--- a/lib/http/timeout/per_operation.rb
+++ b/lib/http/timeout/per_operation.rb
@@ -59,12 +59,7 @@ module HTTP
         # Read data from the socket
         def readpartial(size)
           loop do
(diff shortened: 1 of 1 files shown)


tarcieri:
Guess I'm gonna merge this :shipit:

PR #302, opened 2016-02-07
Added update_timeout API to reconfigure timeouts on the fly

zanker:
This is silly... and well that's about it, but I can't see a better way of doing it so I think it's known silly? I think this is a legitimate use case, but I can't share the code since it's one of our internal gems.

We have a S2S library, two of the main features are it uses a connection pool + conn reuse, and the other is that you can dynamically reconfigure timeouts. Eg, use a short timeout for primary key lookup APIs but a long timeout for search APIs.

We want to share the same pool between the two because the SSL handshake is expensive. We can't do that with what we have today in HTTP.rb, since we also need to reconfigure the socket when the timeout changes.

Currently we do it with `instance_variable_get`, but I'd obviously prefer not to rely on that for longer term. You could make a case that we might want a more generalized way of reconfiguring, but I think timeout is the 99% use case where you want to keep the same socket but change someting.

Thoughts?

### DIFF
--- a/lib/http/client.rb
+++ b/lib/http/client.rb
@@ -91,6 +91,14 @@ module HTTP
       @state = :clean
     end
 
+    # Allow on-the-fly reconfiguration of timeouts
+    def update_timeout(timeout_opts)
+      # Ensure any future defaults use the updated timeouts
+      @default_options.send(:timeout_options=, timeout_opts)
+      # If have any connections open, make sure they're updated
+      @connection.update_timeout(timeout_opts) if @connection
+    end
+
     private
 
     # Verify our request isn't going to be made against another URI
--- a/lib/http/connection.rb
+++ b/lib/http/connection.rb
@@ -139,6 +139,12 @@ module HTTP
       !@conn_expires_at || @conn_expires_at < Time.now
     end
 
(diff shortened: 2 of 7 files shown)


tarcieri:
This seems like a one-off solution for the issue of timeouts when grabbing a connection from a connection pool. It'd be nice if there were a more general solution to this problem.

We can just solve cases like this as they crop up, but the intended goal of the chaining API was to be stateless. Perhaps we can abstract configuration overrides like this into a context object that can get passed in per-request?

zanker:
How many options do you really need to dynamically override though? If you're assuming it's a connection pool you only really are going to toggle cookies/timeouts and that's about it.

ixti:
At least these three should be configurable:

- headers (cookies)
- timeouts
- redirect follow strategy (on / off / max-hops etc)

:+1: on idea of context.

zanker:
Hrm, maybe something like `reconfigure` on the client that gives you a form of chainable that modifies the `@default_options`? And is also filtered to headers/cookies/timeouts/redirects?

tarcieri:
I think #306 (`HTTP::Session`) is the way forward for this. Closing this for now.

PR #305, opened 2016-02-14
Factor docs in README into Wiki

tarcieri:
All of these have been moved into the Wiki where they can be more
easily discovered and collaborated on.


ixti:
:tada:

PR #314, opened 2016-03-05
fixes #312

### LINKED ISSUE #312: Validate charset before force_encoding
`HTTP::Response::Body` will force a response's contents to whatever encoding was specified in the `Content-Type` header. That encoding should be validated before applied since some common encodings don't appear in ruby's `Encoding` lookup table. 

For example, 

```ruby
require 'sinatra'

get '/', provides: 'text/html; charset=utf8' do 
	"Hello!"
end
```

Pulling that page using `HTTP` ...

```ruby
require 'http'

HTTP.get('http://localhost:4567').to_s
```
```
/Users/kyle/.rvm/gems/ruby-2.2.1/gems/http-1.0.2/lib/http/response/body.rb:40:in `force_encoding': unknown encoding name - utf8 (ArgumentError)
	from /Users/kyle/.rvm/gems/ruby-2.2.1/gems/http-1.0.2/lib/http/response/body.rb:40:in `to_s'
	from client.rb:3:in `<main>'
```

kylekyle:
Validate encoding in `Content-Type` before `force_encoding` response contents.

### DIFF
--- a/lib/http/content_type.rb
+++ b/lib/http/content_type.rb
@@ -20,8 +20,13 @@ module HTTP
       # :nodoc:
       def charset(str)
         md = str.to_s.match CHARSET_RE
-       	md = md && md[1].to_s.strip.gsub(/^"|"$/, "")
-				md if Encoding.find(md) rescue nil
+        md &&= md[1].to_s.strip.gsub(/^"|"$/, "")
+        
+        begin 
+          md if Encoding.find(md)
+        rescue ArgumentError
+          nil
+        end
       end
     end
   end
--- a/lib/http/content_type.rb
+++ b/lib/http/content_type.rb
@@ -21,7 +21,7 @@ module HTTP
       def charset(str)
         md = str.to_s.match CHARSET_RE
        	md = md && md[1].to_s.strip.gsub(/^"|"$/, "")
-				Encoding.find md rescue nil
+				md if Encoding.find(md) rescue nil
       end
     end
   end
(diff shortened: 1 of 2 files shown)


tarcieri:
This is the wrong place to fix this: it's changing the type stored in `HTTP::Content#charset` from a `String` to an `Encoding`. While we could debate the relative merits of what type should've been used, this is already a post-1.0 library and this is a breaking API change (as you can see, it's also causing test failures)

kylekyle:
Whoops! Sorry about that. It should always return a `String` now.

tarcieri:
Still failing RuboCop due to the inconsistent alignment (also hard tabs). Make sure to run `bundle exec rake` before committing.

kylekyle:
I've never seen RuboCop before. Pretty sweet. It should be passing now.

tarcieri:
The tests are failing again

tarcieri:
Also I still think this is the wrong place to solve this: this information should still be available even if the encoding isn't known to Ruby.

Instead I'd suggest handling the `force_encoding` exception if the encoding isn't known, so it doesn't bubble up.

kylekyle:
That's a good point. The charset doesn't change just because Ruby doesn't recognize it as a `String` encoding. Okay, I moved the fix to right before the force_encoding call.

tarcieri:
Ok, this looks fine

tarcieri:
Thanks!

PR #315, opened 2016-03-05
Issue 313

kylekyle:
Adds MIME adapters for `text/plain`, `application/xml` / `text/xml`, and `text/html`.


tarcieri:
I'm definitely -1 on adding `nokogiri` as a runtime dependency. That seems like something a user could optionally pull in with an explicit require `http/mime_type/xml`

ixti:
Thanks for your efforts, but the only adapter I can think of being OK to be available out of the box is JSON. I would extend that list to YAML, but because of of some reasons (e.g. custom types) I think it's not good to be part of the core. XML, text, YAML are pretty specific in my opinion. For example at SensorTower we override default JSON adapter to use different JSON parser/dumper.

Issue #317, opened 2016-03-15
Prevent header renaming?


mhenrixon:
How can I prevent header renaming? I need to provide headers with an underscore and I can't find a way to achieve this.

ixti:
Can you explain why you might need this?

ixti:
You're not the first one who was asking this question, so I'm pretty concerned if this is really needed feature (we can probably allow pass header names without normalization)

mhenrixon:
For me it turned out it wasn't needed I just thought it was when I got different errors from CURL when I changed the name of the headers.

ixti:
Ah. OK. Thanks.

tarcieri:
I think in general the way we process headers could use some work (and was a regression from my previous code). We *always* do duplicate allocations for headers, even if they already match the canonical format.

I think we should check first if they match the canonical format, in which case the duplicate allocations should be skipped. Right now we always check and always allocate.

ixti:
> was a regression from my previous code

Fixed in: #318

PR #318, opened 2016-03-16
Reduce amount of objects upon headers normalization

ixti:
In response to https://github.com/httprb/http/issues/317#issuecomment-197042971

### DIFF
--- a/lib/http/headers.rb
+++ b/lib/http/headers.rb
@@ -11,11 +11,11 @@ module HTTP
     include Enumerable
 
     # Matches HTTP header names when in "Canonical-Http-Format"
-    CANONICAL_HEADER = /^[A-Z][a-z]*(-[A-Z][a-z]*)*$/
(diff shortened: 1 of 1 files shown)


ixti:
@httprb/core I'll merge this down unless any objections

zanker:
Looks fine. Only concern would be looking at our header tests it looks like we don't really test our normalization? Could potentially break something without knowing it.

ixti:
@zanker all methods in https://github.com/httprb/http/blob/master/spec/lib/http/headers_spec.rb have examples with canonical header names and not yet normalized forms.

Generally this patch does not changes anything in logic except it simply avoids creation of new (String|Array) objects when it might be avoided.

mhenrixon:
Thanks for this! I was lucky enough that we already implemented sort of a fix where we support both formats for known keys but this allows us to get rid of some internal hacks eventually.

tarcieri:
:thumbsup: thanks for doing this!

Issue #298, opened 2016-01-19
Frequent HTTP::TimeoutError's (Read) on 1.0.x but not on 0.9.x


javierjulio:
I've been using the 0.9.x releases for a few months now and have since upgraded to 1.0.0 and then 1.0.2. I've noticed several HTTP::TimeoutError's on 1.0.x locally but I get it almost every single time I deploy a gem that runs integration tests on CircleCI (Ubuntu). If I update the two gems I've created that depend on http and downgrade it back to `add_dependency "http", '~> 0.9'` all my integration tests that are making live HTTP requests consistently pass without a single timeout error. At first it seemed that the 1.0.2 release was helping but the timeout errors still occur. I'm trying to debug the issue further but figured in the meantime I'd post this and see if anyone else has had a similar issue when upgrading to 1.0.x. As I learn more I'll report back.

For configuring timeouts I went with a default of `{ write: 2, connect: 5, read: 10 }`. Increasing the read timeout value doesn't help. Note that this is per request. While I don't have stats I know that requests don't take 10 seconds each (just the read part fails). The same timeouts are used for http 0.9.x and all integration tests pass with no frequent timeout errors.

I use `http` gem in the following JSON API client gems: [synapse_payments](https://github.com/javierjulio/synapse_payments) and [paysafe](https://github.com/javierjulio/paysafe).

Using Ruby 2.2.3 in app and each gem using http gem.

zanker:
How long does the request normally take, and how long before it times out? Also what version of Ruby?

There's some changes @tarcieri made around the API calls we use for timeouts, but they should work identical to before. I could believe there's a small rounding change where if you were cutting it close on timeouts, but given you said increasing it doesn't help, it's probably not that.

tarcieri:
@javierjulio can you provide any sort of timeline of the events that are happening and how those don't line up with your described timeout model?

javierjulio:
@zanker I don't know how long requests normally take, since I'm making requests to a sandbox service I'd say about a second or so each. Note that the timeouts are for each request (per operation). I've updated the description with this info and Ruby version (2.2.3). Thanks!

[...]

Issue #297, opened 2016-01-19
HTTP::URI.parse doesn’t set default port


sferik:
The `http` gem [changed](https://github.com/httprb/http/commit/81509bc40379dcc69bad9fa9ca435552947468c6) to how it creates URIs between version 0.7 and 0.8. Versions 0.7 and earlier used the standard `URI` library, while more recent versions use a custom `HTTP::URI`, which descends from `Addressable::URI`.

After upgrading `http` dependency in the `twitter` gem from 0.6 to 0.8 (and now to 1.0), some code that expected the default port to be set is now broken. You can read a detailed write up of [the issue](https://github.com/sferik/twitter/issues/709), with the relevant portion reproduced below:

##### `http` gem 0.7.x and earlier
```ruby
[38] pry(main)> uri = URI('https://stream.twitter.com')
=> #<URI::HTTPS:0x007fa54f8d38e8 URL:https://stream.twitter.com>
[39] pry(main)> uri.port
=> 443
```
##### `http` gem 0.8.x and later
```ruby
[40] pry(main)> uri = HTTP::URI.parse('https://stream.twitter.com').normalize
=> #<Addressable::URI:0x3fd2a57bd650 URI:https://stream.twitter.com/>
[41] pry(main)> uri.port
=> nil
```

@tarcieri @zanker @ixti **How would you feel about adding back the default ports (80 for HTTP and 443 for HTTPS) if an explicit port is not specified?**

Quoting from [RFC 7230, section 2.7.1](https://tools.ietf.org/html/rfc7230#section-2.7.1) on http URI Scheme:

> If the port subcomponent is empty or not given, TCP port 80 (the reserved port for WWW services) is the default.

And [the subsequent section](https://tools.ietf.org/html/rfc7230#section-2.7.2) on the https URI Scheme:

>  All of the requirements listed above for the "http" scheme are also requirements for the "https" scheme, except that TCP port 443 is the default if the port subcomponent is empty or not given…

If we all agree this is the correct behavior, I will work on a patch.

tarcieri:
Definitely sounds like the correct behavior to me, and a bug in Addressable...

sferik:
[...]

PR #319, opened 2016-03-18
Drop Ruby 1.9.x support

ixti:


.travis.yml
```diff
@@ -4,12 +4,12 @@ env:
...
 language: ruby
 rvm:
-  - 1.9.3
   - 2.0.0
   - 2.1
   - 2.2
   - 2.3.0
-  - jruby
+  - jruby-9.0.0.0
+  - jruby-9.0.1.0
```
tarcieri: 9.0.5.0 here perhaps?

ixti: Added 9.0.{4,5}.0 as well... But there are couple of specs failing on 9000+ in general o_O

ixti: In fact it fails on `9.0.{0,1,4}.0` only. `9.0.5.0` works fine indeed.

tarcieri: I think it's fine to just test `9.0.5.0`. Those are just patchlevel changes.

ixti:
@tarcieri done. As we test JRuby 9.0.5.0 only I have chnaged `JRuby 9000+` support declaration in readme to `9.0.5.0`

tarcieri:
Cool. I think I'd like to get #298 squared away first and do one more 1.x release before merging this though.

tarcieri:
:shipit:

tarcieri:
@ixti @sferik @zanker FYI to you all, master is now version 2.x of the gem, which requires Ruby 2.0+.

`1-x-stable` is the support branch for doing 1.x bugfixes. Hopefully there won't be that many.

tarcieri:
I didn't do a gem release yet. @ixti you want to do the honors?

tarcieri:
@ixti actually, before doing a 2.x release do you want to weigh in on #297 ?

tarcieri:
2.0.0 release tracking issue here: #321

PR #320, opened 2016-03-19
Fix timeout regression

tarcieri:
Merging this directly into 1-x-stable so it doesn't interfere with #319


tarcieri:
@javierjulio can you confirm this fixes #298 ?

javierjulio:
@tarcieri confirmed! 100% pass completion rate even on multiple attempts with the test script. Thanks!

Issue #323, opened 2016-03-29
URL with underscores don't work


907th:
HTTP incorrectly converts `sms_service.mysite.com` URL into `sms%5Fservice.mysite.com`. And then my request fails with `HTTP::ConnectionError: failed to connect: getaddrinfo: Name or service not known` error.
PS. curl for the same URL works fine.

907th:
My teammate thinks that this issue should be fixed in the Addressable gem. So he created a [similar issue] (https://github.com/sporkmonger/addressable/issues/233) in that gem.

tarcieri:
Addressable has been the source of a lot of problems, I think. We've discussed removing it (or at least minimizing its usage) in a 2.0 release.

tarcieri:
One thing I should note on this: underscores aren't technically allowed in domain names (only `[a-z0-9\-]`)

907th:
@tarcieri You're right, they are [not allowed](http://tools.ietf.org/html/rfc1034#section-3.5) in DNS names. Thanks!

tarcieri:
I suggest either fixing your URLs or fixing Addressable.

sporkmonger:
@tarcieri I'd actually be really interested in hearing more about where Addressable has been a problem.

tarcieri:
@sporkmonger #297 has most of the discussion. For what it's worth I don't think we're planning on removing it anymore.

Issue #206, opened 2015-04-07
Subtle issue with normalizing URIs 


benlovell:
Since 81509bc we've had issues with using this gem to communicate with elasticsearch. It seems that prior to this change, the request URI was being normalized correctly and thus not including the scheme and host segment. As is evidenced below:

![screen shot 2015-04-07 at 11 25 17](https://cloud.githubusercontent.com/assets/41963/7021404/05c4417c-dd19-11e4-9766-a86a9a587b61.png)

And after 81509bc:

![screen shot 2015-04-07 at 11 26 07](https://cloud.githubusercontent.com/assets/41963/7021411/16f869fa-dd19-11e4-97e7-d8b1b89d4e55.png)

It seems some servers are strict in this regard. Elasticsearch apparently being one of them. :see_no_evil: 

This is easily resolved by `uri.normalize.request_uri` instead of just normalizing the URI, but this causes some issues with proxying that I don't really understand.

ixti:
Thanks for report. My bad. Will fix today.

ixti:
It should not send `schema://hostname` inside `METHOD ...` line.

ixti:
Actually I have misunderstood original code and wiped out things that were essential. Will provide a fix ASAP.

ixti:
Released as `0.8.3`.

benlovell:
:heart:

britishtea:
I ran into an issue today due to this commit.

I often use Charles Proxy to inspect HTTP(S) requests. When proxying HTTPS request to , for example, https://www.google.com, I see requests being made to `https://www.google.comhttps://www.google.com/`. I suspect this is due to the headline, which is `"GET https://google.com/ HTTP/1.1"`.

The following code causes the problem:

```ruby
require "http"

HTTP.via("0.0.0.0", 8888).get "https://www.google.com", :ssl => { :verify_mode => OpenSSL::SSL::VERIFY_NONE }
```

When doing the same thing with a persistent connection to http://www.google.com (plain HTTP) everything works as expected. The headline is `"GET / HTTP/1.1"` in that case.

Is `http` doing the correct thing here? I've not had such issues with other ruby libraries when proxying HTTPS requests.

tarcieri:
@britishtea definitely seems like a bug, possibly related to Addressable?

ixti:
@Connorhd fixed issue reported by @britishtea in https://github.com/httprb/http/commit/2625436bb25ab779e79b4b1ad6f2df9f94e43e48

PR #325, opened 2016-04-03
Use object composition rather than inheritance for HTTP::URI

tarcieri:
The Addressable gem has been the source of a number of bugs and performance problems (e.g. #206, #297, #323)

Perhaps in the future it might be worth avoiding using it where `::URI` would suffice, or getting rid of it entirely, but as things stand `Addressable::URI` is the parent class for `HTTP::URI`, which prevents us from doing that without making a breaking release of the library.

This replaces inheritance with object composition: `Forwardable` is used to delegate an explicit set of methods to `Addressable::URI`. This should make it easier to work around Addressable bugs or
potentially even use `::URI` as the internal representation when URIs are simple enough.

### DIFF
--- a/lib/http/chainable.rb
+++ b/lib/http/chainable.rb
@@ -141,7 +141,7 @@ module HTTP
       return p_client unless block_given?
       yield p_client
     ensure
-      p_client.close unless p_client.nil?
+      p_client.close if p_client
     end
 
     # Make a request through an HTTP proxy
--- a/lib/http/chainable.rb
+++ b/lib/http/chainable.rb
@@ -141,7 +141,7 @@ module HTTP
       return p_client unless block_given?
       yield p_client
     ensure
-      p_client.close
+      p_client.close unless p_client.nil?
     end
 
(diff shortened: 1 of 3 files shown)


tarcieri:
Bonus points: this now fixes #297 

@ixti do you have strong feelings about this? I think this gives us more future flexibility around optimizations like e.g. using `::URI` in the event a URI contains all "basic ASCII" and is already in its normalized form

lib/http/chainable.rb
```diff
@@ -141,7 +141,7 @@ def persistent(host)
       return p_client unless block_given?
       yield p_client
     ensure
-      p_client.close
+      p_client.close unless p_client.nil?
```
ixti: Why not `if p_client` instead?

tarcieri: :thumbsup:

ixti:
I think that's awesome! :+1:

tarcieri:
Sweet

PR #329, opened 2016-04-04
Travis config cleanups

tarcieri:
- Only test master: this prevents duplicate builds for branches vs a PR
- Add whitespace
- Move language/sudo directives to the top of the file

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,8 +1,12 @@
+language: ruby
+sudo: false
+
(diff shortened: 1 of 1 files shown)


tarcieri:
My main motivation here is when both the branch and master merge are tested, it doubles the amount of time it takes for the tests to complete (more or less).

Just checking the PR against a potential master merge seems sufficient to me.

PR #330, opened 2016-04-06
Support custom CONNECT headers (both request and response) during HTTPS proxy requests

smudge:
Note: Just throwing this idea out there -- I implemented these changes for my own use case, but I'm happy to rework the implementation if it's not deemed suitable for merging. Let me know.

### The Issue

Some proxies accept custom headers, for things like switching IPs, excluding certain IPs, or specifying custom behaviors (like retrying requests when certain codes are returned). Over normal HTTP, these can be sent along with the rest of your headers, and the proxy will filter the headers before passing the request along. Likewise, the proxy might inject a custom header into the response (such as one indicating the IP your request was passed through).

**However**, everything changes over a secured connection. When connecting to a proxy over HTTPS, this gem makes use of HTTP tunneling via the 'CONNECT'  method. The underlying request and response are encrypted. As a result, the proxy is unable to read any headers you send with the underlying request or inject headers into the underlying response.

So adding custom proxy-specific headers to a HTTPS request looks something like this:

```
CONNECT example.com:1234 HTTP/1.1
Proxy-Authorization: Digest [your auth digest]
X-My-Custom-Header: Yes

<Your Underlying Request Goes Here>
```

And a response from the proxy (with a custom header) looks like this:

```
HTTP/1.1 200 Connection established
X-Custom-Proxy-Response: Success

<The Underlying Response Appears Here>
```

**The problem** is that, aside from `Proxy-Authorization`, this gem does not support adding custom headers (like `X-My-Custom-Header` and `X-Custom-Proxy-Response` above).

### The Change(s)

This PR adds a parameter to the `via` method which allows you to specify custom headers intended for use by the proxy:

```ruby
HTTP.via("proxy-hostname.local", 8080, 'x-retry-if' => 502)
    .get("https://example.com/resource")
HTTP.via("proxy-hostname.local", 8080, USERNAME, PASSWORD, 'x-use-ip' => '123.456.11.11')
    .get("https://example.com/resource")
```

*Note that if the underlying request is 'http' instead of 'https', there is no `CONNECT` tunneling, so the proxy-specific headers will be merged into the requests's primary headers.*

Similarly, if the proxy returns any custom headers (as shown above), they are made available as follows:

```ruby
response = HTTP.via("proxy-hostname.local", 8080).get("https://example.com/resource")

response.proxy_headers
# => { 'x-ip-used' => '123.456.42.42' }
```

*Note: I considered transparently merging these into the main request headers, but I worried that this would cause strange behavior. Either a proxy could overwrite a header you expect from the remote server, or the remote server could overwrite a header you expect from the proxy, and I wasn't sure how to handle these cases. If you have any suggestions I'd be happy to rework the behavior.*

### Questions?

Does all of this make sense? Let me know if you have any questions, ideas, suggestions, etc.

### DIFF
--- a/lib/http/chainable.rb
+++ b/lib/http/chainable.rb
@@ -153,8 +153,10 @@ module HTTP
       proxy_hash[:proxy_port]     = proxy[1] if proxy[1].is_a?(Integer)
       proxy_hash[:proxy_username] = proxy[2] if proxy[2].is_a?(String)
       proxy_hash[:proxy_password] = proxy[3] if proxy[3].is_a?(String)
+      proxy_hash[:proxy_headers]  = proxy[2] if proxy[2].is_a?(Hash)
+      proxy_hash[:proxy_headers]  = proxy[4] if proxy[4].is_a?(Hash)
 
-      raise(RequestError, "invalid HTTP proxy: #{proxy_hash}") unless [2, 4].include?(proxy_hash.keys.size)
+      raise(RequestError, "invalid HTTP proxy: #{proxy_hash}") unless (2..5).cover?(proxy_hash.keys.size)
 
       branch default_options.with_proxy(proxy_hash)
     end
--- a/lib/http/client.rb
+++ b/lib/http/client.rb
@@ -68,12 +68,13 @@ module HTTP
       end
 
       res = Response.new(
-        :status     => @connection.status_code,
-        :version    => @connection.http_version,
-        :headers    => @connection.headers,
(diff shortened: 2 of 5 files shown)


zanker:
Seems reasonable to me, @tarcieri @ixti @sferik ?

tarcieri:
:+1:

tarcieri:
This will go out in 2.0.0 when we eventually release it.

PR #331, opened 2016-04-10
Separate status codes by type and add helper methods to check status

mwitek:
* Break status codes up by type
* Add status boolean helper methods to check status types.
example usage:
```
HTTP.get('http://google.com').status.informational?
HTTP.get('http://google.com').status.success?
HTTP.get('http://google.com').status.redirect?
HTTP.get('http://google.com').status.client_error?
HTTP.get('http://google.com').status.server_error?
```
returns true or false, depending on status code type

### DIFF
--- a/lib/http/response/status.rb
+++ b/lib/http/response/status.rb
@@ -83,6 +83,36 @@ module HTTP
         "#{code} #{reason}".strip
       end
 
+      # Check if status code is informational (1XX)
+      # @return [Boolean]
+      def informational?
+        INFORMATIONAL.keys.include?(code)
+      end
+
+      # Check if status code is successful (2XX)
+      # @return [Boolean]
+      def success?
+        SUCCESSFUL.keys.include?(code)
+      end
+
+      # Check if status code is redirection (3XX)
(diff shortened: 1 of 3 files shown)


ook:
Nice, I like it :+1: 
PR seems valid to me.

tarcieri:
Yeah, looks good to me too

ixti:
Why not simply:

``` ruby
def informational?
  100 <= self && self < 200
end

def success?
  200 <= self && self < 300
end

# ...
```

ixti:
Is there a real need in those constants?

ixti:
There's even a warning line in the beginning of reasons file:
https://github.com/mwitek/http/blob/824b90b33aefcfbf8110106d1f5c29815280c3fe/lib/http/response/status/reasons.rb#L1

ixti:
Just for the record, when you need to check if key is in hash or not it's better to use `Hash#key?` rather than creating new `Array` with `Hash#keys` and then calling `#include?` on that array. :D

mwitek:
:doh:!  Thanks for the tip :)

PR #333, opened 2016-04-11
Headline shouldn't include host for https requests through proxy.

### LINKED ISSUE #206: Subtle issue with normalizing URIs

Connorhd:
I think this actually fixes #206 (at least one of the later comments in it)

### DIFF
--- a/lib/http/request.rb
+++ b/lib/http/request.rb
@@ -142,7 +142,7 @@ module HTTP
 
     # Compute HTTP request header for direct or proxy request
     def headline
-      request_uri = using_proxy? ? uri : uri.omit(:scheme, :authority)
+      request_uri = (using_proxy? && !uri.https?) ? uri : uri.omit(:scheme, :authority)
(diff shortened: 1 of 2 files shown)


ixti:
Merged via https://github.com/httprb/http/commit/2625436bb25ab779e79b4b1ad6f2df9f94e43e48

ixti:
Thanks!

PR #336, opened 2016-04-29
Simplify a couple of methods

pavlik4k:


lib/http/response/parser.rb
```diff
@@ -34,17 +34,12 @@ def on_headers_complete(headers)
...
-        end
+        @chunk ||= ''
+        @chunk << chunk
       end
 
       def chunk
-        chunk  = @chunk
-        @chunk = nil
-        chunk
+        remove_instance_variable(:@chunk)
```
sferik: missing: `if defined?(:@chunk)`

tarcieri:
This is not semantically equivalent (and breaks the tests), nor would I consider it a simplification

pavlik4k:
Actually it doesn't break specs. You should check your ci

tarcieri:
On Friday, April 29, 2016, pavlik4k <notifications@github.com> wrote:

> Actually it doesn't break specs. You should check your ci
>
No, your code is broken (see @sferik's earlier comment):

 Failure/Error: remove_instance_variable(:@chunk)

 NameError:

       instance variable @chunk not defined



-- 
Tony Arcieri

lib/http/response/parser.rb
```diff
@@ -34,17 +34,12 @@ def on_headers_complete(headers)
       end
 
       def on_body(chunk)
-        if @chunk
-          @chunk << chunk
-        else
-          @chunk = chunk
-        end
+        @chunk ||= ''
```
ixti: this allocates new string.
and honestly i don't see significant improvement here. if/else condition looks easier to read imho in this particular case.

ixti:
I don't find this to be improvement either.

lib/http/response/parser.rb
tarcieri: Yeah +1, this adds an additional allocation and makes the code more complicated, in addition to just breaking things.

PR #338, opened 2016-05-10
update README examples to reflect actual use

tonyta:


tarcieri:
Thanks! Looks like an improvement.

PR #339, opened 2016-05-11
Always use byte methods when writing/slicing the write buffer

zanker:
This doesn't actually fail in our tests, because the test server we use apparently is broken too. I'm adding it more for posterity than it helps prevent regressions. I tested it against Passenger where it does fix it though!

We send `Content-Length` using `bytesize`, but we sliced the buffers using `length`. When those don't match up, things blow up because we send corrupted data. This would only happen if you're sending enough data that it can't be buffered into one `write` call.

### DIFF
--- a/lib/http/request/writer.rb
+++ b/lib/http/request/writer.rb
@@ -95,8 +95,8 @@ module HTTP
       def write(data)
         until data.empty?
           length = @socket.write(data)
-          break unless data.length > length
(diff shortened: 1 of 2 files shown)


tarcieri:
Worked through this one with @zanker and I concur `#bytesize` and `#byteslice` are the right things to use in this case

PR #341, opened 2016-05-12
String Manipulation and Comparison Operations

tonyta:
Hi again!

I thought I'd suggest some changes after reading through this beautiful source!

**Changes:**

1. Refactor some string manipulations so they are more performant (up to 3-4x faster) and more concise.
2. Swapped around some comparisons to be more conventional: with the (more) constant object on the right of the operator.

Hope this is helpful! 😸

### DIFF
--- a/lib/http/redirector.rb
+++ b/lib/http/redirector.rb
@@ -68,13 +68,13 @@ module HTTP
     # Check if we reached max amount of redirect hops
     # @return [Boolean]
     def too_many_hops?
-      1 <= @max_hops && @max_hops < @visited.count
+      @max_hops > 0 && @visited.count > @max_hops
     end
 
     # Check if we got into an endless loop
     # @return [Boolean]
     def endless_loop?
-      2 <= @visited.count(@visited.last)
+      @visited.count(@visited.last) > 1
     end
 
     # Redirect policy for follow
@@ -90,7 +90,7 @@ module HTTP
         verb = :get
       end
 
-      verb = :get if !SEE_OTHER_ALLOWED_VERBS.include?(verb) && 303 == code
+      verb = :get if !SEE_OTHER_ALLOWED_VERBS.include?(verb) && code == 303
 
       @request.redirect(uri, verb)
     end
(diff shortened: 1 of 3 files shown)


lib/http/redirector.rb
```diff
@@ -68,13 +68,13 @@ def perform(request, response)
     # Check if we reached max amount of redirect hops
     # @return [Boolean]
     def too_many_hops?
-      1 <= @max_hops && @max_hops < @visited.count
+      @max_hops > 0 && @visited.count > @max_hops
```
britishtea: I think `@visited.size` or `@visited.length` would be more descriptive here.

ixti:
Agree with `#1` - I like what you've did to content type parsing and will cherry-pick it.
But I have pretty strong objections against second one. :D

lib/http/response/status.rb
```diff
@@ -43,7 +43,7 @@ def coerce(object)
         # @param [#to_s] str
         # @return [Symbol]
         def symbolize(str)
-          str.to_s.downcase.tr("-", " ").gsub(/[^a-z ]/, "").gsub(/\s+/, "_").to_sym
+          str.to_s.downcase.tr("- ", "_").delete("^a-z_").to_sym
```
ixti:
```
[1] pry(main)> "Bad  request".downcase.tr("- ", "_").delete("^a-z_").to_sym
=> :bad__request
[2] pry(main)> "Bad  request".downcase.tr("-", " ").gsub(/[^a-z ]/, "").gsub(/\s+/, "_").to_sym
=> :bad_request
```

ixti:
Partially merged in https://github.com/httprb/http/commit/015a7070a08ae41951d54bb143d7d0a5a2b8a794

tonyta:
@ixti Thanks for the quick feedback. I'm just curious, what are your objections to swapping the variables?

ixti:
Well, mostly pretty subjective personal vision of beauty :D I prefer to see constants on the left of expression. I guess this is actually the only reason :D after all...

PR #343, opened 2016-05-12
faster string manipulation in response status

ixti:
Introduces backward-incompatible change:

```
[1] pry(main)> HTTP::Response::Status.coerce "Bad    Request"
HTTP::Error: Can't coerce String(Bad    Request) to HTTP::Response::Status
from /home/ixti/Projects/httprb/http/lib/http/response/status.rb:29:in `coerce'
```

Behavior before this patch:

```
[1] pry(main)> HTTP::Response::Status.coerce "Bad    Request"
=> 400
```

### DIFF
--- a/lib/http/response/status.rb
+++ b/lib/http/response/status.rb
@@ -43,7 +43,7 @@ module HTTP
         # @param [#to_s] str
         # @return [Symbol]
         def symbolize(str)
-          str.to_s.downcase.tr("-", " ").gsub(/[^a-z ]/, "").gsub(/\s+/, "_").to_sym
+          str.to_s.downcase.tr("- ", "_").delete("^a-z_").to_sym
         end
       end
 


ixti:
@httprb/core I like the change proposed by @tonyta but I'm worried a little bit about backward compatibility. I believe this should not be an issue in real life, but would like to get your thumbs up first, friends.

digitalextremist:
I like the change also. However I would define the method dynamically so it keeps backward compatibility, but has the new speed increases for environments that are compatible. Not sure exactly what the detection method would be if it would be... Probably a specific version condition before defining the method.

tonyta:
I thought about the duplicated separator, but did not see a failing test and didn't think it would be a expected behavior. That being said. I'd be happy to add a test for it and change `#tr` to `#tr_s` if support for this is preferred.

For your reference:
``` ruby
require 'benchmark/ips'

STRING = "Bad     Request"

def old_gsub_method
  STRING.to_s.downcase.tr("-", " ").gsub(/[^a-z ]/, "").gsub(/\s+/, "_").to_sym
end

def new_tr_method
  STRING.to_s.downcase.tr("- ", "_").delete("^a-z_").to_sym
end

def new_tr_s_method
  STRING.to_s.downcase.tr_s("- ", "_").delete("^a-z_").to_sym
end

puts "old_gsub_method: #{old_gsub_method.inspect}"
puts "new_tr_method: #{new_tr_method.inspect}"
puts "new_tr_s_method: #{new_tr_s_method.inspect}"

Benchmark.ips do |x|
  x.config warmup: 2, time: 5
  x.report("old_gsub_method") { old_gsub_method }
  x.report("new_tr_method") { new_tr_method }
  x.report("new_tr_s_method") { new_tr_s_method }
  x.compare!
end

# old_gsub_method: :bad_request
# new_tr_method: :bad_____request
# new_tr_s_method: :bad_request

# Calculating -------------------------------------
#      old_gsub_method    18.722k i/100ms
#        new_tr_method    49.107k i/100ms
#      new_tr_s_method    39.134k i/100ms
# -------------------------------------------------
#      old_gsub_method    212.507k (± 6.7%) i/s -      1.067M
#        new_tr_method    789.947k (± 7.3%) i/s -      3.978M
#      new_tr_s_method    550.160k (± 6.5%) i/s -      2.739M

# Comparison:
#        new_tr_method:   789947.3 i/s
#      new_tr_s_method:   550160.3 i/s - 1.44x slower
#      old_gsub_method:   212507.2 i/s - 3.72x slower      
```

zanker:
My main concern is that HTTP is a bit of a crazy standard. While I can't imagine why someone would have a server sending a response like that, I can also believe it happens. I'd defer to `curl`, if it can support `Bad     Request` we probably should too.

Thanks for your optimizations @tonyta!

ixti:
I believe I must explain a bit here. We don't care about response status description in fact.
`http_parser` throws only it's code. The only time we actually use this symbolization is when we converting auto-generated list of well-known *status code -> status reason* mapping to *symbolized reason -> status code` one. Backward incompatibility I was referring to is pretty exotic. And might occur only if sombody have something like this in their code:

``` ruby
module MyApplication
  BAD_REQUEST = HTTP::Response::Status["Bad Request"]
  # ...
end
```

In other words, I believe I have over-saturated *backward compatibility* term here ;))

ixti:
I think Status needs to be refactored completely in fact. Want to get rid of `Delegate` inheritance.

tarcieri:
This looks extremely stale...

ixti:
Closing this for now. Status will need some love and refactoring in future indeed.

PR #344, opened 2016-05-15
Add frozen_string_literal magic header

ixti:


ixti:
@httprb/core RFC

tarcieri:
Nice!

PR #345, opened 2016-05-15
Improve body encoding enforcement

ixti:
- validate encoding upon body instance creation
- enforce encoding on each and readpartial too
- return Enumerator if no block was given to `#each`

### DIFF
--- a/lib/http/response/body.rb
+++ b/lib/http/response/body.rb
@@ -14,20 +14,30 @@ module HTTP
         @client    = client
         @streaming = nil
         @contents  = nil
-        @encoding  = encoding
+
+        # see issue 312
+        begin
+          @encoding = Encoding.find encoding
+        rescue ArgumentError
+          @encoding = Encoding::BINARY
+        end
       end
 
       # (see HTTP::Client#readpartial)
       def readpartial(*args)
         stream!
-        @client.readpartial(*args)
+        force_encoding @client.readpartial(*args)
       end
 
       # Iterate over the body, allowing it to be enumerable
       def each
+        return to_enum __method__ unless block_given?
+
         while (chunk = readpartial)
           yield chunk
         end
+
+        self
       end
 
       # @return [String] eagerly consume the entire body as a string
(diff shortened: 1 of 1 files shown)


ixti:
@httprb/core comments/suggestions/critics are highly welcome. :D
@kylekyle as you were the one who added encoding validation, would like to know your opinion.

lib/http/response/body.rb
```diff
@@ -14,20 +14,30 @@ def initialize(client, encoding = Encoding::BINARY)
...
+        rescue ArgumentError
+          @encoding = Encoding::BINARY
+        end
       end
 
       # (see HTTP::Client#readpartial)
       def readpartial(*args)
         stream!
-        @client.readpartial(*args)
+        force_encoding @client.readpartial(*args)
```
tarcieri: The downside of `String#force_encoding` vs `String#encoding` is it will allow invalid encodings through. Should we do that in the event the upstream responder produces invalid encoded output? It's leaving a sort of lazy string encoding time bomb... I'm curious if there's precedent for how this sort of thing is handled elsewhere

ixti: Well, this particular patch just makes behavior of `readpartial`/`each` inline with `to_s`. But, yeah, would be interesting to know how others handle issue you mentioned.

tarcieri:
To handle string encodings properly we'd need the entire API to work around characters instead of bytes (handling of which has been a bit conflated, see #339, but Ruby doesn't make this easy)

`#force_encoding` will let through invalid data, which is fine if the caller is buffering, but working with bytewise operations for the underlying I/O you may wind up with partial multibyte characters.

Generally in other languages I've seen a sort of two-stage processing pipeline involving a byte array or byte buffer type that you then convert into a string type (which is typically canonically utf-8 in most modern languages)

lib/http/response/body.rb
tarcieri: But `#to_s` doesn't have to deal with partial data which ends in the middle of multibyte characters

ixti: Oh. Right. Overlooked that indeed.

ixti:
Thanks for detailed explanation @tarcieri.

tarcieri:
@ixti err, you don't have to close this PR, you just need to make sure the encoding is handled correctly :smile:

I think you can pass the `:encoding` option when you perform the `#readpartial` on the underlying socket and Ruby should buffer along character boundaries.

ixti:
Yeah, but that sounds like a totally different PR :D

PR #346, opened 2016-05-17
Hooks

paul:
Add some basic hooks around perform.

I was chatting a bit with @tarcieri in twitter about adding back caching and
mocking support. I think the best approach is to just add the hooks around
perform, and then implement that functionality as plugins to be vetted and
eventually merged.

To that end, I've implemented a the most basic hooks feature imaginable,
because I wanted to get feedback on the approach before I took it any further.
The next step I believe would be to implement the `Redirector` in terms of
these hooks, to ensure it is sufficient. I intend to do that in the PR, so
please do not merge it until that work in completed. 

In the meantime, please provide any feedback on this approach so we can discuss
it before I go any further.

### DIFF
--- a/lib/http/client.rb
+++ b/lib/http/client.rb
@@ -14,6 +14,18 @@ module HTTP
     extend Forwardable
     include Chainable
 
+    def before(&block)
+      @before_hooks << block
+    end
+
+    def around(&block)
+      @around_hooks << block
+    end
+
+    def after(&block)
+      @after_hooks << block
+    end
+
     KEEP_ALIVE         = "Keep-Alive".freeze
     CLOSE              = "close".freeze
 
@@ -23,6 +35,9 @@ module HTTP
       @default_options = HTTP::Options.new(default_options)
       @connection = nil
       @state = :clean
+      @before_hooks = []
+      @around_hooks = []
+      @after_hooks  = []
     end
 
     # Make an HTTP request
@@ -41,7 +56,19 @@ module HTTP
         :body    => body
       )
 
-      res = perform(req, opts)
+      @before_hooks.each { |hook| hook.call(req) }
+
+      if @around_hooks.any?
+        initial = method(:perform)
+        res = @around_hooks.inject(initial) do |result, hook|
+          lambda { |r, o| hook.call r, o, &result }
+        end.call(req, opts)
+      else
+        res = perform(req, opts)
+      end
+
+      @after_hooks.reverse.each { |hook| hook.call(req, opts, res) }
+
       return res unless opts.follow
 
       Redirector.new(opts.follow).perform(req, res) do |request|
--- a/spec/lib/http/hooks_spec.rb
+++ b/spec/lib/http/hooks_spec.rb
@@ -1,0 +1,151 @@
+# frozen_string_literal: true
+
+require "support/dummy_server"
+
+RSpec.describe HTTP, "hooks" do
+  run_server(:dummy) { DummyServer.new }
+  let(:client) { HTTP::Client.new }
+
+  describe "before hook" do
+    context "a hook that simply decorates the request" do
+      before { client.before { |req| req.headers.add("X-Hook-Test", "1") } }
+
+      subject(:response) { client.get dummy.endpoint + "/headers" }
+
+      it "should pass the decorated request through" do
+        request_headers = YAML.load(response.to_s)
+        expect(request_headers["x-hook-test"]).to eq(["1"])
+      end
+    end
+
+    context "muliple hooks" do
+      let(:hook_calls) { [] }
+
+      before do
+        client.before do |req, opts|
+          hook_calls << :first
+          req.headers.add("X-Hook-Test-A", "1")
+        end
+        client.before do |req, opts|
+          hook_calls << :second
+          req.headers.add("X-Hook-Test-B", "2")
+        end
+      end
+
+      subject(:response) { client.get dummy.endpoint + "/headers" }
+
+      it "should pass all request changes through" do
+        request_headers = YAML.load(response.to_s)
+        expect(request_headers["x-hook-test-a"]).to eq(["1"])
+        expect(request_headers["x-hook-test-b"]).to eq(["2"])
+      end
+
+      it "should call hooks in defined order" do
+        response
+        expect(hook_calls).to eq([:first, :second])
+      end
+    end
+  end
+
+  describe "after hook" do
+    context "a hook that simply decorates the response" do
+      before { client.after { |req, opts, res| res.headers.add("X-Hook-Test", "1") } }
+
+      subject(:response) { client.get dummy.endpoint }
+
+      it "should pass the decorated request through" do
+        expect(response.headers["X-Hook-Test"]).to eq("1")
+      end
+    end
+
+    context "muliple hooks" do
+      let(:hook_calls) { [] }
+
+      before do
+        client.after do |req, opts, res|
+          hook_calls << :first
+          res.headers.add("X-Hook-Test-A", "1")
+        end
+        client.after do |req, opts, res|
+          hook_calls << :second
+          res.headers.add("X-Hook-Test-B", "2")
+        end
+      end
+
+      subject(:response) { client.get dummy.endpoint }
+
+      it "should pass all request changes through" do
+        expect(response.headers["X-Hook-Test-A"]).to eq("1")
+        expect(response.headers["X-Hook-Test-B"]).to eq("2")
+      end
+
+      it "should call hooks in reverse order" do
+        response
+        expect(hook_calls).to eq([:second, :first])
+      end
+    end
+
+  end
+
+  describe "around hook" do
+    context "a hook that decorates the request and response" do
+      before do
+        client.around do |req, opts, &b|
+          req.headers.add("X-Hook-Test-Req", "1")
+          res = b.call req, opts
+          res.headers.add("X-Hook-Test-Resp", "2")
+          res
+        end
+      end
+
+      subject(:response) { client.get dummy.endpoint + "/headers" }
+
+      it "should pass request and response changes through" do
+        request_headers = YAML.load(response.to_s)
+        expect(request_headers["x-hook-test-req"]).to eq(["1"])
+        expect(response.headers["X-Hook-Test-Resp"]).to eq("2")
+      end
+    end
+
+    context "nested hooks" do
+      let(:hook_calls) { [] }
+
+      let(:inner_hook) do
+        lambda do |req, opts, &b|
+          hook_calls << :inner_before
+          res = b.call req, opts
+          res.headers["X-Inner-Hook"] = 1
+          hook_calls << :inner_after
+          res
+        end
+      end
+
+      let(:outer_hook) do
+        lambda do |req, opts, &b|
+          hook_calls << :outer_before
+          res = b.call req, opts
+          res.headers["X-Outer-Hook"] = 1
+          hook_calls << :outer_after
+          res
+        end
+      end
+
+      before do
+        client.around(&inner_hook)
+        client.around(&outer_hook)
+      end
+
+      subject(:response) { client.get dummy.endpoint + "/headers" }
+
+      it "should pass all request changes through" do
+        expect(response.headers["X-Inner-Hook"]).to eq("1")
+        expect(response.headers["X-Outer-Hook"]).to eq("1")
+      end
+
+      it "should call the earliest defined hooks as the innermost hook" do
+        response
+        expect(hook_calls).to eq([:outer_before, :inner_before, :inner_after, :outer_after])
+      end
+    end
+  end
+end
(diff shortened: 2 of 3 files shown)


ixti:
I remember @tarcieri was against (so do I) middleware-alike handlers (`around` hooks in your PR). I would like to better see event alike notifiers with API like:

``` ruby
client.on :request do
  # request is about to be performed
end
```

Probably with throw/catch signaling to achieve early loop termination if needed.

Caching should be part of core I believe. It just needed to be re-written, to handle all edge cases properly and do not interfere with other parts when not in use. What comes to redirector, I don't see how it's logic can be implemented with hooks without making code messy and less testable.

tarcieri:
I think we need something like this and I'd like for it to be as lightweight as possible so as not to adversely affect performance. This seems like a decent start.

I too am curious if the around callbacks are actually necessary for caching, but see the argument of how the Redirector acts as a special-case middleware of sorts.

paul:
I'm also not a fan of the "nested-middleware" style of hooks, but I don't think simple notifiers like your example are sufficient to accomplish everything. My primary use-cases are:

1. HTTP Caching
1. Intercepting requests for testing higher layers, similar to [request_interceptor](https://github.com/t6d/request_interceptor)
1. (Possibly) Adding serializers/deserializers. I'm not yet convinced if this is a good idea.

Here's an example of how I imagine this working. I think HTTP.rb should stay as lean as possible, but provide attachment points that would allow the consumers to create their own subclass client that can be enhanced with the features they need.

```ruby
class MyHttpClient < HTTP::Client
  include HTTP::Cache

  self.cache_store = redis_connection

  before :serialize_payload
  after  :deserialize_body

  def serialize_payload(request, opts)
    request.headers["Content-Type"] = "application/vnd+myapp+json"
    request.body = opts.payload.serialize
  end

  def deserialize_body(request, response)
    data = JSON.load(response.body.to_s)
    obj = data["type"].constantize.deserialize(data)
    response.payload = obj
  end
end

class User < Struct.new(:name, :email)

  def self.deserialize(json)
    self.new(json["name"], json["email"])
  end

  def serialize
    {
      "type"  => "User",
      "name"  => name,
      "email" => email
    }
  end

end

user     = User.new("paul", "psadauskas@gmail.com")
client   = MyHttpClient.new
response = client.post("/users", payload: user)

response.payload.class #=> User

```

The hooks here allow a user to implement whatever de/serializers they need, without HTTP.rb needing to support it explicitly. Additionally, the `HTTP::Cache` gets mixed in that adds its functionality via those hooks.

Caching is a pretty complicated beast. It couldn't be written using a single `on :request` event, it would at least need an `on :response do |request, response|`, because it needs the metadata from the request when storing the response in the cache. Additionally, to make a conditional request, or when retrieving the cached response, it would need to interrupt the request perform and not just return `nil`, but return the cached response as well.

The Redirector also has a similar problem, it needs to not just make another request, but it needs to go through the whole event pipeline again to see if anything would have changed the request being made. Additionally, to be "to spec", some of the redirect codes instruct the client they should ask the user explicitly before following the redirect. We tackled this problem in [resourceful][] by allowing the consumer to specify a callback so they can conditionally decide to follow the redirect.

[resourceful]: https://github.com/paul/resourceful#redirection-based-on-callback-results

In any case, I'm happy to do something other than the nested `around` callbacks, but I don't know of any other solution that meets the needs as succinctly. I'm also happy to implement the additional functionality outlined here as add-on gems, I just need these attachment points to accomplish that.

ixti:
re redirect code. yeah. that's why redirector has strict and non strict modes atm :D
https://github.com/httprb/http/issues/112

paul:
I wrote up a simple benchmark for comparing nested method vs my naive approach vs AS::Callbacks. My approach is slower by ~14x, and AS::C is another ~2x slower than that for a single callback. With many nested callbacks (10 deep), my approach is still 14x slower, while AS::C is somehow 2x faster? I'm not sure how that is, maybe because AS::C doesn't support passing arguments to the callbacks, which I think is needed here.

I'm also not sure how much this actually matters, since the single around cb adds 4 µsec to a single request, and the 10-deep hooks adds 12 µsec.

tarcieri:
Crazy idea: keep the callbacks/middleware in a separate module that can be
included into HTTP::Client subclasses that want to leverage them

On Tuesday, May 17, 2016, Paul Sadauskas <notifications@github.com> wrote:

> I'm also not a fan of the "nested-middleware" style of hooks, but I don't
> think simple notifiers like your example are sufficient to accomplish
> everything. My primary use-cases are:
>
>    1. HTTP Caching
>    2. Intercepting requests for testing higher layers, similar to
>    request_interceptor <https://github.com/t6d/request_interceptor>
>    3. (Possibly) Adding serializers/deserializers. I'm not yet convinced
>    if this is a good idea.
>
> Here's an example of how I imagine this working. I think HTTP.rb should
> stay as lean as possible, but provide attachment points that would allow
> the consumers to create their own subclass client that can be enhanced with
> the features they need.
>
> class MyHttpClient < HTTP::Client
>   include HTTP::Cache
>
>   self.cache_store = redis_connection
>
>   before :serialize_payload
>   after  :deserialize_body
>
>   def serialize_payload(request, opts)
>     request.headers["Content-Type"] = "application/vnd+myapp+json"
>     request.body = opts.payload.serialize
>   end
>
>   def deserialize_body(request, response)
>     data = JSON.load(response.body.to_s)
>     obj = data["type"].constantize.deserialize(data)
>     response.payload = obj
>   endend
> class User < Struct.new(:name, :email)
>
>   def self.deserialize(json)
>     self.new(json["name"], json["email"])
>   end
>
>   def serialize
>     {
>       "type"  => "User",
>       "name"  => name,
>       "email" => email
>     }
>   end
> end
>
> user     = User.new("paul", "psadauskas@gmail.com <javascript:_e(%7B%7D,'cvml','psadauskas@gmail.com');>")
> client   = MyHttpClient.new
> response = client.post("/users", payload: user)
>
> response.payload.class #=> User
>
> The hooks here allow a user to implement whatever de/serializers they
> need, without HTTP.rb needing to support it explicitly. Additionally, the
> HTTP::Cache gets mixed in that adds its functionality via those hooks.
>
> Caching is a pretty complicated beast. It couldn't be written using a
> single on :request event, it would at least need an on :response do
> |request, response|, because it needs the metadata from the request when
> storing the response in the cache. Additionally, to make a conditional
> request, or when retrieving the cached response, it would need to interrupt
> the request perform and not just return nil, but return the cached
> response as well.
>
> The Redirector also has a similar problem, it needs to not just make
> another request, but it needs to go through the whole event pipeline again
> to see if anything would have changed the request being made. Additionally,
> to be "to spec", some of the redirect codes instruct the client they should
> ask the user explicitly before following the redirect. We tackled this
> problem in resourceful
> <https://github.com/paul/resourceful#redirection-based-on-callback-results>
> by allowing the consumer to specify a callback so they can conditionally
> decide to follow the redirect.
>
> In any case, I'm happy to do something other than the nested around
> callbacks, but I don't know of any other solution that meets the needs as
> succinctly. I'm also happy to implement the additional functionality
> outlined here as add-on gems, I just need these attachment points to
> accomplish that.
>
> —
> You are receiving this because you were mentioned.
> Reply to this email directly or view it on GitHub
> <https://github.com/httprb/http/pull/346#issuecomment-219796618>
>


-- 
Tony Arcieri

ixti:
@paul 

I was definitely not advocating for AS::C. In fact I would never even proposed to pull any part of AS... IMHO it's just overkill :D

Just saying that I don't like russian-doll style middlewares :D Although can't propose anything better atm.


@tarcieri 

Sounds awesome. Not sure how doable is it though ;))

pezra:
It would be possible, i think, to do this with just before and after hooks if the before hooks had a way to skip to the response handling phase of the `#request` method. Eg, if a before hook returned a response object the before hook execution code could immediately skip to the after hooks using that response. I don't know if that would be more appealing or not.

As a side note, this hook implementation requires mutation of the request and responses objects. Personally, i prefer a more immutable style. Perhaps hooks should be required to return the request or response object to be used going forward?

tarcieri:
On Tuesday, May 17, 2016, Alexey Zapparov <notifications@github.com> wrote:
>
> @tarcieri <https://github.com/tarcieri>
>
> Sounds awesome. Not sure how doable is it though ;))
>
What if the module redefined the #perform method to run the callbacks, and
called "super" where it presently is doing "method(:perform)"? Wonder if
method(:super) works... (I'm on a phone or otherwise I'd check)


-- 
Tony Arcieri

paul:
@ixti I was absolutely not advocating the use of AS::C here, either, I just included it in my benchmark as a point of comparison.

@pezra Using only `#before`, how would one implement a callback that would differentiate between "modify the outgoing request", "halt the request on some condition and return nothing" and "return a response-like object without making a request"? `#around` allows you a great deal more control over the flow of the request/response.

@tarcieri I was totally planning on extracting this code to an `HTTP::Hooks` module. I'm not sure if its optional, though, unless we also made the Redirector optional as well, since it would rely on it.

paul:
@pezra I'm also :+1: about immutable requests/responses. This implementation mutates only because the existing request/response are simple to mutate, and would need a bit of glue code to make them easier to dupe. OTOH, I'm uncertain about the costs associated with duping an entire request object just to eg add a single header to it.

pezra:
@paul, i am basically suggesting turning the before and after hook evaluations into a reduce. Eg,

```ruby
req = @before_hooks.reduce(req) { |req, hook| hook.call(req) }
```
It might be that hooks would still choose mutate the request, or response, but it would allow hooks authors to decide.

pezra:
@paul, re short circuiting the before hook chain: I think you could do it based on type of the return value from hook. If a hook returns an `HTTP::Request` then proceed with the next hook; if it returns an `HTTP::Response` then break then skip the rest of the before hooks and the perform and go directly to the after hook chain.

paul:
@pezra I'm not sure about the return type altering behavior. For example, if we did that, I would expect that if the `before` block returned false-y or `nil`, then execution should stop. However, that would mean something like `before { |req| req.body = nil }` would halt execution of the request, which would not be expected at all.

paul:
@pezra converting the `before` and `after` hooks to a reduce would cause them to have all the same downsides as the `around`, eg deeply nested callbacks with associated the performance impact.

paul:
I was talking with @pezra on slack a bit (is there an http.rb slack?). What do you guys think of something like this, borrowing the `throw`/`catch` idiom from Sinatra:

```ruby
# halt the request
before do |req|
  if bad_request(req)
    throw :halt
  end
end

# or return an early response
before do |req|
  if cached_response = fetch_from_cache(req)
    throw :response, cached_response
  end
end
```

These could be further simplified with helpers, `halt` and `respond(cached_response)`, respectively. Thoughts, @ixti and @tarcieri ?

tarcieri:
Oi, non-local flow control. Not super enthusiastic about it, but not adamantly against it either.

ixti:
I was proposing throw/catch above. And although I also a bit discomfort with non-local flow control, I think it's the best option here to keep overhead as minimal as possible.

zanker:
I'd like to keep it as minimum as possible, and not turn it into a crazy callback system like what Faraday has. Besides that, I don't have too much an opinion on what we do.

paul:
The third option, of course, would just be to module prepend for plugins.

```ruby
module HTTP::Caching
  def perform(req, opts)
    return resp if resp = fetch_from_cache(req)
    super
  end
  # class methods, etc...
end

class MyClient
  prepend HTTP::Caching
  self.cache_store = redis_connection
  # ...
end

```

This is probably the simplest, but I don't know how intuitive it would be for users. It would also make the method `Client#perform` a public API, if it isn't already.

tarcieri:
@paul you don't need `prepend` for that. `include` will work just fine (unless you're trying to compose with other modules in the same class that overload perform)

I'm fine with `HTTP::Client#perform` being a public API

pezra:
The `prepend`/`include` approach to extension effectively makes, some parts of, of `#request`'s implementation part of the public contract. That could significantly constrain future changes to request processing.

tarcieri:
@pezra is that really the case, or would it just make `#perform` part of the public contract?

@paul I think the least invasive way to proceed is to make `HTTP::CachingClient` which redefines `#perform`. Does that seem OK to you?

We can re-examine a callback system if ^^^ is insufficient.

pezra:
@tarcieri, it expands the contract of `#request` to include the fact that it must call `#perform` to accomplish any actual HTTP requests. That clearly constraints future changes. Not saying such a constraint is unacceptable, or even bad, just pointing it out.

tarcieri:
@pezra yeah, completely agreed on that, but it seems like the most minimal touch point for solving this problem

tarcieri:
I'm going to go ahead and close this issue out for now. Feel free to re-open it if you think there's a legitimate use case for flexible hooks like this.

paul:
Yeah, I'm perfectly happy using modules if you're ok with making `#perform` into a public API contract.

PR #349, opened 2016-05-27
adding 418 status code

mdayaram:


ixti:
Reasons Hash is auto-generated (which is clearly stated in the head of the file: https://github.com/httprb/http/blob/master/lib/http/response/status/reasons.rb#L2) from: http://www.iana.org/assignments/http-status-codes/http-status-codes.xml

ixti:
I plan to refactor `Status` class in general, so that it will allow you to add extra reasons.

ixti:
Closing this in favor of #350

mdayaram:
woops, missed that, sorry.

PR #350, opened 2016-05-27
Support extra non-assigned HTTP statuses

ixti:
@mdayaram wanted to add `I'm a teapot` status code in #349 into list of known status codes.
Personally I have no strong opinion on that. So @httprb/core vote in, guys!

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -32,17 +32,26 @@ task :generate_status_codes do
 
   url = "http://www.iana.org/assignments/http-status-codes/http-status-codes.xml"
   xml = Nokogiri::XML HTTP.get url
-  arr = xml.xpath("//xmlns:record").reduce [] do |a, e|
-    code = e.xpath("xmlns:value").text.to_s
-    desc = e.xpath("xmlns:description").text.to_s
+  known_codes = {
+    418 => "I'm a Teapot"
+  }
 
-    next a if "Unassigned" == desc || "(Unused)" == desc
+  xml.xpath("//xmlns:record").each do |node|
+    code = node.xpath("xmlns:value").text.to_s
+    desc = node.xpath("xmlns:description").text.to_s
 
-    a << "#{code} => #{desc.inspect}"
+    next if "Unassigned" == desc || "(Unused)" == desc
+
+    known_codes[code.to_i] = desc
   end
 
   File.open("./lib/http/response/status/reasons.rb", "w") do |io|
+    reasons = known_codes.keys.sort.map do |code|
+      "\n              #{code} => #{known_codes[code].inspect}"
+    end
+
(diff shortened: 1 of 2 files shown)


tarcieri:
Haha uhh, interesting...

Rakefile
```diff
@@ -32,17 +32,26 @@ task :generate_status_codes do
 
   url = "http://www.iana.org/assignments/http-status-codes/http-status-codes.xml"
   xml = Nokogiri::XML HTTP.get url
-  arr = xml.xpath("//xmlns:record").reduce [] do |a, e|
-    code = e.xpath("xmlns:value").text.to_s
-    desc = e.xpath("xmlns:description").text.to_s
+  known_codes = {
+    418 => "I'm a Teapot"
```
mdayaram: yay!

britishtea:
What would be you opinion on not `#freeze`ing the Hash itself, so non-standard status codes can be added at run time if required?

This would also be useful for e.g. request methods.

ixti:
@britishtea We're defining helpers on status based on known response descriptions.
So one can do something like `response.status.ok?`. At the moment I'm thinking on providing a DSL so that one will be able to define non-standard statuses on his own, e.g.:

``` ruby
HTTP::Response::Status.register(418, "I'm a teapot")
```

ixti:
Also, I would like to get any thoughts on removing those helpers all together.

After all we have more generic helpers: `#info?`, `#success?`, `#redirect?`, `#client_error?` and `#server_error?`. The only helper I guess that really makes sense is `#ok?` that will be a syntax sugar for `200 == response.status` and it should be defined on response itself. I'm gonna provide a PR with all the changes I have in mind todya-tomorrow, so it will become easier to discuss this :D

britishtea:
@ixti What I meant was keep the `HTTP::Response::Status::REASONS` as is, but remove the last call to `#freeze` on [`      }.each { |_, v| v.freeze }.freeze`](https://github.com/httprb/http/blob/master/lib/http/response/status/reasons.rb#L77). Standard `#[]` and `#[]=` can then be used to "register" non-standard reasons at run time.

In a similar spirit (but slightly off-topic), I'd like something similar for `HTTP::Request::METHODS`. At work we're using a non-standard request method (`PURGE`) to communicate with a Varnish cache server. We're currently using this nasty piece of code for that:

```ruby
tmp, $VERBOSE = $VERBOSE, nil
HTTP::Request::METHODS = HTTP::Request::METHODS.dup
HTTP::Request::METHODS << :purge
HTTP::Request::METHODS.freeze
$VERBOSE = tmp
```

Obviously the DSL you propose would work fine too.

tarcieri:
This also looks extremely stale...

ixti:
Same as other status related PR - closing for now.

PR #352, opened 2016-05-29
Cleanup dev dependencies

tarcieri:
- Remove celluloid-io: we don't use it anymore
- Remove Ruby 1.9isms
- Add more :require => false


ixti:
:+1:

PR #353, opened 2016-06-10
Avoid a dependency cycle between Client and Connection classes.

jhbabon:
I found out that sometimes you can get this error when using the gem:

    from /.../gems/http-2.0.1/lib/http.rb:8:in  `<top (required)>'
    from /.../gems/http-2.0.1/lib/http.rb:8:in  `require'
    from /.../gems/http-2.0.1/lib/http/client.rb:7:in  `<top (required)>'
    from /.../gems/http-2.0.1/lib/http/client.rb:7:in  `require'
    from /.../gems/http-2.0.1/lib/http/connection.rb:4:in  `<top (required)>'
    from /.../gems/http-2.0.1/lib/http/connection.rb:4:in  `require'

The problem was that `http.rb` was requiring `http/client.rb` that at the same time was requiring `http/connection.rb` which was requiring `http/client.rb` again.

I think that in general is not very good to have a cycle like this in the dependency graph of the code, so I decided to remove that dependency by moving the constants `KEEP_ALIVE` and `CLOSE` to the Connection class. This way Connection is now independent from Client.

Maybe this constants could go to another module but I couldn't think in a better place.

### DIFF
--- a/lib/http/client.rb
+++ b/lib/http/client.rb
@@ -14,10 +14,7 @@ module HTTP
     extend Forwardable
     include Chainable
 
-    KEEP_ALIVE         = "Keep-Alive".freeze
-    CLOSE              = "close".freeze
-
-    HTTP_OR_HTTPS_RE   = %r{^https?://}i
+    HTTP_OR_HTTPS_RE = %r{^https?://}i
 
     def initialize(default_options = {})
       @default_options = HTTP::Options.new(default_options)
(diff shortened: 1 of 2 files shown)


ixti:
:+1: Thank you!

jhbabon:
That was fast. Thanks 😄

jhbabon:
@ixti Hi!
Sorry to bother you, but do you know if there is any plan to release a new version of the gem with this fix?

ixti:
Sorry for delay. `v2.0.2` with this patch just got released. Thanks for your contribution once again!

jhbabon:
@ixti thank you for your work!

Issue #335, opened 2016-04-25
Content-Length not being set for POST requests?


bsodmike:
Hi,

I was integrating reCaptcha today and can into something curious, `HTTP.post('https://www.google.com/recaptcha/api/siteverify', params: { secret: secret, response: response })` and got

```
<p><b>411.</b> <ins>That’s an error.</ins>\n  <p>POST requests require a <code>Content-length</code> header.  <ins>That’s all we know.</ins>\n"
```

I tried setting the content length to some random value but I found it would hang of I did,

``` ruby
HTTP['Content-Length': 500].post('https://www.google.com/recaptcha/api/siteverify', params: { secret: secret, response: response })

#...or

HTTP[content_length: 500].post('https://www.google.com/recaptcha/api/siteverify', params: { secret: secret, response: response })
```

I suppose, this is because the content length is invalid; I didn't have time prying into the library to figure this out as I've gotten HTTParty working just fine.

Thanks!

tarcieri:
You're correct we should be sending Content-Length if the length of the request body is known in advance. That said, in your case the request body is empty, so sending an arbitrary length like that will hang the remote server as it waits for the rest of the body.

bsodmike:
Ah no wonder, it's just strange that the reCaptcha server has this set.  I wonder if we could add some log notification if the gem if waiting for rest of the body and there isn't any response after a certain timeout?  It's just to potentially prevent others seeing something similar, even though I'd agree this is not a typical use-case.

Cheers @tarcieri

tarcieri:
This gem has extensive timeout support: https://github.com/httprb/http/wiki/Timeouts

bsodmike:
Ooh fantastic, thanks.

bsodmike:
This does the trick, thanks!

```ruby
HTTP[content_length: 0]. \
  timeout(write: 2, connect: 5, read: 20). \
  post('https://www.google.com/recaptcha/api/siteverify', params: {
  secret: secret, response: response 
})

=> {"success"=>true,
 "challenge_ts"=>"2016-04-25T15:53:14Z",
 "hostname"=>"localhost"}
```

tarcieri:
Reopening as we should probably set the request Content-Length if it's known in advance.

ixti:
Resolved by #360

PR #360, opened 2016-07-20
Set Content-Length request header when request body is nil

britishtea:
This closes issue #335 by checking if the request body is `nil` and setting the `Content-Length` header to 0.

When the body is a String, the Content-Length header was already set.

### DIFF
--- a/lib/http/request/writer.rb
+++ b/lib/http/request/writer.rb
@@ -53,6 +53,8 @@ module HTTP
       def add_body_type_headers
         if @body.is_a?(String) && !@headers[Headers::CONTENT_LENGTH]
           @request_header << "#{Headers::CONTENT_LENGTH}: #{@body.bytesize}"
+        elsif @body.nil? && !@headers[Headers::CONTENT_LENGTH]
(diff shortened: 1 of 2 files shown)


zanker:
LG, although it seems like our build broke which isn't your fault

ixti:
CA authority gem requries activesupport, which is resolved to 5.0.0, which in it's turn has strict supported ruby version: 2.2.2+

ixti:
So, we should either get rid of ca-authority dep, or lock down activesupport to `< 5` and add a comment why we ever have that dependency in gemfile :D

tarcieri:
I can potentially try to get rid of `certificate_authority`. I wrote a bash script to generate long-lived certs that can just be checked in for redis-rb:

https://github.com/redis/redis-rb/blob/master/test/support/ssl/gen_certs.sh

Short term we should probably just lock to active_support 4.x as a dev dependency

ixti:
Thanks. Merged via 39e9932285b9d1f6c4517849fb9357813d4c38e1

PR #362, opened 2016-07-20
connect_ssl uses connect_timeout (Closes #359)

### LINKED ISSUE #359: #connect_ssl influenced by read/write timeouts, not connect
https://github.com/httprb/http/blob/master/lib/http/timeout/per_operation.rb#L30-L36

The rescue's seem ok, problem is the timeout used underneath. The correct timeout in this case should be connect_timeout, but in both blocks two different timeouts will be used. 

This could be easily handled if the rescue_ methods would accept a timeout parameter which would overwrite the default IMO.

TiagoCardoso1983:


TiagoCardoso1983:
someone?

lib/http/timeout/null.rb
```diff
@@ -55,19 +55,19 @@ def write(data)
...
 
       # Retry writing
-      def rescue_writable
+      def rescue_writable(tout = write_timeout)
         yield
       rescue IO::WaitWritable
-        retry if @socket.to_io.wait_writable(write_timeout)
-        raise TimeoutError, "Write timed out after #{write_timeout} seconds"
+        retry if @socket.to_io.wait_writable(tout)
+        raise TimeoutError, "Write timed out after #{tout} seconds"
```
ixti: can't `tout` be names full-length `timeout`?

tarcieri: 👍

ixti:
If you will rebase on top of master it should become green on CI

tarcieri:
@TiagoCardoso1983 thanks!

ixti:
:clap:

PR #364, opened 2016-07-23
Add HTTP::Response#connection

janko-m:
The [Down](https://github.com/janko-m/down) gem provides the ability to wrap anything that streams chunks into an IO object. I want to add support for HTTP.rb to Down so that you can do the following:

```rb
response = HTTP.any.request.options.get("http://example.com")
remote_file = Down.open(response)
remote_file # IO object which downloads only how much it needs
```

For this I need to be able to access the connection through the response object, so that I can close the connection if either the downloading finishes or user calls `remote_file.close`.

### DIFF
--- a/lib/http/response/body.rb
+++ b/lib/http/response/body.rb
@@ -10,17 +10,22 @@ module HTTP
       include Enumerable
       def_delegator :to_s, :empty?
 
-      def initialize(client, encoding = Encoding::BINARY)
-        @client    = client
-        @streaming = nil
-        @contents  = nil
-        @encoding  = encoding
+      # The connection object used to make the corresponding request.
+      #
+      # @return [HTTP::Connection]
+      attr_reader :connection
+
+      def initialize(connection, encoding = Encoding::BINARY)
+        @connection = connection
+        @streaming  = nil
+        @contents   = nil
+        @encoding   = encoding
       end
 
       # (see HTTP::Client#readpartial)
       def readpartial(*args)
         stream!
-        @client.readpartial(*args)
+        @connection.readpartial(*args)
       end
 
       # Iterate over the body, allowing it to be enumerable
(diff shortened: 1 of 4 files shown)


ixti:
I believe connection/client thingy might be changed in relatively near future.

ixti:
@zanker @tarcieri I think you guys know more of this ;))

janko-m:
@ixti In which way might it be changed? I think it would still make sense to have it available through the response object, because at the moment I have to do

```rb
response.body.instance_variable_get("@client").close
```

to close the connection

janko-m:
Ping 😃. Since you said it might be changed soon, that's all the more reason to create a public method for accessing the connection.

lib/http/response/body.rb
```diff
@@ -10,6 +10,8 @@ class Body
       include Enumerable
       def_delegator :to_s, :empty?
 
+      attr_reader :client
```
tarcieri: So, it appears as if the `@client` instance variable is in fact an `HTTP::Connection`. In that case, I suggest renaming it (and the `initialize` param) to `@connection` and exposing an `attr_reader: connection` here.

janko-m: Done!

tarcieri:
Looks good to me 👍

janko-m:
Thanks!

ixti:
@janko-m I'm very sorry. Got lots of things moving around this time of my life (which seems like happen every August to me). I'm totally fine with the PR (I should have noted that more explicitly). There were ideas of moving pretty much of sockets work things into separate gem.

There definitely will be changes to API of HTTP methods once I'll find time to work on that (block form that will be closing underlying connection auto-matically).

ixti:
In any case. Thanks for your PR! :+1:

janko-m:
Don't want to be nagging, but it would be awesome to get a new release with `HTTP::Response#connection` and `HTTP::Response#content_length`, since I would like to start adding support for HTTP.rb to [Down](https://github.com/janko-m/down) 😃

tarcieri:
Released as 2.1.0.

janko-m:
@tarcieri Thank you! ❤️

PR #365, opened 2016-07-23
Add HTTP::Response#content_length

janko-m:
When converting an HTTP response object to an IO, I wanted that `#size` returns the value of the Content-Length header converted to integer, or nil if it's not present:

```rb
Integer(response["Content-Length"]) if response["Content-Length"]
```

I thought it would be nice to add this convenience method to HTTP.rb.

### DIFF
--- a/lib/http/response.rb
+++ b/lib/http/response.rb
@@ -86,6 +86,13 @@ module HTTP
       self
     end
 
+    # Value of the Content-Length header
+    #
+    # @return [Integer]
+    def content_length
+      Integer(headers[Headers::CONTENT_LENGTH]) if headers[Headers::CONTENT_LENGTH]
+    end
+
     # Parsed Content-Type header
     #
     # @return [HTTP::ContentType]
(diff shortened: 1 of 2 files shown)


ixti:
Thanks! Merged via fb1e90c2a22221ee15f93f537008c6466ab56fc3.

britishtea:
Perhaps a bit late, but I'm not a big fan of this method raising an `ArgumentError` when `Content-Language` has an unexpected value.

Can we rescue the exception and return `nil` instead in this case.

```ruby
    context "with Content-Length not an integer" do
      let(:headers) { {"Content-Length" => "foo"} }

      it { is_expected.to be_nil }
    end
```

janko-m:
@britishtea I'm ok with this idea 👍

ixti:
It's never too late. If you will send a PR - will merge it shortly - I'm on the side of this proposal too.

tarcieri:
👍

ixti:
@britishtea pushed fixed with your proposal.

Issue #357, opened 2016-07-20
Re-add io/wait operations


TiagoCardoso1983:
@tarcieri 

Coming from #298 and #322 

## Why

Just to understand my motivation, all ruby net/protocol libraries use ```io/wait```, so I'd say it's a must for libraries like ```http``` to update support.

## How

I was testing with the example from #298 until I got the semantics right. Here is the script I used:

```ruby
# test_timeout.rb
require 'http'
$strategy = ENV["STRATEGY"] || "master"

# from master
class HTTP::Timeout::PerOperation

  def readpartial(size)
    loop do
      result = @socket.read_nonblock(size, :exception => false)

      return :eof   if result.nil?
      return result if result != :wait_readable

      case $strategy
        when 'master'
          unless IO.select([@socket], nil, nil, read_timeout)
            raise TimeoutError, "Read timed out after #{read_timeout} seconds"
          end
        when 'bug'
          unless @socket.to_io.wait_readable(read_timeout)
            raise TimeoutError, "Read timed out after #{read_timeout} seconds"
          end
        when 'fix'
          case @socket.to_io.wait_readable(read_timeout)
            # http://ruby-doc.org/stdlib-2.2.5/libdoc/io/wait/rdoc/IO.html#method-i-wait
            # quote: .... returns self or nil when EOF is reached.
            when nil
              return :eof
            when false
              raise TimeoutError, "Read timed out after #{read_timeout} seconds"
            # else continue, you have socket
          end
      end
    end
  end
end

url = "https://play.google.com/store/apps/" \
      "category/GAMES/collection/top?start=400&num=100&hl=en&gl=us"

42.times do |i|
  start = Time.now.to_f

  print format("%3d ... ", i + 1)
  begin
    # http gem
    res = HTTP.timeout(:read => 5, :connect => 5, :write => 5).
               get(url)
    res.flush
     puts "pass (#{Time.now.to_f - start}) -> code: #{res.code}"
   rescue => e
     puts "fail (#{Time.now.to_f - start})"
     raise e
   end
 end
```

Run this script with the 3 strategies and compare the results. 

This might be controversial (I also don't fully understand it, as most of the times ```#wait_readable``` returns nil when it should return false), but you can with [the implementation on net/protocol](https://github.com/ruby/ruby/blob/trunk/lib/net/protocol.rb#L153-L169). 

[...]

PR #368, opened 2016-08-08
back to io/wait

HoneyryderChuck:
should take care of #357

### DIFF
--- a/lib/http/errors.rb
+++ b/lib/http/errors.rb
@@ -18,9 +18,9 @@ module HTTP
   # Generic Timeout error
   class TimeoutError < Error; end
 
-  # Header name is invalid
-  class InvalidHeaderNameError < Error; end
-
   # Header value is of unexpected format (similar to Net::HTTPHeaderSyntaxError)
-  class HeaderSyntaxError < Error; end
+  class HeaderError < Error; end
+
+  # @deprecated will be removed in v3.0.0
+  InvalidHeaderNameError = HeaderError
 end
--- a/lib/http/errors.rb
+++ b/lib/http/errors.rb
@@ -20,4 +20,7 @@ module HTTP
 
   # Header name is invalid
   class InvalidHeaderNameError < Error; end
+
+  # Header value is of unexpected format (similar to Net::HTTPHeaderSyntaxError)
+  class HeaderSyntaxError < Error; end
 end
--- a/lib/http/errors.rb
+++ b/lib/http/errors.rb
@@ -21,6 +21,4 @@ module HTTP
   # Header value is of unexpected format (similar to Net::HTTPHeaderSyntaxError)
   class HeaderError < Error; end
 
-  # @deprecated will be removed in v3.0.0
-  InvalidHeaderNameError = HeaderError
 end
--- a/lib/http/errors.rb
+++ b/lib/http/errors.rb
@@ -20,5 +20,4 @@ module HTTP
 
   # Header value is of unexpected format (similar to Net::HTTPHeaderSyntaxError)
   class HeaderError < Error; end
-
 end
--- a/lib/http/response.rb
+++ b/lib/http/response.rb
@@ -142,6 +142,7 @@ module HTTP
 
       encoding = @headers.get(Headers::TRANSFER_ENCODING)
 
+      # TODO: "chunked" is frozen in the request writer. How about making it accessible?
       encoding.last == "chunked"
     end
 
--- a/lib/http/response.rb
+++ b/lib/http/response.rb
@@ -110,7 +110,7 @@ module HTTP
       begin
         Integer(value)
       rescue ArgumentError
-        raise(HeaderSyntaxError, 'wrong Content-Length format')
+        raise(HeaderError, 'wrong Content-Length format')
       end
     end
 
--- a/lib/http/response.rb
+++ b/lib/http/response.rb
@@ -110,7 +110,6 @@ module HTTP
       begin
         Integer(value)
       rescue ArgumentError
-        raise(HeaderError, 'wrong Content-Length format')
       end
     end
 
--- a/lib/http/response.rb
+++ b/lib/http/response.rb
@@ -132,6 +132,31 @@ module HTTP
       end
     end
 
+    def content_length
+      @content_length ||= begin
+        # http://greenbytes.de/tech/webdav/rfc7230.html#rfc.section.3.3.3
+        # Clause 3: "If a message is received with both a Transfer-Encoding
+        # and a Content-Length header field, the Transfer-Encoding overrides the Content-Length.        
+        return nil if headers[Headers::TRANSFER_ENCODING]
+
+	clen = headers[Headers::CONTENT_LENGTH]
+	if clen
+          len = clen.slice(/\d+/) ||
+              raise(HeaderSyntaxError, 'wrong Content-Length format')
+          len.to_i
+        end
+      end	
+    end
+
+    def chunked?
+      encoding = headers.get(Headers::TRANSFER_ENCODING)
+      case encoding
+        when String then encoding == CHUNKED
+        when Enumerable then encoding.include?("chunked")
+        else false
+      end
+    end
+
     # Parse response body with corresponding MIME type adapter.
     #
     # @param [#to_s] as Parse as given MIME type
--- a/lib/http/response.rb
+++ b/lib/http/response.rb
@@ -51,7 +51,7 @@ module HTTP
 
         stream = body_stream_for(connection, opts)
 
-        @body = Response::Body.new(connection, stream, encoding)
+        @body = Response::Body.new(connection, stream, encoding: encoding, length: content_length)
       else
         @body = opts.fetch(:body)
       end
--- a/lib/http/response.rb
+++ b/lib/http/response.rb
@@ -142,7 +142,7 @@ module HTTP
 
       encoding = @headers.get(Headers::TRANSFER_ENCODING)
 
-      encoding.include?("chunked")
+      encoding.last == "chunked"
     end
 
     # Parse response body with corresponding MIME type adapter.
--- a/lib/http/response.rb
+++ b/lib/http/response.rb
@@ -51,7 +51,7 @@ module HTTP
 
         stream = body_stream_for(connection, opts)
 
-        @body = Response::Body.new(connection, stream, encoding: encoding, length: content_length)
+        @body = Response::Body.new(connection, stream, :encoding => encoding, :length => content_length)
       else
         @body = opts.fetch(:body)
       end
--- a/lib/http/response.rb
+++ b/lib/http/response.rb
@@ -110,6 +110,7 @@ module HTTP
       begin
         Integer(value)
       rescue ArgumentError
+        nil
       end
     end
 
--- a/lib/http/response.rb
+++ b/lib/http/response.rb
@@ -50,7 +50,6 @@ module HTTP
         encoding   = opts[:encoding] || charset || Encoding::BINARY
         stream     = body_stream_for(connection, opts)
 
-
         @body = Response::Body.new(stream, :encoding => encoding, :length => content_length)
       else
         @body = opts.fetch(:body)
--- a/lib/http/response.rb
+++ b/lib/http/response.rb
@@ -99,13 +99,18 @@ module HTTP
     #   (not an integer, e.g. empty string or string with non-digits).
     # @return [Integer] otherwise
     def content_length
+      # http://greenbytes.de/tech/webdav/rfc7230.html#rfc.section.3.3.3
+      # Clause 3: "If a message is received with both a Transfer-Encoding
+      # and a Content-Length header field, the Transfer-Encoding overrides the Content-Length.        
+      return nil if headers[Headers::TRANSFER_ENCODING]
+
       value = @headers[Headers::CONTENT_LENGTH]
       return unless value
 
       begin
         Integer(value)
       rescue ArgumentError
-        nil
+        raise(HeaderSyntaxError, 'wrong Content-Length format')
       end
     end
 
@@ -130,22 +135,6 @@ module HTTP
       @cookies ||= headers.each_with_object CookieJar.new do |(k, v), jar|
         jar.parse(v, uri) if k == Headers::SET_COOKIE
       end
-    end
-
-    def content_length
-      @content_length ||= begin
-        # http://greenbytes.de/tech/webdav/rfc7230.html#rfc.section.3.3.3
-        # Clause 3: "If a message is received with both a Transfer-Encoding
-        # and a Content-Length header field, the Transfer-Encoding overrides the Content-Length.        
-        return nil if headers[Headers::TRANSFER_ENCODING]
-
-	clen = headers[Headers::CONTENT_LENGTH]
-	if clen
-          len = clen.slice(/\d+/) ||
-              raise(HeaderSyntaxError, 'wrong Content-Length format')
-          len.to_i
-        end
-      end	
     end
 
     def chunked?
--- a/lib/http/response.rb
+++ b/lib/http/response.rb
@@ -102,7 +102,7 @@ module HTTP
       # http://greenbytes.de/tech/webdav/rfc7230.html#rfc.section.3.3.3
       # Clause 3: "If a message is received with both a Transfer-Encoding
       # and a Content-Length header field, the Transfer-Encoding overrides the Content-Length.        
-      return nil if headers[Headers::TRANSFER_ENCODING]
+      return nil if @headers[Headers::TRANSFER_ENCODING]
 
       value = @headers[Headers::CONTENT_LENGTH]
       return unless value
--- a/lib/http/response.rb
+++ b/lib/http/response.rb
@@ -102,10 +102,10 @@ module HTTP
       # http://greenbytes.de/tech/webdav/rfc7230.html#rfc.section.3.3.3
       # Clause 3: "If a message is received with both a Transfer-Encoding
       # and a Content-Length header field, the Transfer-Encoding overrides the Content-Length.        
-      return nil if @headers[Headers::TRANSFER_ENCODING]
+      return nil if @headers.include?(Headers::TRANSFER_ENCODING) ||
+                    !@headers.include?(Headers::CONTENT_LENGTH)
 
       value = @headers[Headers::CONTENT_LENGTH]
-      return unless value
 
       begin
         Integer(value)
--- a/lib/http/response.rb
+++ b/lib/http/response.rb
@@ -101,7 +101,7 @@ module HTTP
     def content_length
       # http://greenbytes.de/tech/webdav/rfc7230.html#rfc.section.3.3.3
       # Clause 3: "If a message is received with both a Transfer-Encoding
-      # and a Content-Length header field, the Transfer-Encoding overrides the Content-Length.        
+      # and a Content-Length header field, the Transfer-Encoding overrides the Content-Length.
       return nil if @headers.include?(Headers::TRANSFER_ENCODING)
 
       value = @headers[Headers::CONTENT_LENGTH]
--- a/lib/http/response.rb
+++ b/lib/http/response.rb
@@ -138,12 +138,11 @@ module HTTP
     end
 
     def chunked?
-      encoding = headers.get(Headers::TRANSFER_ENCODING)
-      case encoding
-        when String then encoding == CHUNKED
-        when Enumerable then encoding.include?("chunked")
-        else false
-      end
+      return false unless @headers.include?(Headers::TRANSFER_ENCODING)
+
+      encoding = @headers.get(Headers::TRANSFER_ENCODING)
+
+      encoding.include?("chunked")
     end
 
     # Parse response body with corresponding MIME type adapter.
--- a/lib/http/response.rb
+++ b/lib/http/response.rb
@@ -102,10 +102,10 @@ module HTTP
       # http://greenbytes.de/tech/webdav/rfc7230.html#rfc.section.3.3.3
       # Clause 3: "If a message is received with both a Transfer-Encoding
       # and a Content-Length header field, the Transfer-Encoding overrides the Content-Length.        
-      return nil if @headers.include?(Headers::TRANSFER_ENCODING) ||
-                    !@headers.include?(Headers::CONTENT_LENGTH)
+      return nil if @headers.include?(Headers::TRANSFER_ENCODING)
 
       value = @headers[Headers::CONTENT_LENGTH]
+      return nil unless value
 
       begin
         Integer(value)
--- a/lib/http/response/body.rb
+++ b/lib/http/response/body.rb
@@ -55,7 +55,7 @@ module HTTP
 
           length = @length || Float::INFINITY
 
-          while (length > 0 && chunk = @stream.readpartial)
+          while length > 0 && (chunk = @stream.readpartial)
             length -= chunk.bytesize
             @contents << chunk.force_encoding(encoding)
           end
--- a/lib/http/response/body.rb
+++ b/lib/http/response/body.rb
@@ -15,7 +15,7 @@ module HTTP
       # @return [HTTP::Connection]
       attr_reader :connection
 
-      def initialize(connection, stream, :length => nil, :encoding => Encoding::BINARY)
+      def initialize(connection, stream, length: nil, encoding: Encoding::BINARY)
         @connection = connection
         @streaming  = nil
         @contents   = nil
--- a/lib/http/response/body.rb
+++ b/lib/http/response/body.rb
@@ -15,7 +15,7 @@ module HTTP
       # @return [HTTP::Connection]
       attr_reader :connection
 
-      def initialize(connection, stream, encoding: Encoding::BINARY, length: nil)
+      def initialize(connection, stream, :length => nil, :encoding => Encoding::BINARY)
         @connection = connection
         @streaming  = nil
         @contents   = nil
--- a/lib/http/response/body.rb
+++ b/lib/http/response/body.rb
@@ -15,12 +15,13 @@ module HTTP
       # @return [HTTP::Connection]
       attr_reader :connection
 
-      def initialize(connection, stream, encoding = Encoding::BINARY)
+      def initialize(connection, stream, encoding: Encoding::BINARY, length: nil)
         @connection = connection
         @streaming  = nil
         @contents   = nil
         @stream     = stream
         @encoding   = encoding
+        @length     = length || Float::INFINITY
       end
 
       # (see HTTP::Client#readpartial)
@@ -53,7 +54,7 @@ module HTTP
           @streaming  = false
           @contents   = String.new("").force_encoding(encoding)
 
-          length = @length || Float::INFINITY
+          length = @length
 
           while length > 0 && (chunk = @stream.readpartial)
             length -= chunk.bytesize
--- a/lib/http/response/body.rb
+++ b/lib/http/response/body.rb
@@ -54,6 +54,10 @@ module HTTP
           @contents   = String.new("").force_encoding(encoding)
 
           while (chunk = @stream.readpartial)
+          length = @length || Float::INFINITY
+
+          while (length > 0 && chunk = @client.readpartial)
+            length -= chunk.bytesize
             @contents << chunk.force_encoding(encoding)
           end
         rescue
--- a/lib/http/response/body.rb
+++ b/lib/http/response/body.rb
@@ -53,10 +53,9 @@ module HTTP
           @streaming  = false
           @contents   = String.new("").force_encoding(encoding)
 
-          while (chunk = @stream.readpartial)
           length = @length || Float::INFINITY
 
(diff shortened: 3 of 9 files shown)


TiagoCardoso1983:
@tarcieri we can continue the discussion here.

Checking ruby's source code, I see the same ["faulty" wait_* methods handling which caused the known bug](https://github.com/ruby/ruby/blob/trunk/lib/net/protocol.rb#L153-L169). The main difference being, [net/http closes connections by default](https://github.com/ruby/ruby/blob/trunk/lib/net/http.rb#L1401-L1404). 

In the case of this gem, the preferred default is Keep-Alive, which is a design choice. 

This is the sequence of events which randomly triggers the issue (refers to https://github.com/httprb/http/blob/master/lib/http/timeout/per_operation.rb#L61-L72):

* ```read_nonblock(size, :exception => false)``` returns ```:wait_readable```
* ```wait_readable(read_timeout)``` returns nil.
  * the current code (as net/http) interprets nil as timeout, when according to the documentation, can also mean "EOF is reached". As it is, it ends in a timeout, although the whole response has been buffered. 
  *  if I choose to ignore nil and let it run (I interpret it as "EOF is reached"), next ```read_nonblock(size, :exception => false)``` will return ```nil```, and this means also "EOF is reached", therefore would be successfully completed.

So, this might have been already fixed in ruby trunk (according to the code it now [nly handles time outs](https://github.com/ruby/ruby/blob/trunk/ext/io/wait/wait.c#L123-L125), although implementation could mean something else), which means we can't rely on the ```false``` return value. 

So, I see these possible solutions:

* first ```nil for timeout``` is ignored, only "marked for timeout", and if the read_nonblock delivers anything other than nil, we trigger the timeout (otherwise EOF);

```ruby
tout = false
loop do                                                               
  result = @socket.read_nonblock(size, :exception => false)           
                                                                      
  return :eof if result.nil?                                        
  raise TimeoutError, "Read timed out after #{read_timeout} seconds" if tout
  return result if result != :wait_readable                           
                                                                      
  unless @socket.to_io.wait_readable(read_timeout)                    
    tout = true
  end                                                                 
end                                                                   
```
* [gem could follow the HTTP spec and use the header length to fetch only the necessary body length](http://greenbytes.de/tech/webdav/rfc7230.html#message.body.length). From what I understood, this gem (as net/http) is using a default window size to read content from socket to buffer, and uses io events to signal end of read/write tasks. This seems wrong (is it?) and logic could be rewritten use the window to read to buffer only until the Message-Length value is exhausted. This probably has to be handled differently for chunked responses. 

What do you think? First option seems to demand less overhead, but second option seems more correct, but might demand more internal refactoring.

tarcieri:
@TiagoCardoso1983 for what it's worth, a GSoC student has been working on an implementation of a Java-like (fixed-sized) ByteBuffer type which bypasses Ruby I/O and does native I/O on native buffers (or at least that's the goal, some implementations still pending, but GSoC isn't over yet): https://github.com/celluloid/nio4r/pull/95/files

TiagoCardoso1983:
@tarcieri that's good to hear, but is it cancelling this fix? IMO it shouldn't be hard to get something which just works with the stdlib, even with the current weird ```io/wait``` semantics. What do you think of the 2 suggestions?

tarcieri:
@TiagoCardoso1983 I don't entirely understand what you're proposing in regard to the HTTP semantics you describe. I guess I'll have to further study your links to have an opinion.

Regarding MRI async I/O errata, this is the sort of thing that really feels like it needs regression tests. I would really like to capture the previous regressions we've seen in an integration test.

TiagoCardoso1983:
Going forward with it, and judginy by [this](https://github.com/httprb/http/blob/61d8b5c8a3b94004f221027a87e4bd5f9e59d9bb/lib/http/client.rb#L62-L75) and [this](https://github.com/httprb/http/blob/2fa9ec4965d27b07cbeeb7fab972e33c314cb199/lib/http/response.rb#L44), the Response object has access to the response headers before reading the http response body, hence the 'Message-Length'. We could therefore pass this value to the Response Body object, which means that the response body would know when to stop reading the body without relying in the IO events. 
Anyway, this is all theoretical by now and I'll try to add a proof-of-concept for the failing request.

tarcieri:
@TiagoCardoso1983 I can't find any information on this `Message-Length` header you keep talking about. Can you point to a specific part of the HTTP RFC where it is described?

Regardless, you still need to handle cases where these sorts of headers aren't provided by non-compliant servers.

TiagoCardoso1983:
I already linked in a message above. And it's ```Content-Length``` .

tarcieri:
@TiagoCardoso1983 we already support `Content-Length`, however it's not always available

tarcieri:
Ok, looking at the code again it seems the response `Content-Length` is no longer used when reading the response body. Once upon a time it used to be.

That said, not all servers send it or even know it in advance, so at best it can be used opportunistically.

TiagoCardoso1983:
@tarcieri can you have a look at the rfc link I've linked a few messages above? There are some corner cases regarding the response body length which this gem might not address, besides that one.

Regarding the example from the related issue (the https link to google play store), which issues a 404 status code, the response comes with "Transfer-Encoding: chunked". The corrected(?) link which responds to 200 has the Content-Length set. Use curl for these two links to see it for yourself:

* https://play.google.com/store/apps/category/GAMES/collection/top?start=400 (404)
* https://play.google.com/store/apps/category/GAME/collection/topselling_free?start=400 (200)

The way I see it, even if not being fully compliant with the RFC, this gem could at least handle these two cases, in which when length is present, you only read as much, and when response is chunked, you read until the termination character (here's how it's defined in [rack](https://github.com/rack/rack/blob/master/lib/rack/chunked.rb#L21-L31). 

I don't agree with what you said. Servers which do not know content length in advance, encode the response as chunked. If you decide to ignore those headers because some non-compliant servers exist, you are opening yourself to a DoS vector, in that you read the body indefinitely until the "malicious" server stops sending data (right?).

TiagoCardoso1983:
Just for comparison sake, here's how net-http reads the body: https://github.com/ruby/ruby/blob/trunk/lib/net/http/response.rb#L280-L301

tarcieri:
Yes, it should use `Content-Length` when available.

TiagoCardoso1983:
@tarcieri I think I got it this time, and the failing example from #298 


So, the ```Content-Length``` issue is now handled, but the failing case which caused the revert was on a chunked response. I investigated how this happens here, and it seems that the chunks are read by the connection and passed to the parser (```HTTP::Parser``` instance). This will handle the chunked responses internally and give back a proper string back, which will be added to the contents. 

The caveat on the ```http``` side is that it relies on io events to decide what to do next. I don't think that is correct, but I don't know what is the correct alternative. So here I'll be shooting in the air from now on:

* ```http-parser``` object should provide a way to say it is done handling the chunks. I saw that the output coming from the socket contained the termination character ```0\r\n``` and the parser correctly concatenated the output, but it doesn't provide a hook stating that body parsing is complete. If it did, the client could terminate body parsing as per the spec instead of "waiting on the socket". Do you know if it supports this already?
* Either ```io/wait```doesn't support the case in which the socket eof's, or there was a race condition between reading/waiting which ```IO.select``` somehow bypasses. Check this example

```ruby
require 'socket'
require 'io/wait'

Thread.new do
  server = TCPServer.new 2000 # Server bind to port 2000
  loop do
    client = server.accept    # Wait for a client to connect
    client.puts "Hello !"
    client.puts "Time is #{Time.now}"
    sleep 3
    client.close
  end
end

puts "server is starting..."
sleep 2

puts "connecting now"
s = TCPSocket.new 'localhost', 2000

puts "connected"

10.times do
  k = s.wait_readable(2)
  puts "wait: #{k.inspect}"
  v = s.read_nonblock(8, exception: false)
  puts "val: #{v.inspect}"

  break if v == nil  
end

```

I will fill a bug to see what the core team thinks, but this makes indeed leveraging ```io/wait``` harder.

TiagoCardoso1983:
https://bugs.ruby-lang.org/issues/12674

tarcieri:
@TiagoCardoso1983 I believe the `on_message_complete` callback should fire at the end of each chunk, and an empty chunk should signify the end of the body.

lib/http/errors.rb
```diff
@@ -20,4 +20,7 @@ class TimeoutError < Error; end
 
   # Header name is invalid
   class InvalidHeaderNameError < Error; end
+
+  # Header value is of unexpected format (similar to Net::HTTPHeaderSyntaxError)
+  class HeaderSyntaxError < Error; end
```
ixti: What do you think of changing this to:

``` ruby
class HeaderError < Error; end
class InvalidHeaderNameError < HeaderError; end
class InvalidHeaderValueError < HeaderError; end
```

TiagoCardoso1983: humm, this one is tricky... I think the class of errors that can come from headers are not covered by these 2. Maybe the separator is wrong, for example. Also, this is in theory a ResponseError, as it's not a request header, and etc. 

IMO I think the best solution could be just to create a simple HeaderError and use it everywhere, and add the details in the message. Unless I'm missing the motive for so much error granularity. 

Let me know what you think, I can change to your suggestion for now, as soon as I come back to this. I work behind a proxy during the week, which is contatenating chunked responses, thereby blocking my use case.

ixti: Mmm. I tend to agree that single `class HeaderError < Error; end` should be enough.
But in order to make it backward compatible I propose to do this:

``` ruby
class HeaderError < Error; end

# @deprecated
InvalidHeaderNameError = HeaderError
```

lib/http/response.rb
```diff
@@ -92,13 +92,18 @@ def flush
...
+      return nil if headers[Headers::TRANSFER_ENCODING]
+
       value = @headers[Headers::CONTENT_LENGTH]
       return unless value
 
       begin
         Integer(value)
       rescue ArgumentError
-        nil
+        raise(HeaderSyntaxError, 'wrong Content-Length format')
```
britishtea: Can you explain why it's important that an error is raised here? It seems that a faulty `Content-Length` header will now prevent the `HTTP::Response` from being even instantiated.

See also the discussion here: https://github.com/httprb/http/pull/365.

ixti: I agree. I think logic in that part should remain as is. Also I think that if Content-Length is given with wrong value we should treat it as "no content length" given. I agree that it's probably pretty rare edge case, but if it will happen I think it's developer who should decide how to handle that (read till the end or terminate) not our gem.

TiagoCardoso1983: I tried to follow the [spec](http://greenbytes.de/tech/webdav/rfc7230.html#message.body.length), point 4:

> If a message is received without Transfer-Encoding and with either multiple Content-Length header fields having differing field-values or a single Content-Length header field having an invalid value, then the message framing is invalid and the recipient must treat it as an unrecoverable error

TiagoCardoso1983: @britishtea how would it prevent the response from being instantiated? `#content_length` is an instance method from it, therefore you need the instance to call the method, right?

britishtea: @TiagoCardoso1983 `#content_length` is called in `#initialize` (line 51). The exception is not rescued, so `HTTP::Response.new` will fail.

TiagoCardoso1983:
So I've reverted the exception for content length, and it returns nil, and the response can be instantiated. Which means that the body will be fetched until end of connection, potentially ad eternum. 

Besides the security reason there, I think that this feature doesn't conform to the HTTP/1.1 spec, and even [net/http does it](https://github.com/ruby/ruby/blob/trunk/lib/net/http/header.rb#L276-L281). I don't know why one needs to instantiate the response if the response/headers is/are invalid, but then again, I only contributed a few commits and you know more about the usage history of this gem. 

In case one wants to satisfy both requirements, the way the response is instantiated has to change, and this should maybe be the task of the core team.

tarcieri:
@TiagoCardoso1983 chalk it up to Postel's law. That said, the "security" issue you're describing exists for a compliant implementation which follows an endless chunked response and exists in every compliant HTTP/1.1 client.

lib/http/response.rb
```diff
@@ -92,13 +92,17 @@ def flush
...
+      return nil if headers[Headers::TRANSFER_ENCODING]
+
       value = @headers[Headers::CONTENT_LENGTH]
       return unless value
 
       begin
         Integer(value)
       rescue ArgumentError
-        nil
       end
```
ixti: Can you, please, return `nil` in here. It will make rubocop a bit pleased ;))

lib/http/response.rb
```diff
@@ -92,13 +92,17 @@ def flush
     #   (not an integer, e.g. empty string or string with non-digits).
     # @return [Integer] otherwise
     def content_length
+      # http://greenbytes.de/tech/webdav/rfc7230.html#rfc.section.3.3.3
+      # Clause 3: "If a message is received with both a Transfer-Encoding
+      # and a Content-Length header field, the Transfer-Encoding overrides the Content-Length.        
+      return nil if headers[Headers::TRANSFER_ENCODING]
+
       value = @headers[Headers::CONTENT_LENGTH]
       return unless value
```
ixti: I know it was empty before, but IMHO it worth to tell explicitly `return nil` here, so it will be similar to other places of this method :D

lib/http/response.rb
```diff
@@ -125,6 +129,15 @@ def cookies
       end
     end
 
+    def chunked?
+      encoding = headers.get(Headers::TRANSFER_ENCODING)
```
ixti: `Headers#get` always return an `Array`, even if there's no such header being set.

lib/http/response.rb
```diff
@@ -125,6 +129,15 @@ def cookies
...
+      case encoding
+        when String then encoding == CHUNKED
+        when Enumerable then encoding.include?("chunked")
+        else false
+      end
```
ixti: If I understand correctly, what you actually need here is:

``` ruby
@headers.get(Headers::TRANSFER_ENCODING).include?(CHUNKED)
```

lib/http/response.rb
```diff
@@ -92,13 +92,17 @@ def flush
     #   (not an integer, e.g. empty string or string with non-digits).
     # @return [Integer] otherwise
     def content_length
+      # http://greenbytes.de/tech/webdav/rfc7230.html#rfc.section.3.3.3
+      # Clause 3: "If a message is received with both a Transfer-Encoding
+      # and a Content-Length header field, the Transfer-Encoding overrides the Content-Length.        
+      return nil if headers[Headers::TRANSFER_ENCODING]
```
ixti: can you please either change this to be variable `@headers` or below change variable to getter `headers`

ixti: also, I have just pushed small improvement to `Headers` class, so this line can benefit of using it:

``` ruby
return nil if @headers.include?(Headers::TRANSFER_ENCODING)
```

`Headers#include?` is more efficient (does not allocates new objects in comparison to `#[]`)

tarcieri: I would suggest `@headers` to avoid a needless method invocation

lib/http/response.rb
TiagoCardoso1983: @ixti , you are right, it's a needless case statement. However, consider the [RFC](http://greenbytes.de/tech/webdav/rfc7230.html#message.body.length) and help me figure out point 3:

> If a Transfer-Encoding header field is present in a response and the chunked transfer coding is not the final encoding, the message body length is determined by reading the connection until it is closed by the server. If a Transfer-Encoding header field is present in a request and the chunked transfer coding is not the final encoding, the message body length cannot be determined reliably; the server must respond with the 400 (Bad Request) status code and then close the connection.

Should one tag the response as `chunked?` if it is not the last encoding? I'll add it, unless you feel differently.

TiagoCardoso1983:
@tarcieri that is correct, I meant more as an unexpected security concern, as the protocol defines those use cases. My goal is just to document it in case future issues arise from it.

lib/http/response.rb
ixti: Oh. Now I see... Well, according to RFC logic should be a bit more "complex"

``` ruby
encodings = @headers.get(Headers::TRANSFER_ENCODING)

if CHUNKED == encodings.last
  read_till_last_chunk
elsif encodings.include?(CHUNKED)
  read_till_server_closes_connection
else
  read_content_length_amount_of_byte_or_till_nothing_got_sent
end
```

Although, I think we can simplify that to:

``` ruby
if CHUNKED == @headers.get(Headers::TRANSFER_ENCODING).last
  read_till_last_chunked
else
  read_as_usual
end
```

Leading us towards:

``` ruby
def chunked?
  CHUNKED == @headers.get(Headers::TRANSFER_ENCODING).last
end
```

ixti: In fact the only thing that annoys me in "complex" logic of pseudo-code above is how it will work with persistent connection (AKA `Keep-Alive`) :D Probably we can actually implement it and raising `ResponseError` if connection is persistnet and transfer encoding is chunked (but it's not final encoding) - I believe it's the most retarded and rare case ever :D

lib/http/response.rb
```diff
@@ -96,8 +96,13 @@ def flush
     #   (not an integer, e.g. empty string or string with non-digits).
     # @return [Integer] otherwise
     def content_length
+      # http://greenbytes.de/tech/webdav/rfc7230.html#rfc.section.3.3.3
+      # Clause 3: "If a message is received with both a Transfer-Encoding
+      # and a Content-Length header field, the Transfer-Encoding overrides the Content-Length.        
+      return nil if @headers.include?(Headers::TRANSFER_ENCODING) ||
+                    !@headers.include?(Headers::CONTENT_LENGTH)
```
ixti: I would split those in two separate statements for sake of simplicity (took me a while to understand that this is correct statement -- was pretty sure there's a mistake) as it was.

``` ruby
return nil if @headers.include?(Headers::TRANSFER_ENCODING)

value = @headers[Headers::CONTENT_LENGTH]
return nil unless value

# ...
```

Calling `#include?` and then `#[]` will involve extra call in case of "happy path". Although `#include?` is more efficient than `#[]`, when you end up having those two - it worth to keep `#[]` only (at least in this particular place). Just to explain:

``` ruby
def Headers#include?(name)
  @headers_array.any? { |key, val| key == name }
end

def Headers#[](name)
  values = @headers_array.each_with_object([]) { |(k, v), a| a << v if k == name }
  return nil if values.empty?
  return values.first if 1 == values.count
  values
end
```

TiagoCardoso1983: I see. Actually what one needs is something equivalent do `Enumerable#find`. As Headers is Enumerable, this shouldn't be an issue.

ixti: Yeah. You can call find on Headers instance:

``` ruby
HTTP.get("https://ya.ru").headers.find { |key, val| "Content-Length" == key } 
# => ["Content-Length", "9578"]
```

ixti: In fact I think toward changing the way headers are kept internally, that will make all operations on Headers fast and efficient. Right now it's an array of `[key, val]` arrays.
The only good thing about this format is that it allows "recreate" headers list as is. BUT. We normalize header names, so it's not actually guaranteed. So I lean towards changing data under the hood to be `Hash{String, Array<String>}`. That will make most operations `O(1)`.

ixti: The only thing is that if that is gonna happen (change to underlying data structure) it will be released only as next major release due to introduction of backward incompatible changes.

tarcieri: Time for 3.0 already, eh? Maybe we can drop support for Ruby 2.0 and 2.1 while we're at it...

ixti: Mmm. I'm in! :D

tarcieri: I might also try extracting the timeout code into a separate gem...

ixti: I was pretty sure @zanker was doing something related to that, although I might misunderstood something.

tarcieri: Pretty sure he abandoned his efforts.

ixti: Oh. Ok.

TiagoCardoso1983: Indeed. Just out of curiosity, what's the idea of normalizing the header key?

lib/http/response.rb
```diff
@@ -99,10 +99,10 @@ def content_length
       # http://greenbytes.de/tech/webdav/rfc7230.html#rfc.section.3.3.3
       # Clause 3: "If a message is received with both a Transfer-Encoding
       # and a Content-Length header field, the Transfer-Encoding overrides the Content-Length.        
-      return nil if @headers.include?(Headers::TRANSFER_ENCODING) ||
-                    !@headers.include?(Headers::CONTENT_LENGTH)
+      return nil if @headers.include?(Headers::TRANSFER_ENCODING)
 
       value = @headers[Headers::CONTENT_LENGTH]
+      return nil unless value
```
ixti: :bow:

TiagoCardoso1983: had to google that emoji, it's tiny 😃

tarcieri:
@HoneyryderChuck this still looks like good stuff (at least for 3.0). Mind rebasing?

lib/http/response/body.rb
```diff
@@ -53,7 +53,10 @@ def to_s
           @streaming  = false
           @contents   = String.new("").force_encoding(encoding)
 
-          while (chunk = @stream.readpartial)
+          length = @length || Float::INFINITY
```
HoneyryderChuck: @tarcieri the purpose of this variable got lost in the rebase. I was passing the content length (when available) to the body to prevent mismatches and eventual infinite loops. But the initialize signature changed. I leave the decision to you to take advantage of this or remove it.

HoneyryderChuck:
@tarcieri done. Although, I got the impression that this bit was supposed to go into socketry eventually. Did that change?

tarcieri:
@HoneyryderChuck I'd still like to do it eventually, but this is lower hanging fruit

Looks like you have some RuboCop violations though

tarcieri:
Also I'll be out for the next week or so and won't have time to review. Perhaps @ixti can though.

We'll also need to start treating `master` as `3.x` if we merge this.

HoneyryderChuck:
@tarcieri @ixti done!

lib/http/errors.rb
```diff
@@ -18,6 +18,9 @@ class StateError < ResponseError; end
   # Generic Timeout error
   class TimeoutError < Error; end
 
-  # Header name is invalid
-  class InvalidHeaderNameError < Error; end
+  # Header value is of unexpected format (similar to Net::HTTPHeaderSyntaxError)
+  class HeaderError < Error; end
+
+  # @deprecated will be removed in v3.0.0
```
ixti: As we're gonna release this PR as part of `v3.0.0` it worth to change this to be *will be removed in v3.1.0*

lib/http/response/body.rb
ixti: I think we should follow your initial idea of limiting amount of data to given Content-Length when available and should be respected according to RFC.

HoneyryderChuck: @ixti agree, but how should I "import" the length? The initializer signature changed, and I lost my assumptions. Just in case this PR takes again longer than planned: what do you suggest as the long term way to pass the content length from the response to the body?

When I implemented it, I changed the signature to use kwargs and added the encoding and content length as optional parameters. Can I go back to that strategy? as @tarcieri said, this is anyway planned to a major bump, so in theory I could.

ixti: Yes, go ahead. I'm fine with keyword args. And it won't take long this time, I promise - will merge as soon as it's ready :D before this weekend.

lib/http/errors.rb
HoneyryderChuck: But if it is to be removed in 3.0.0 and this PR is about 3.0.0, then why not just go and remove support? I'd save this tweak to upstream though, it's beyond the scope of the PR IMO.

ixti: Mmm. I guess I'm agree. Just remove this deprecated class then.

lib/http/response/body.rb
HoneyryderChuck: > will merge as soon as it's ready :D before this weekend.

Famous last words :D 

I have a tentative commit here which I don't know if I'll include: I was going to make `Response#content_length` return `Float::INFINITY` when header not explicitly set. It's a bit avant-garde, since it's not completely correct, as 304 and 204 technically have content length 0. This could be arranged, though. This would make the method semantically more correct, but it's public API that could break some clients. Which could be ok, as it's v3, so breaking is acceptable if justified. 

I pass the potato to you.

HoneyryderChuck:
can I make a few suggestions regarding rubocop? Because, if you notice the ammount of useful commits against rubocop tweaks in this PR, you can realize how annoying that might have been. I get the usefulness of maintaining a standard, but there are some inconsistencies which make one lose focus. Main one being, the hash-rocket insistence. Main reasons are:

* you can't use kwargs **and** hash-rockets at the same time. 
```ruby
# syntax error
def bla(:a => nil, :b => nil)
```
* You can pass hash-rocket like parameters as kwargs
```ruby
# if I don't call #bla as this, rubocop complaints
bla(:a => 1, :b => 2)
```

I also suggest dropping this spec given that minimum requirement of this gem is ruby 2.0, and the "new" hash syntax exists already since 1.9 .

ixti:
Will merge this today.
Re hash syntax: https://github.com/httprb/http/pull/342

HoneyryderChuck:
> Will merge this today.

hopefully you make it, as "before the weekend" just passed ;) not to sound like an a-hole, but I've been mostly fighting conflicts and rubocop for the last week, which is time and nerve-consuming, for a gem I haven't been using lately.

tarcieri:
@ixti remember to push master to `2-x-stable` prior to merging this, as it's a breaking change, so with this `master` will officially become the 3.x development branch.

We should probably also add some README text to clarify the version numbers and development state of each branch (I can take care of that)

@HoneyryderChuck I'm 👍 on moving to the new hash syntax across the board, @ixti is the detractor there.

re: RuboCop, are you aware of `rubocop -a`? (which will autofix your code)

HoneyryderChuck:
@tarcieri I actually wasn't, thx for the tip 👍

ixti:
I have squashed and merged this: https://github.com/httprb/http/commit/cb44c5dfc7fe2bc7d81a1465a561e5b8764749f6

ixti:
@tarcieri 2-x-stable was actually created as soon as we released v2.0.0, and now it's in sync with v.2.2.1, while master contains this PR

ixti:
Re: hash syntax. Yes. I am strongly against new hash syntax.

tarcieri:
@ixti yeah, just making sure it was up-to-date before merging this. I'll add some clarifying information to the READMEs tonight.

re: hash syntax, I'm only weakly in favor of the new syntax, so I guess your strong objection wins 😜

HoneyryderChuck:
@ixti respect your opinion, and as one of the moderators of this project, your counts the most. In terms of creating arbitrary hashes, I have to say that I don't have a strong enough counter-argument, since as stated in the linked issue, you can't represent non-symbol types with the new syntax. 

However, I think one could "relax" the policy and open an exception, namely when you pass the options directly in the method signature. From what I understood, you're not against kwargs. I'd suggest "enforcing" them in the internal http API (at least), as it adds argument-not-type checks on method call which map to more meaningful error messages, and will remove options hash custom handling [like this one bit here (see #fetch calls)](https://github.com/httprb/http/blob/master/lib/http/response.rb#L41-L56). Makes a maintainer's life much easier and lessens "primitive obsession". 

That being said, I think that insisting in the old hash-rocket *for method calls* creates confusion, as there is dissonance between defining a method and calling a method:

```ruby
def person(first: , last: )
...

# same
person(first: "Dirty", last: "Harry")
# not same
person(:first => "Dirty", :last => "Harry")
```

To summarize, I agree with you in the hash thing, but the fact that you can't define kwargs in an hash-rocket-compatible forces you to handle the hash syntax inconsistency one one or another. The only way to prevent this would be not to use kwargs at all, but I hope I made a strong argument about why you **should** use kwargs. And therefore, I think that the new syntax could be accepted as long as the new syntax could only be used in method calls directly and not to instantiate hashes. Whether rubocop handles such an exception, that's another topic.

Those are my 2 cents. As stated, I respect the logic of the choice.

ixti:
@HoneyryderChuck We're all for kwargs. It's just we dropped Ruby [*relatively* not so long ago](https://github.com/httprb/http/blob/master/CHANGES.md#200-2016-04-23), so nobody was working on rewriting internal API. I believe that's a great idea (to get rid of opts hashes) for 3.0.0 release.

I agree with you that current restriction introduces a small confusion (in method calling vs method definition), but I find it not so bad, after all method definition differs from method calling in general, e.g.:

``` ruby
def say(what = "wut?")
  # ...
end

# you can:
say("hi!")

# but can't (obviously)
say(what = "hi")
```

So I prefer to look at this particular problem (of kwargs definition/usage difference) through this prism :D Relaxing of rubocop on has syntax will add extra burden on those who review to check styling as well as one will have to verify if used Hash is not a hash but kwargs.

tarcieri:
I think, at least, we can enable the `OptionHash` cop so as to forbid option hashes:

https://github.com/bbatsov/rubocop/pull/2078/files

PR #369, opened 2016-08-12
Test on ActiveModel 4.x

janko-m:
ActiveModel 5.x doesn't install on Ruby versions lower than 2.2.2.


tarcieri:
These are semantically equivalent and `master` is already passing post-Rails 5. If your branch is failing you just need to rebase.

janko-m:
Ok, gotcha!

PR #377, opened 2016-09-24
[WIP] Convert async I/O backend to Socketry

tarcieri:
Socketry uses async I/O to implement thread-safe timeouts:

https://github.com/socketry/socketry

cc @ixti @zanker @TiagoCardoso1983


lib/http/connection.rb
```diff
@@ -37,13 +37,22 @@ def initialize(req, options)
...
+      end
+
+      @socket.connect(req.socket_host, req.socket_port)
+      @socket.nodelay = options.nodelay if options.nodelay
 
       send_proxy_connect_request(req)
-      start_tls(req, options)
       reset_timer
-    rescue IOError, SocketError, SystemCallError => ex
+    rescue Socketry::Error => ex
```
tarcieri: There are probably more places this sort of change needs to be made

tarcieri:
Note: Socketry's timeouts can subsume `HTTP::Timeout::Null` and `HTTP::Timeout::PerOperation`, but `HTTP::Timeout::Global` will need to be (re)implemented.

tarcieri:
Note per cb420da this is a breaking change which will require a new major release (3.x) as among other things it will require Ruby 2.2+ due to the Socketry dependency.

Since Rails 5 is also 2.2+, hopefully this is ok.

lib/http/connection.rb
```diff
@@ -142,22 +151,6 @@ def expired?
 
     private
 
-    # Sets up SSL context and starts TLS if needed.
-    # @param (see #initialize)
-    # @return [void]
-    def start_tls(req, options)
-      return unless req.uri.https? && !failed_proxy_connect?
```
zanker: I suspect it is, I didn't add that at least.

tarcieri:
Well, still a WIP (especially timeout support) but it's passing consistently on 2.3 and JRuby 9.1.6.0 (and sporadically on 2.2)

tarcieri:
If anyone wants to resume work on this, let me know.

PR #381, opened 2016-11-28
Include proxy authentication headers for CONNECT proxy requests

somethingnew2-0:


lib/http/request.rb
```diff
@@ -106,7 +106,7 @@ def redirect(uri, verb = @verb)
 
     # Stream the request to a socket
     def stream(socket)
-      include_proxy_headers if using_proxy? && !@uri.https?
```
zanker: Is there a reason this wouldn't have been enabled previously? I can't think of one off the top of my head.

somethingnew2-0: It should be possible according to RFC 2817 to add a Proxy-Authorization header to a CONNECT request, I believe. https://tools.ietf.org/html/rfc2817#section-5.2

Connorhd: I believe that in a https connection this is the SSL stream, not the CONNECT request, so the proxy headers should have already been sent with the CONNECT (see proxy_connect_headers), not through the encrypted stream

somethingnew2-0: Oh, I think you're correct. https://github.com/httprb/http/blob/master/lib/http/request.rb#L161

PR #383, opened 2016-12-21
Fixing up the #timeout docs on Chainable

ramfjord:
A minor typo makes it very difficult to parse the docs for the `#timeout` method: http://www.rubydoc.info/github/httprb/http/HTTP/Chainable#timeout-instance_method

This just correct the `@overload` declarations and adds a tiny bit more info.


tarcieri:
Thanks!

PR #390, opened 2017-01-22
Add REPORT as valid HTTP verb

### LINKED ISSUE #387: Support for CardDAV protocol
Would like to request HTTP gem playing well with rfc6352 CardDAV protocol. REPORT method is not supported.

ixti:
Resolves #387.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,10 +2,14 @@ language: ruby
 sudo: false
 
 before_install:
-  - gem update --system
-  - gem install bundler
+  - gem update --system 2.6.8
+  - gem --version
+  - gem install bundler --version 1.13.7 --no-rdoc --no-ri
+  - bundle --version
 
-bundler_args: --without development doc
+install: bundle _1.13.7_ install --without development doc
+
+script: bundle _1.13.7_ exec rake
 
 env:
   global:
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,14 +12,21 @@ env:
     - JRUBY_OPTS="$JRUBY_OPTS --debug"
 
 rvm:
-  - jruby-9.1.6.0
   - 2.0.0
   - 2.1
   - 2.2
   - 2.3.0
+  - 2.3.1
+  - 2.3.2
+  - 2.3.3
   - 2.4.0
+  - ruby-head
+  - jruby-9.1.6.0
+  - jruby-9.1.7.0
 
 matrix:
+  allow_failures:
+    - rvm: ruby-head
   fast_finish: true
 
 branches:
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,6 +2,7 @@ language: ruby
 sudo: false
 
 before_install:
+  - gem update --system
   - gem install bundler
 
 bundler_args: --without development doc
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,21 +12,14 @@ env:
     - JRUBY_OPTS="$JRUBY_OPTS --debug"
 
 rvm:
+  - jruby-9.1.7.0
   - 2.0.0
   - 2.1
   - 2.2
-  - 2.3.0
-  - 2.3.1
-  - 2.3.2
   - 2.3.3
   - 2.4.0
-  - ruby-head
-  - jruby-9.1.6.0
-  - jruby-9.1.7.0
 
 matrix:
-  allow_failures:
-    - rvm: ruby-head
   fast_finish: true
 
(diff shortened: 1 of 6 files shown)


ixti:
#387 issue and this PR actually brings us to the need of rethinking how we validate HTTP verbs. `REPORT` used in [RFC 6352](https://tools.ietf.org/html/rfc6352#section-8.1) extends [RFC 3253](https://tools.ietf.org/html/rfc3253). And taking a glance look at last one it's obvious that our current way will fail on [`VERSION-CONTROL`](https://tools.ietf.org/html/rfc3253#section-6.7.1) method.

ixti:
@tarcieri you might wanna about some of the changes made with this pr ;D

ixti:
@tarcieri I tend to merge this PR as failures on jruby now are pretty specific to travis and bundler. Somhow it simply fails to run on jruby on travis - but runs locally.

.travis.yml
```diff
@@ -11,14 +12,21 @@ env:
...
 
 rvm:
-  - jruby-9.1.6.0
   - 2.0.0
   - 2.1
   - 2.2
   - 2.3.0
+  - 2.3.1
+  - 2.3.2
+  - 2.3.3
```
tarcieri: What's the point in testing every version of 2.3 under the sun?

.travis.yml
```diff
@@ -11,14 +12,21 @@ env:
...
   - 2.4.0
+  - ruby-head
+  - jruby-9.1.6.0
+  - jruby-9.1.7.0
```
tarcieri: If you put jruby at the top it will start running first... which is a good thing as it takes the longest

.travis.yml
```diff
@@ -11,14 +12,21 @@ env:
...
-  - jruby-9.1.6.0
   - 2.0.0
   - 2.1
   - 2.2
   - 2.3.0
+  - 2.3.1
+  - 2.3.2
+  - 2.3.3
   - 2.4.0
+  - ruby-head
```
tarcieri: I used to be a fan of putting "ruby-head" in but nowadays I think it's just pointless

.gitignore
```diff
@@ -3,6 +3,7 @@
 .config
 .rvmrc
 .yardoc
+.ruby-version
```
tarcieri: What? I can see an argument for not having one, but why put it in .gitignore?

ixti: Adding `.ruby-version` to `.gitignore` so that nobody will commit it accidentally... :D I mean I'm fine when `.ruby-version` is committed in repo, when only that ruby version is supported (e.g. for some ruby app), but overwise it's as good as adding `ruby 2.4` line into `Gemfile`.

.travis.yml
ixti: Well, I tend to agree. I mean I think it would make sense if build would be triggered at least nightly if not on each ruby-head commit.

.travis.yml
ixti: Agreed!

.travis.yml
ixti: Honestly I have no answer on this question.

.gitignore
tarcieri: I've added a `.ruby-version` to most projects, with the latest version being supported. It automatically configures multiple development tools and is quite handy that way.

tarcieri:
Looks good... well aside from the JRuby failure

.gitignore
ixti: Somehow it was giving me more of a headache than helping. But I will revert that commit.

ixti:
@tarcieri master is also failing on jruby ;)) btw

tarcieri:
@ixti I'm asking on the Bundler Slack about it.

Maybe we should pin to a known-working Bundler.

tarcieri:
This appears to be the issue: https://github.com/bundler/bundler/issues/5340

tarcieri:
This will be fixed in Bundler 1.14.2

ixti:
Yeah. Thanks for digging into that. I have locked down bundler on travis for now to the one that worked last time for us ;))

ixti:
Okay. I have no clue how to lock bundler...

tarcieri:
Let me take a crack at just fixing that issue...

ixti:
After some time and couple of rounds when I thought I finally got insane - I have managed travis to run with bundler / rubygems known to work :D

tarcieri:
Looks good!

PR #392, opened 2017-01-23
Update Travis config to use Bundler 1.14.2

tarcieri:


ixti:
@tarcieri nope... 1.14.2 did not fixed the issue...

ixti:
Yeah, issue still not fixed (it's not only us affected): https://github.com/bundler/bundler/issues/5349

tarcieri:
Looks like they might have a path to a real fix though. I'll update for 1.14.3 when it comes out

PR #397, opened 2017-02-07
Only run RuboCop on 2.4.0

tarcieri:
Don't redundantly run RuboCop on every Ruby under the sun.

This will inadvertently fix the build as well. Woohoo! (Ok, not actually inadvertently)


ixti:
@tarcieri we should run yardstick on latest ruby only too

ixti:
Oh you already did that

PR #402, opened 2017-03-08
Add URI#dup method to fix a bug with shared internal instance variables

### LINKED ISSUE #388: HTTP::Uri#dup doesn't duplicate the internal url variable
I'm using `HTTP::Response#uri` method to get the final destination url during redirects, and I've been getting a strange bug when running it in production. 

When I make a request to url with query parameters, they are being cut off in `HTTP::Response#uri`.

```
resp = HTTP.get "http://google.com?xx=1"
resp.uri == "http://google.com"
```

Surprisingly, this bug disappears in the ruby console.

-----

During debugging, I've discovered that NewRelic gem was the source of the problem (that's why the bug hasn't been present in the console).

In short, NewRelic tries to duplicate `HTTP::Request#uri` object, in order to modify it for its own use.
(see: https://github.com/newrelic/rpm/blob/master/lib/new_relic/agent/cross_app_tracing.rb#L194 and https://github.com/newrelic/rpm/blob/master/lib/new_relic/agent/http_clients/uri_util.rb#L18)

But `HTTP::URI` doesn't implement `#dup` method correctly, and the internal `@url` variable is being reused in the duplicated copies of the URI objects.

Steps to reproduce
====

```ruby
# Initialize original url and duplicate it
orig = HTTP::URI.parse("http://example.org?xx=1")
copy = orig.dup

# Modify internal url in original
orig.query = nil
orig.to_s == "http://example.org"
```

## Actual results:

Duplicated uri has been also changed.

```ruby
copy.to_s == "http://example.org"
copy.query == nil
```

## Expected results:

Duplicated uri should've remained the same.

```ruby
copy.to_s == "http://example.org?xx=1"
copy.query == "xx=1"
```

---

Ruby version: `ruby 2.3.3p222 (2016-11-21 revision 56859) [x86_64-darwin16]`
Gem version: `http (2.1.0)` and `addressable (2.5.0)`

scarfacedeb:
Resolves #388.

### DIFF
--- a/lib/http/uri.rb
+++ b/lib/http/uri.rb
@@ -114,6 +114,11 @@ module HTTP
       HTTPS_SCHEME == scheme
     end
 
+    # @return [Object] duplicated URI
+    def dup
+      self.class.new @uri.dup
+    end
+
     # Convert an HTTP::URI to a String
     #
     # @return [String] URI serialized as a String
--- a/spec/lib/http/uri_spec.rb
+++ b/spec/lib/http/uri_spec.rb
@@ -18,4 +18,14 @@ RSpec.describe HTTP::URI do
   it "sets default ports for HTTPS URIs" do
     expect(https_uri.port).to eq 443
   end
+
+  describe "#dup" do
+    it "doesn't share internal value between duplicates" do
+      duplicated_uri = http_uri.dup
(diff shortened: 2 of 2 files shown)


ixti:
Thanks!

PR #405, opened 2017-04-12
Make keep_alive_timeout configurable

nestegg:
Allow keep_alive_timeout to be set in persistent:

HTTP.persistent("http...", timeout: 120)

### DIFF
--- a/lib/http/chainable.rb
+++ b/lib/http/chainable.rb
@@ -109,12 +109,13 @@ module HTTP
       )
     end
 
-    # @overload persistent(host)
+    # @overload persistent(host, timeout: nil)
     #   Flags as persistent
-    #   @param [String] host
-    #   @raise [Request::Error] if Host is invalid
+    #   @param  [String] host
+    #   @option [Integer] timeout Keep alive timeout
+    #   @raise  [Request::Error] if Host is invalid
     #   @return [HTTP::Client] Persistent client
-    # @overload persistent(host, &block)
+    # @overload persistent(host, timeout: nil, &block)
     #   Executes given block with persistent client and automatically closes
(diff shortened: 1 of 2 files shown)


nestegg:
This CI failure is unrelated to the changes.

lib/http/chainable.rb
```diff
@@ -138,8 +140,11 @@ def timeout(klass, options = {}) # rubocop:disable Style/OptionHash
     #
     #   @yieldparam [HTTP::Client] client Persistent client
     #   @return [Object] result of last expression in the block
-    def persistent(host)
```
ixti: We officially support ruby 2.0+, so it's better to use kwargs in here:

``` ruby
def persistent(host, timeout: nil)
  # ...
```

spec/lib/http_spec.rb
```diff
@@ -288,6 +289,14 @@
         end
       end
     end
+
+    context "with timeout specified" do
+      subject { HTTP.persistent host, :timeout => 100 }
+      it "sets keep_alive_timeout" do
+        options = subject.default_options
```
ixti: use named subject please:

``` ruby
subject(:client) { HTTP.persistent ... }
# ...
options = client.default_options
```

spec/lib/http_spec.rb
```diff
@@ -1,3 +1,4 @@
+# coding: utf-8
```
ixti: this duplicates line 3

nestegg:
This has been updated.

ixti:
Merged as https://github.com/httprb/http/commit/4738d22c4b615aba193710aee83cc34b5c4792de

PR #406, opened 2017-04-14
test uri behavior on travis

ixti:
**DO NOT MERGE**

### DIFF
--- a/spec/lib/http/uri_spec.rb
+++ b/spec/lib/http/uri_spec.rb
@@ -19,6 +19,13 @@ RSpec.describe HTTP::URI do
     expect(https_uri.port).to eq 443
   end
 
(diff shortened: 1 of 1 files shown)


olleolleolle:
(Triaging) Hi, @ixti! Is this still a relevant Issue to keep open?

ixti:
Mm. Thanks for reminder. Will check out today.

ixti:
Honesly I forgot what was the issue in fact. It works now locally for me... Not sure if behavior is what we desire, but it should be opened (if needed) as an issue:

``` ruby
HTTP::URI.parse("http://example.com:0000").port # => 80
URI.parse("http://example.com:0000").port # => 0
```

PR #409, opened 2017-05-01
Enable request body streaming with an IO object

janko-m:
This change adds the ability to stream content of an IO object (an object that responds to #read) into the request body. This is convenient when you want to send large amounts of data as the request body, but you don't want to load all of it into memory.

```rb
HTTP.post("http://example.org/upload", body: File.open("video.mp4"))
HTTP.post("http://example.org/upload", body: StringIO.new("file data"))
```

It was already possible to send the request body in chunks by passing an Enumerable to `:body`, but that can (and should) only be used for making "Transfer-Encoding: chunked" requests.

This will also allow us to extend form_data.rb to stream the multipart-encoded request, instead having to load the whole body into memory before writing it to the socket.

### DIFF
--- a/lib/http/request/body.rb
+++ b/lib/http/request/body.rb
@@ -3,9 +3,6 @@
 module HTTP
   class Request
     class Body
-      # Types valid to be used as body source
-      VALID_BODY_TYPES = [String, NilClass, Enumerable].freeze
-
       # Maximum chunk size used for reading content of an IO
       BUFFER_SIZE = Connection::BUFFER_SIZE
 
@@ -46,7 +43,11 @@ module HTTP
       private
 
       def validate_body_type!
-        return if VALID_BODY_TYPES.any? { |type| @body.is_a? type }
+        return if @body.is_a?(String)
+        return if @body.respond_to?(:read)
+        return if @body.is_a?(Enumerable)
+        return if @body.nil?
+
         raise RequestError, "body of wrong type: #{@body.class}"
       end
     end
--- a/lib/http/request/body.rb
+++ b/lib/http/request/body.rb
@@ -33,8 +33,10 @@ module HTTP
         if @body.is_a?(String)
           yield @body
         elsif @body.respond_to?(:read)
-          chunk = String.new # rubocop:disable Style/EmptyLiteral
-          yield chunk while @body.read(BUFFER_SIZE, chunk)
+          loop do
+            data = @body.read(BUFFER_SIZE) or break
+            yield data
+          end
         elsif @body.is_a?(Enumerable)
           @body.each { |chunk| yield chunk }
         end
--- a/lib/http/request/body.rb
+++ b/lib/http/request/body.rb
@@ -33,8 +33,7 @@ module HTTP
         if @body.is_a?(String)
           yield @body
         elsif @body.respond_to?(:read)
-          loop do
-            data = @body.read(BUFFER_SIZE) or break
+          while (data = @body.read(BUFFER_SIZE))
             yield data
           end
         elsif @body.is_a?(Enumerable)
--- a/lib/http/request/body.rb
+++ b/lib/http/request/body.rb
@@ -1,0 +1,54 @@
+# frozen_string_literal: true
+
+module HTTP
+  class Request
+    class Body
+      # Types valid to be used as body source
+      VALID_BODY_TYPES = [String, NilClass, Enumerable].freeze
+
+      # Maximum chunk size used for reading content of an IO
+      BUFFER_SIZE = Connection::BUFFER_SIZE
+
+      def initialize(body)
+        @body = body
+
+        validate_body_type!
+      end
+
+      # Returns size which should be used for the "Content-Length" header.
+      def size
+        if @body.is_a?(String)
+          @body.bytesize
+        elsif @body.respond_to?(:read)
+          raise RequestError, "IO object must respond to #size" unless @body.respond_to?(:size)
+          @body.size
+        elsif @body.nil?
+          0
+        else
+          raise RequestError, "cannot determine size of body: #{@body.inspect}"
+        end
+      end
+
+      # Yields chunks of content to be streamed to the request body.
+      def each
+        return enum_for(__method__) unless block_given?
+
+        if @body.is_a?(String)
+          yield @body
+        elsif @body.respond_to?(:read)
+          chunk = String.new # rubocop:disable Style/EmptyLiteral
+          yield chunk while @body.read(BUFFER_SIZE, chunk)
+        elsif @body.is_a?(Enumerable)
+          @body.each { |chunk| yield chunk }
+        end
+      end
+
+      private
+
+      def validate_body_type!
+        return if VALID_BODY_TYPES.any? { |type| @body.is_a? type }
+        raise RequestError, "body of wrong type: #{@body.class}"
+      end
+    end
+  end
+end
--- a/lib/http/request/body.rb
+++ b/lib/http/request/body.rb
@@ -13,6 +13,8 @@ module HTTP
       end
 
       # Returns size which should be used for the "Content-Length" header.
+      #
+      # @return [Integer]
       def size
         if @body.is_a?(String)
           @body.bytesize
@@ -27,6 +29,8 @@ module HTTP
       end
 
       # Yields chunks of content to be streamed to the request body.
+      #
+      # @yieldparam [String]
       def each
         return enum_for(__method__) unless block_given?
 
--- a/lib/http/request/writer.rb
+++ b/lib/http/request/writer.rb
@@ -56,6 +56,7 @@ module HTTP
         if @body.is_a?(String)
           @request_header << "#{Headers::CONTENT_LENGTH}: #{@body.bytesize}"
         elsif @body.respond_to?(:read)
+          raise(RequestError, "IO object must respond to #size") unless @body.respond_to?(:size)
           @request_header << "#{Headers::CONTENT_LENGTH}: #{@body.size}"
         elsif @body.nil?
           @request_header << "#{Headers::CONTENT_LENGTH}: 0"
--- a/lib/http/request/writer.rb
+++ b/lib/http/request/writer.rb
@@ -47,9 +47,9 @@ module HTTP
       # Adds the headers to the header array for the given request body we are working
       # with
       def add_body_type_headers
-        unless @headers[Headers::CONTENT_LENGTH] || chunked?
-          @request_header << "#{Headers::CONTENT_LENGTH}: #{@body.size}"
-        end
+        return if @headers[Headers::CONTENT_LENGTH] || chunked?
+
+        @request_header << "#{Headers::CONTENT_LENGTH}: #{@body.size}"
       end
 
       # Joins the headers specified in the request into a correctly formatted
--- a/lib/http/request/writer.rb
+++ b/lib/http/request/writer.rb
@@ -51,12 +51,16 @@ module HTTP
       # Adds the headers to the header array for the given request body we are working
       # with
       def add_body_type_headers
-        if @body.is_a?(String) && !@headers[Headers::CONTENT_LENGTH]
-          @request_header << "#{Headers::CONTENT_LENGTH}: #{@body.bytesize}"
-        elsif @body.nil? && !@headers[Headers::CONTENT_LENGTH]
-          @request_header << "#{Headers::CONTENT_LENGTH}: 0"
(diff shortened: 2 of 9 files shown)


lib/http/request/writer.rb
```diff
@@ -51,12 +51,16 @@ def connect_through_proxy
       # Adds the headers to the header array for the given request body we are working
       # with
       def add_body_type_headers
-        if @body.is_a?(String) && !@headers[Headers::CONTENT_LENGTH]
-          @request_header << "#{Headers::CONTENT_LENGTH}: #{@body.bytesize}"
-        elsif @body.nil? && !@headers[Headers::CONTENT_LENGTH]
-          @request_header << "#{Headers::CONTENT_LENGTH}: 0"
-        elsif @body.is_a?(Enumerable) && CHUNKED != @headers[Headers::TRANSFER_ENCODING]
-          raise(RequestError, "invalid transfer encoding")
+        unless @headers[Headers::CONTENT_LENGTH]
```
ixti: it's better to use guard here:

``` ruby
return if @headers[Headers::CONTENT_LENGTH]
```

lib/http/request/writer.rb
```diff
@@ -51,12 +51,16 @@ def connect_through_proxy
...
+          if @body.is_a?(String)
+            @request_header << "#{Headers::CONTENT_LENGTH}: #{@body.bytesize}"
+          elsif @body.respond_to?(:read)
+            @request_header << "#{Headers::CONTENT_LENGTH}: #{@body.size}"
```
ixti: if we are checking `respond_to? :read` we should also check here `respond_to? size`, no? So I believe the best will be:

``` ruby
if @body.respond_to?(:bytesize)
  @request_header << "#{Headers::CONTENT_LENGTH}: #{@body.bytesize}"
elsif @body.respond_to?(:size) && @body.respond_to?(:read)
elsif @body.nil?
  @request_header << "#{Headers::CONTENT_LENGTH}: 0"
elsif ...
end
```

@httprb/core please take a look if you see any gotchas here

lib/http/request/writer.rb
janko-m: I agree, updated in https://github.com/httprb/http/pull/409/commits/abe1b08637523ef4ec4b76255f4e67cf4a181fb2.

lib/http/request/writer.rb
janko-m: The reason why I wasn't checking here whether the IO object responds to `size` was because I wanted that it's *required* for the IO object to respond to `size` if `Content-Length` wasn't passed in, because for non-chunked requests `Content-Length` header is required (I would document this in the wiki).

ixti: I mean you are checking for `#read` method, but using `#size` then which might be not available:

``` ruby
reader, writer = IO.pipe

reader.respond_to? :read # => true
reader.respond_to? :size # => false
```

ixti: Oh. I got what you mean. I guess then we should raise an exception:

``` ruby
elsif @body.respond_to? :read
  raise "IO object must respond to #size" unless @body.respond_to? :size
  # ...
```

janko-m: Yes, when `IO.pipe` is passed in *and* `Content-Length` is not explicitly passed in, then the request will fail with a `NoMethodError` because `IO.pipe` doesn't respond to `#size`. This can be solved by passing `Content-Length` explicitly:

```rb
reader, writer = IO.pipe

writer.write("content")
writer.close

HTTP.headers("Content-Length" => 7).post("http://example.org", body: reader)
```

If the exception isn't raised when `Content-Length` is absent and the IO doesn't respond to `#size`, the user could accidentally send the request without a `Content-Length`, which isn't a valid HTTP request (unless it's a chunked request).

The error is currently just a `NoMethodError`, which probably doesn't communicate well to the user, would it help if I improved the error message? Or are you in disagreement on this behaviour in general?

janko-m: I saw your last comment only after I posted the previous one, great, I'll update the error message then.

janko-m: Updated in https://github.com/httprb/http/pull/409/commits/e53c71f3b5288d813b09b155cafa4998a2744f1e.

ixti:
Let's wait for input (if any) of @httprb/core and then I'll fix rubocop and merge!

tarcieri:
`IO` is `Enumerable`:

http://ruby-doc.org/core-2.0.0/IO.html

I'm curious why this wasn't working with the existing `Enumerable` support

janko-m:
@tarcieri While `IO` objects can be streamed using their `Enumerable` interface, there are several limitations of this.

First, HTTP.rb's `Enumerable` support implies `Transfer-Encoding: chunked` requests, which means that it's not possible to use it for streaming the body of a regular request.

Second, `IO#each` iterates over lines of the file, which means if we have big binary file which doesn't have a single newline, the line will equal to the whole content, so the whole file will potentially be loaded into memory during upload.

Third, by relying only on `#read` and `#size`, we allow a much broader array of inputs that aren't necessarily descendants of `IO`, such as:

* `Tempfile` (it's just a delegator to the underlying `File` object, and it doesn't implement `Enumerable`)
* `ActionDispatch::Http::UploadedFile` (a delegator to the underlying `Tempfile` object)
* Rack input for any web server (`Unicorn::StreamInput`, `Passenger::TeeInput`, ...)
* [`Shrine::UploadedFile`](https://github.com/janko-m/shrine/blob/1da66f0fc6140aa9a67efc9d063c649319cec68f/README.md#uploaded-file) from the [Shrine](https://github.com/janko-m/shrine) gem (including IO objects created by [`data_uri`](https://github.com/janko-m/shrine/blob/1da66f0fc6140aa9a67efc9d063c649319cec68f/lib/shrine/plugins/data_uri.rb#L178) and [`rack_file`](https://github.com/janko-m/shrine/blob/1da66f0fc6140aa9a67efc9d063c649319cec68f/lib/shrine/plugins/rack_file.rb#L108) plugins)
* [`Down::ChunkedIO`](https://github.com/janko-m/down/blob/75bd757b231c8d67a574e63febda19a9853b6f3c/README.md#streaming) from the [Down](https://github.com/janko-m/down) gem

ixti:
We definitely need to improve work with IO bodies... so that it will allow streaming chunked data of undetermined length (e.g. stream of JSON documents)

janko-m:
> We definitely need to improve work with IO bodies... so that it will allow streaming chunked data of undetermined length (e.g. stream of JSON documents)

Hmm, isn't that what we already have with `Enumerable` support?

ixti:
@janko-m I guess so:

``` ruby
require "http"
require "yaml"

rd, wd = IO.pipe

writer = Thread.new(wd) do |io|
  Thread.stop

  puts "READER: Start feeding data"

  %w(a b c).each do |char|
    io.write(%({"#{char}":#{rand}}\n))
    sleep 1
  end

  puts "READER: All data was feeded"

  io.close
end

reader = Thread.new(rd) do |io|
  client   = HTTP.headers({ "Transfer-Encoding" => "chunked" })

  puts "WRITER: start post request"

  response = client.post("http://httpbin.org/post", :body => io).flush

  puts "WRITER: post request finished"

  puts "== RESPONSE ==\n#{response.parse.to_yaml}"
end

sleep 0.1 until "sleep" == writer.status

writer.wakeup
reader.join
```

ixti:
Ignore me! It's httpbin is not understanding chunked requests. Tried that with netcat - all correct:

```
>>= ncat -l 8080
POST / HTTP/1.1
Transfer-Encoding: chunked
Connection: close
Host: localhost:8080
User-Agent: http.rb/2.2.2

19
{"a":0.6089059717403343}

19
{"b":0.7655504586106395}

19
{"c":0.6511286447469131}

0
```

britishtea:
I think `HTTP::Request::Writer#validate_body_type!` should also be updated. Currently it will raise when given a body that responds to `#read` and `#size` but is not `Enumerable`. Perhaps change the test such that it doesn't provide an `Enumerable`?

Looks like the `else` clause in `#send_request` can also be removed, since `#validate_body_type!` is called before.

Could bodies with a `#read` but without a `#size` (or a `#size` returning `nil`) be handled as `Enumerable` bodies are right now?

ixti:
IMHO it worth to create `Request::Body` wrapper class that will encapsulate and hide away all logic of content-length calculation, type validation and streaming.

janko-m:
> I think HTTP::Request::Writer#validate_body_type! should also be updated. Currently it will raise when given a body that responds to #read and #size but is not Enumerable. Perhaps change the test such that it doesn't provide an Enumerable?

You're right, the current code wouldn't work with an IO object which isn't `Enumerable`, I need to add a test that uses something other than a `StringIO`.

> Could bodies with a #read but without a #size (or a #size returning nil) be handled as Enumerable bodies are right now?

You mean, that if a bodies doesn't respond to `#size`, that it implies a `Transfer-Encoding: chunked` request? Yeah, we could convert such an IO into an `Enumerable`, and have the `case` clause which handles `Enumerable` handle it.

```rb
Enumerator.new do |yielder|
  loop do
    data = @io.read(Connection::BUFFER_SIZE, buffer ||= String.new)
    break if data.nil?
    yielder << data
  end
end
```

> IMHO it worth to create Request::Body wrapper class that will encapsulate and hide away all logic of content-length calculation, type validation and streaming.

Sounds good, I'm sold 💰

janko-m:
I extracted body-related behaviour into an `Request::Body` class, which deals with validating input, determining size of the body, and retrieving content to be written to the socket. It doesn't know anything about request headers, chunked encodings or sockets.

During this refactoring, I implemented @britishtea's suggestion of allowing `Transfer-Encoding: chunked` to be used with an IO object. With the extraction of `Request::Body` it was simpler to generalize how we treat the scenario when `Transfer-Encoding: chunked` is set, to support chunked requests using *any* body type, not just `Enumerable`. I like this change.

I also modified tests to verify *all* of the content that was written to the socket, not just whether it starts or ends with a string, so that I have greater confidence that I didn't break anything. I also tested streaming to the socket with both Enumerable and non-Enumerable IOs, to verify that Enumerable IOs are treated as IOs, not Enumerables. I'm not sure whether this is a breaking change, because this will change the chunked request body for Enumerable IOs, but then again it shouldn't matter in what format is the chunked request body, as long as all of the content is sent.

I see now that I've triggered quite a few Rubocop offences due to my style, I will let you correct or ignore offences which suit your preference. Investigating the JRuby failure now.

lib/http/request/body.rb
```diff
@@ -33,8 +33,10 @@ def each
         if @body.is_a?(String)
           yield @body
         elsif @body.respond_to?(:read)
-          chunk = String.new # rubocop:disable Style/EmptyLiteral
-          yield chunk while @body.read(BUFFER_SIZE, chunk)
+          loop do
+            data = @body.read(BUFFER_SIZE) or break
```
tarcieri: Personally I'd prefer something like:

```ruby
while (data = @body.read(BUFFER_SIZE)
  yield data
end
```

It better communicates intent, IMO, despite assigning in a branch.

janko-m: @tarcieri I would also prefer that, but Ruby prints a warning when you attempt to use an assignment in conditionals (Ruby thinks that you wanted to use `==`). It happens even when warning aren't enabled, just when running `ruby -e 'nil if a = 1'`.

tarcieri: Try the exact code I gave (with parens). It does not print a warning.

janko-m: That is wonderful, I'll update the code then 😃

janko-m: @tarcieri Updated.

tarcieri:
RuboCop ProTip(TM): `rubocop -a` (auto-correct)

janko-m:
I updated `http-form_data` to the prerelease version that has the streaming capabilities, and removed `to_s` so that it's used as an IO by `Request::Writer`. I also removed assigning `Content-Length`, as it will automatically be inferred from `FormData#size` by the `Request::Writer`, or omitted if the `Transfer-Encoding` is `chunked`.

I was thinking whether we need some integration tests between `HTTP` and `HTTP::FormData`, but considering that both gems are in my opinion very well-tested, and that the only thing `HTTP` needs here is `#read` and `#size` (which both `FormData::Multipart` and `FormData::Urlencoded` implement), I thought that it's not necessary.

This PR should be complete now.

ixti:
Please rebase this branch and I'll be happy to merge it down

janko-m:
@ixti Done!

ixti:
Awesome! Thanks!

PR #410, opened 2017-05-18
Let Host be automatically inferred on redirects

janko-m:
When redirecting on `http://localhost:<port>` URLs, the Host will be correctly set to `http://localhost:<port>` on the initial request, but on following the redirect the port is lost, because we're only assigning the host.

Therefore when we initialize the follow-up request object we de-assign "Host", and let HTTP::Request set it in the same way as it is set on the initial request.

Since we're stubbing redirects at the moment, I didn't know how to write a test for this.

### DIFF
--- a/lib/http/request.rb
+++ b/lib/http/request.rb
@@ -94,7 +94,10 @@ module HTTP
 
     # Returns new Request with updated uri
     def redirect(uri, verb = @verb)
-      req = self.class.new(
+      headers = self.headers.dup
+      headers.delete(Headers::HOST)
+
+      self.class.new(
         :verb    => verb,
         :uri     => @uri.join(uri),
         :headers => headers,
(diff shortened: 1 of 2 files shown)


tarcieri:
This seems to break quite a few tests

janko-m:
Oh, that's probably because I forgot that `HTTP::Response#headers` is a `HTTP::Headers` instance. I'll fix it tomorrow.

janko-m:
Fixed the implementation and added a test that's failing before this change.

ixti:
Thanks for the PR. Can you please add spec for the issue too?

janko-m:
I added it already :smiley:

ixti:
Oh. Indeed. Missed that somehow. Sorry, and thanks again!

PR #412, opened 2017-05-22
Make deflate work with any body

janko-m:
Currently auto deflating works only for string bodies, so we generalize the implementation to work with any request body (such as IO and Enumerable), by relying on the `Request::Body` abstraction. For that I had to first extract `Request::Body` outside of `Request::Writer`, so that `Request::Writer` doesn't have to know about deflating.

On "Transfer-Encoding: chunked" requests the compressed body will be directly streamed to the socket as it is deflated, while on regular requests all of the body has to be deflated first so that we can obtain the compressed body size for the "Content-Length" header before we start writing content.

~We also improve memory usage by yielding deflated chunks instead of returning the whole deflated body.~ (JRuby doesn't support it yet, see https://github.com/jruby/jruby/issues/2128)

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -78,3 +78,6 @@ Style/StringLiterals:
 
 Style/TrivialAccessors:
   Enabled: false
+
+Performance/RedundantBlockCall:
+  Enabled: false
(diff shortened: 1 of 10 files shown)


ixti:
Mered! Thanks!

PR #414, opened 2017-05-24
Return response body chunks in specified encoding

### LINKED ISSUE #413: Differences in response body encoding on JRuby
When I run the following:

```rb
require "http"

response = HTTP.get("http://httpbin.org/image/png")
response.body.each do |chunk|
  puts chunk.encoding
end
```

MRI outputs `ASCII-8BIT`, whereas JRuby outputs `UTF-8` encodings. The same thing happens when I swap the endpoint with `http://httpbin.org/encoding/utf8`.

I'm note sure what should be the correct general behaviour, whether chunks should always be in binary encoding, or only when charset isn't specified. But it seems that me that in the latter case the encoding should always be binary, both in MRI and JRuby.

I'm not sure yet whether this cross-Ruby inconsistency comes from HTTP.rb or http_parser.rb, but just wanted to report it here. I think a good solution would be to call `force_encoding` on the result of `Response::Body#readpartial` (just like we're doing in `Response::Body#to_s`).

janko-m:
MRI will return content read from the socket in the ASCII-8BIT (binary) encoding, whereas JRuby will return it in the UTF-8 encoding. In whichever encoding the body is retrieved, we want to force its encoding to the one specified (charset response header if present, otherwise binary). This is already the behaviour in `Response::Body#to_s`, we just extend it to `Response::Body#readpartial` as well.

Fixes #413.

### DIFF
--- a/lib/http/response/body.rb
+++ b/lib/http/response/body.rb
@@ -20,14 +20,15 @@ module HTTP
         @connection = stream.is_a?(Inflater) ? stream.connection : stream
         @streaming  = nil
         @contents   = nil
-        @encoding   = encoding
+        @encoding   = find_encoding(encoding)
         @length     = length || Float::INFINITY
       end
 
       # (see HTTP::Client#readpartial)
       def readpartial(*args)
         stream!
-        @stream.readpartial(*args)
+        chunk = @stream.readpartial(*args)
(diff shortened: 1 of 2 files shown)


ixti:
Thanks!

PR #418, opened 2017-07-01
Further upload improvements

janko-m:
This pull request adds two more upload improvements. First commit changes `HTTP::Request::Body` to use `IO#readpartial` if available, and second commit adds a buffer string for memory-efficient uploads; both are explained in detail in their commit descriptions.

I was hoping that, once this is merged (or rejected), we could release a new version of HTTP.rb and form_data.rb with the streaming upload feature.

### DIFF
--- a/lib/http/request/body.rb
+++ b/lib/http/request/body.rb
@@ -4,7 +4,7 @@ module HTTP
   class Request
     class Body
       # Maximum chunk size used for reading content of an IO
-      BUFFER_SIZE = Connection::BUFFER_SIZE
+      CHUNK_SIZE = Connection::BUFFER_SIZE
 
       def initialize(body)
         @body = body
@@ -37,8 +37,14 @@ module HTTP
         if @body.is_a?(String)
           yield @body
         elsif @body.respond_to?(:read)
-          while (data = @body.read(BUFFER_SIZE))
-            yield data
+          read_method = @body.respond_to?(:readpartial) ? :readpartial : :read
+
+          begin
+            while (data = @body.send(read_method, CHUNK_SIZE))
+              yield data
+            end
+          rescue EOFError
+            # done reading
           end
         elsif @body.is_a?(Enumerable)
           @body.each { |chunk| yield chunk }
--- a/lib/http/request/body.rb
+++ b/lib/http/request/body.rb
@@ -29,8 +29,6 @@ module HTTP
       #
       # @yieldparam [String]
       def each(&block)
-        return enum_for(__method__) unless block_given?
-
         if @body.is_a?(String)
           yield @body
         elsif @body.respond_to?(:read)
--- a/lib/http/request/body.rb
+++ b/lib/http/request/body.rb
@@ -32,7 +32,7 @@ module HTTP
         if @body.is_a?(String)
           yield @body
         elsif @body.respond_to?(:read)
-          IO.copy_stream(@body, BlockIO.new(block))
+          IO.copy_stream(@body, ProcIO.new(block))
         elsif @body.is_a?(Enumerable)
           @body.each(&block)
         end
@@ -49,7 +49,10 @@ module HTTP
         raise RequestError, "body of wrong type: #{@body.class}"
       end
 
-      class BlockIO
+      # This class provides a "writable IO" wrapper around a proc object, with
+      # #write simply calling the proc, which we can pass in as the
+      # "destination IO" in IO.copy_stream.
+      class ProcIO
         def initialize(block)
           @block = block
         end
--- a/lib/http/request/body.rb
+++ b/lib/http/request/body.rb
@@ -29,6 +29,8 @@ module HTTP
       #
       # @yieldparam [String]
       def each(&block)
+        return enum_for(__method__) unless block_given?
+
         if @body.is_a?(String)
           yield @body
         elsif @body.respond_to?(:read)
--- a/lib/http/request/body.rb
+++ b/lib/http/request/body.rb
@@ -3,9 +3,6 @@
 module HTTP
   class Request
     class Body
-      # Maximum chunk size used for reading content of an IO
-      CHUNK_SIZE = Connection::BUFFER_SIZE
-
       def initialize(body)
         @body = body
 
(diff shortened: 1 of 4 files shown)


mikegee:
Does this need an addition to the test suite to exercise both `#read` and `#readpartial`?

janko-m:
@mikegee The example that uses `StringIO` as a "Enumerable IO" already exercises `#readpartial`, and the "non-Enumerable" `FakeIO` example exercises `#read`.

mikegee:
Thanks for the detailed commit messages. I was about to ask why `#readpartial` was preferable.

janko-m:
@mikegee I made what I think is an improvement in the last commit. Instead of writing out the `#read`/`#readpartial` logic and using a buffer string, we can use `IO.copy_stream` which will do all that for us (see the [Rubinius implementation](https://github.com/rubinius/rubinius/blob/1a98abf9e1913e0cbdd6d2d65f3ae74371c6f99f/core/io.rb#L1099-L1108)), with the help of a `BlockIO` object which acts as a "destination IO" and just calls the block with the data that `IO.copy_stream` "writes".

Here we use a trick of creating a Proc objects from an implicit block by calling `Proc.new` without any arguments. So, this:

```rb
def foo
  block = Proc.new
end
```

is equivalent to

```rb
def foo(&block)
end
```

This way we can still use `yield` in other branches of the conditional.

ixti:
Wow! TIL about `def foo; Proc.new; end`

lib/http/request/body.rb
```diff
@@ -37,8 +37,14 @@ def each
         if @body.is_a?(String)
           yield @body
         elsif @body.respond_to?(:read)
-          while (data = @body.read(BUFFER_SIZE))
-            yield data
+          read_method = @body.respond_to?(:readpartial) ? :readpartial : :read
+
+          begin
+            while (data = @body.send(read_method, CHUNK_SIZE))
```
ixti: probably it's better to call `public_send` in here.

lib/http/request/body.rb
```diff
@@ -32,14 +29,10 @@ def size
       #
       # @yieldparam [String]
       def each
-        return enum_for(__method__) unless block_given?
```
ixti: Why removing this?

lib/http/request/body.rb
```diff
@@ -32,14 +29,10 @@ def size
...
-
         if @body.is_a?(String)
           yield @body
         elsif @body.respond_to?(:read)
-          while (data = @body.read(BUFFER_SIZE))
-            yield data
-          end
+          IO.copy_stream(@body, BlockIO.new(Proc.new))
```
ixti: IMHO it worth to simply change `each` signature to `def each(&block)` and pass the block here, and in next clause (`@body.each(&block)`)

lib/http/request/writer.rb
```diff
@@ -61,25 +61,20 @@ def join_headers
...
         # possible in order to play nicely with Nagle's algorithm. Making
         # two writes in a row triggers a pathological case where Nagle is
         # expecting a third write that never happens.
-        begin
-          first_chunk = encode_chunk(chunks.next)
-          write(headers << first_chunk)
-        rescue StopIteration
-          write(headers)
-        end
+        data = headers
```
ixti: we can get rid of previous line in this case in fact:

``` ruby
data = join_headers
```

lib/http/request/body.rb
janko-m: I removed this because we don't need it anymore, and because how JRuby implements Enumerators (https://github.com/jruby/jruby/issues/4583) it's not usable on JRuby when fetching next element mutates the previous one (reading file with a buffer string). So I thought it's better to remove it if it's not being used.

janko-m:
@ixti Updated. I brought back returning Enumerator in `HTTP::Request::Body#each` because it makes tests simpler. Also, because of https://github.com/jruby/jruby/issues/4701 I changed the tests for IO objects to just verify that yielded chunks sum up to the entire content.

lib/http/request/body.rb
ixti: Hm. If this works with issues on  JRuby, I think we should remove it then indeed :D

janko-m: Agreed, updated in a7dee75.

lib/http/request/body.rb
```diff
@@ -28,11 +28,11 @@ def size
...
       # @yieldparam [String]
-      def each
+      def each(&block)
         if @body.is_a?(String)
           yield @body
         elsif @body.respond_to?(:read)
-          IO.copy_stream(@body, BlockIO.new(Proc.new))
+          IO.copy_stream(@body, BlockIO.new(block))
         elsif @body.is_a?(Enumerable)
           @body.each { |chunk| yield chunk }
```
britishtea: The captured block can safely be used for the `is_a?(Enumerable)` case, too :)

lib/http/request/body.rb
ixti: Can you please also add a comment so that in future we won't add regression:

``` ruby
# @note This method does not returns Enumerator due to known issue on JRuby:
#   https://github.com/jruby/jruby/issues/4583
```

Asking that, beacause I know at least one developer (me) who tries to sneak `to_enum __method__ unless block_given?` in every `each` method even if there's no real need in that :D

lib/http/request/body.rb
```diff
@@ -55,6 +48,17 @@ def validate_body_type!
 
         raise RequestError, "body of wrong type: #{@body.class}"
       end
+
+      class BlockIO
```
ixti: Probably worth name it `ProcIO`. It took a while for me to understand that *block* here refers to ruby block, rather than to *blocking*.

janko-m:
@britishtea @ixti Updated.

janko-m:
@ixti @britishtea Do you think this one is ready for merging?

ixti:
Sorry I was on a vacation and completely off the grid. Will handle this ASAP.

janko-m:
@ixti Quick bump 😃

ixti:
Sorry took me so long. Merged. Thank you!

janko-m:
@ixti No problem at all!

Sorry that I've been nagging about it, I just thought about writing a section in  [Shrine](https://github.com/janko-m/shrine) documentation about how to upload files with various HTTP libraries, and I wanted to wait until HTTP.rb had streaming IO uploads, so that I don't have to change the documentation later.

Now that we merged these final memory optimizations, and made deflating work with IO/Enumerable uploads (https://github.com/httprb/http/pull/412), do you think we can release a new HTTP.rb version with the new streaming IO uploads feature?

ixti:
@janko-m I want to work-on/merge all outstanding PRs that can be merged in relatively fast, and then we'll be good to release 3.0.0.pre1 I just want to merge max breaking changes as we ae doing that anyway ;))

janko-m:
@ixti Sounds great, thank you!

PR #424, opened 2017-08-03
Allow http_parser to decide if there is still body to read

Bonias:
It mostly fixes problem when connection which uses "Transfer-Encoding: chunked" header is closed to early, before the terminating chunk is received. In such case "http" returned partial body instead of raising error.

---

Here is how the problem can be reproduced:

Run this app (`rackup  -p 3000 -E none -s puma`):
```
$ cat config.ru 
class App
  def call(env)
    [200, {}, self]
  end

  def each
    yield 'one'
    sleep 1
    yield 'two'
    sleep 2
    yield 'three'
    sleep 3
    yield 'four'
    sleep 4
    yield 'five'
    sleep 5
    yield 'six'
  end
end

use Rack::Chunked
run App.new
```

Use irb console to make http request to the app and then kill the app (in the middle of the request):

```
2.2.3 :001 > require 'http'
 => true 
2.2.3 :002 > HTTP.get('http://localhost:3000/').body.to_s
 => "onetwo" 
```

----

All tests are passing so my fix seems to be fine, but if you think otherwise let me know and I'll try change it.

### DIFF
--- a/lib/http/connection.rb
+++ b/lib/http/connection.rb
@@ -211,6 +211,7 @@ module HTTP
 
       value = @socket.readpartial(size)
       if value == :eof
+        @parser << ""
         :eof
       elsif value
         @parser << value
--- a/lib/http/response.rb
+++ b/lib/http/response.rb
@@ -50,7 +50,7 @@ module HTTP
         encoding   = opts[:encoding] || charset || Encoding::BINARY
         stream     = body_stream_for(connection, opts)
 
-        @body = Response::Body.new(stream, :encoding => encoding, :length => content_length)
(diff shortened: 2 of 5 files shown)


woutdegeyter:
@ixti I am wondering if you would happen to have any news on this PR? 
We have encountered multiple issues due to this already

ixti:
Thanks! Merged as 72723df4a7574448ef4c3a90666e954704fd52b1

Bonias:
@ixti Sorry to bother you, is there any chance for release a new version in the near future?

ixti:
@Bonias thanks for reminder :D I will handle that during this week (or over this weekend).

Issue #422, opened 2017-07-23
Upgrade header which advertises protocol support being treated as an actual upgrading


unarist:
from [RFC7230 sec. 6.7](https://tools.ietf.org/html/rfc7230#section-6.7):

> A server MAY send an Upgrade header field in any other response to advertise that it implements support for upgrading to the listed protocols, in order of descending preference, when appropriate for a future request.

However, current http.rb treats it as an actual upgrading, and doesn't process response body.

c.f. https://github.com/nodejs/http-parser/issues/363

Gargron:
:+1:

tarcieri:
Picking up the discussion from #508, hey @ixti, is #433 still the path forward on that?

Perhaps we should open an upstream issue about this on http_parser.rb (unless there already is one?)

janko-m:
@tarcieri Considering that http_parser.rb doesn't work on JRuby 9.2.0.0, I think the path forward really is switching to `http-parser` gem. [http-protocol](https://github.com/socketry/http-protocol) also looks good, if we want pure Ruby.

ixti:
Yeah. my plan was to switch to http-parser but with shim for it (I don't like it's API).

tarcieri:
@ixti that's basically what you were working on in #433, right?

ixti:
@tarcieri Yup!

ixti:
I have cut 5.0.0 pre release version of gem with switch to updated http parser. This should be fixed now.

PR #433, opened 2017-10-01
Fix HTTP parsing

ixti:
This is a discussion starter for resolving #422

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -5,6 +5,8 @@ ruby RUBY_VERSION
 
 gem "rake"
 
+gem "http_parser.rb", :git => "https://github.com/ixti/http_parser.rb.git"
+
 group :development do
   gem "guard-rspec", :require => false
   gem "nokogiri",    :require => false
--- a/spec/regression_specs.rb
+++ b/spec/regression_specs.rb
@@ -14,4 +14,11 @@ RSpec.describe "Regression testing" do
       expect { HTTP.get(google_uri).to_s }.not_to raise_error
     end
   end
+
+  describe "#422" do
+    it "reads body when 200 OK response contains Upgrade header" do
+      res = HTTP.get("https://httpbin.org/response-headers?Upgrade=h2,h2c")
+      expect(res.parse(:json)).to include("Upgrade" => "h2,h2c")
+    end
+  end
 end


ixti:
@httprb/core @janko-m @Bonias What do you think about us changing underlying http parser. 

The problem with existing one is that it's locked to the old version of nodejs/http-parser. That can be fixed by updating underlying C version, but Java version of it is not maintained at all. So I think that the best option for us is to use FFI.

I'm not big fan of http-parser ffi wrapper but mostly due to it's API (it seems a bit ugly to me) and after all we end up by wrapper over wrapper. So probably if there are no strong objections against we can have our own http parser ffi version :D

ixti:
Invited [mastodon](https://github.com/tootsuite/mastodon) devs to participate in discussion as well.

//cc @Gargron

Gargron:
cc @unarist @akihikodaki

ixti:
So, the principal questions I propose to focus on are:

- if there are any objections against [FFI](https://github.com/ffi/ffi)
- if no objects, should we use existing gems for that or implement ourselves

tarcieri:
I have no objections to FFI. It should provide a more consistent experience
between MRI and JRuby.

That said, I would recommend packaging the parser in the gem as opposed to
requiring an external dependency. rbnacl-libsodium does this, and builds on
both MRI and JRuby:

https://github.com/cryptosphere/rbnacl-libsodium
-- 
Tony Arcieri

akihikodaki:
I have no objections to FFI.
I suggest to use the existing gem. Though its API may look odd, it is just conformant with the underlying the library and legitimate as a language binding.
I do not think _wrapper over wrapper_ is really bad in this case. You just have different wrappers playing different roles: a wrapper providing APIs convenient for http.rb and a language binding.

tarcieri:
Curious what @tmm1 thinks about converting the upstream http_parser.rb gem to use FFI and dropping the extant JRuby extension

zanker:
It'll still work with JRuby via FFI though in that case, right?

tarcieri:
@zanker yes, although it will build the native code library at the time you install the gem

zanker:
That sounds fine to me. Don’t think compiling is the end of the world unless we’re trying to target something that’s 100% java which seems unnecessary.

ixti:
@akihikodaki Well, I guess I have expressed myself a bit wrong. I'm totally fine with using existing gems even if it leads to wrapper over wrapper. I have only 2 problems with that particular gem:

- Naming. I find naming and some ruby API decisions pretty terrible. Not talking about things that are mapped to underlying library (Just general namespace and HttpParser::Parser.instance call makes me a bit sceptical)
- Consumption of body. I did not found streaming API there it seems like it fires body in one chunk after it was consumed (in case of non-chunked responses).

ixti:
Okay. Excuse me. I was wrong. `on_body` actually fired everytime there's some data in body arrived.

akihikodaki:
Oh, I see. `HttpParser::Parser`, `HttpParser::Instance`, `HttpParser::Parser.new_instance`… Indeed that sounds bad.

janko-m:
After finding out that http_parser.rb doesn't work on JRuby 9.2.0.0 anymore (https://github.com/httprb/http/issues/475), I realized why this PR is important 😃 

It's great that there already exists the [http-parser](https://github.com/cotag/http-parser) gem that builds upon Node.js' [http-parser](https://github.com/nodejs/http-parser) (just like the current [http_parser.rb](https://github.com/tmm1/http_parser.rb)), but uses FFI instead of native extensions. I'm totally for using the http-parser gem, if any functionality is missing we can send PRs there.

Thanks @ixti for working on this!

PR #434, opened 2017-10-13
Remove http-form_data pre-release from gemspec

janko-m:
The gemspec included the pre-release of `http-form_data` at the moment the major version wasn't out yet, and we forgot to remove it once it was released. This PR takes care of that.


ixti:
Nice!

PR #435, opened 2017-10-14
Remove .ruby-version file

janko-m:
The HTTP.rb codebase runs just as well on other Ruby versions, it's not necessary to use 2.4.0 specifically. This is inconvenient when using rbenv, because rbenv will automatically try to switch to the version specified in .ruby-version, and complain if you don't have it installed.


ixti:
Can you please add it to `.gitignore` then?

janko-m:
@ixti I can add it to my global `.gitignore`, but it won't change anything because it's still checked into the version control here. If I delete it locally the git status will still be left "dirty", because its deletion is tracked.

janko-m:
Oh, I see that you might have meant this `.gitignore`, I'll add it.

ixti:
Thanks! RuboCop also uses .ruby-version if no trarget specified in config. :D
Please fix .rubocop.yml as well by adding `TargetRubyVersion: 2.4` under `AllCops` section.

janko-m:
@ixti It seems that GitHub isn't updating the PR with the new changes I'm pushing 🤔

ixti:
Thanks! Merged in!

PR #436, opened 2017-10-14
Raise ConnectionError when writing to socket fails

### LINKED ISSUE #430: Handling EPIPE during write
While trying to send a request, something happens (still investigating the source of the problem, it might be the server dropping the connection), resulting in a `EPIPE` error.
Would I be correct in assuming that this best be caught in the request writer? If so, I can prepare a PR to that effect.
Of course I'm open to other suggestions :)

```
(main)> err.backtrace
=> [
    [ 0] "<internal:prelude>:134:in `__write_nonblock'",
    [ 1] "<internal:prelude>:134:in `write_nonblock'",
    [ 2] "/app/vendor/bundle/ruby/2.3.0/gems/http-2.2.2/lib/http/timeout/per_operation.rb:77:in `block in write'",
    [ 3] "/app/vendor/bundle/ruby/2.3.0/gems/http-2.2.2/lib/http/timeout/per_operation.rb:76:in `loop'",
    [ 4] "/app/vendor/bundle/ruby/2.3.0/gems/http-2.2.2/lib/http/timeout/per_operation.rb:76:in `write'",
    [ 5] "/app/vendor/bundle/ruby/2.3.0/gems/http-2.2.2/lib/http/request/writer.rb:99:in `write'",
    [ 6] "/app/vendor/bundle/ruby/2.3.0/gems/http-2.2.2/lib/http/request/writer.rb:87:in `block in send_request'",
```

janko-m:
Writing to the TCP socket can also fail with various exceptions, just like when opening or reading from it. For example, it happened to me that an `Errno::EPIPE` exceptions was raised during file upload, presumably because the server closed the connection early.

In order to be consistent with re-raising low-level exceptions as `HTTP::ConnectionError` when opening or reading from the TCP socket, we also add the same behaviour when writing to the socket.

Closes #430

### DIFF
--- a/lib/http/request/writer.rb
+++ b/lib/http/request/writer.rb
@@ -100,6 +100,8 @@ module HTTP
           break unless data.bytesize > length
           data = data.byteslice(length..-1)
         end
+      rescue IOError, SocketError, SystemCallError => ex
+        raise ConnectionError, "error writing to socket: #{ex}", ex.backtrace
       end
     end
   end
--- a/spec/lib/http/request/writer_spec.rb
+++ b/spec/lib/http/request/writer_spec.rb
@@ -74,5 +74,15 @@ RSpec.describe HTTP::Request::Writer do
         ].join
       end
     end
+
+    context "when writing to socket raises an exception" do
+      before do
+        expect(io).to receive(:write).and_raise(Errno::EPIPE)
+      end
+
+      it "raises a ConnectionError" do
+        expect { writer.stream }.to raise_error(HTTP::ConnectionError)
+      end
+    end
   end
 end


ixti:
Thank you!

tarcieri:
If only I ever finished up the [socketry branch](https://github.com/httprb/http/tree/socketry)...

https://github.com/socketry/socketry/blob/master/lib/socketry/tcp/socket.rb#L301

janko-m:
@tarcieri Man, Socketry looks awesome. Yeah, I've noticed that open pull request for replacing the direct socket communication with Socketry.

Btw, shouldn't in this case Socketry re-raise other `SystemCallException` as `Socketry::Error` exceptions (in addition to `IOError`)?

tarcieri:
@janko-m good call, want to submit a PR to Socketry? 😄

Issue #437, opened 2017-10-17
Logging request payload\data\body with httplog is broken in 3.0.0


igorigorigorigor:
Httplog(https://github.com/trusche/httplog) logs kind of memory pointer instead of the payload\data\body itself.
`D, [2017-10-17T18:42:06.013390 #53965] DEBUG -- : [httplog] Sending: POST https://address.domain/api_method`
`D, [2017-10-17T18:42:06.013533 #53965] DEBUG -- : [httplog] Data: #<HTTP::Request::Body:0x007f93735632c8>`

Here is my code for this
`payload = { :param_1 => param_1.to_s, :param_2 => param_2.to_s, :param_3 => param_3.to_s}.to_json`
`HTTP.headers("id" => ID, :"User-Agent" => user_agent).post(url), :body => payload)`

janko-m:
The change introduced in https://github.com/httprb/http/pull/412 is responsible for this behaviour. Before `HTTP::Request#body` would return the object that was assigned as `:body`, but now it returns an instance of `HTTP::Request::Body`. The fix is very easy, I'll send a PR to httplog.

janko-m:
@igorigorigorigor This PR should bring necessary changes to HTTP.rb for me to then be able to send a PR to httplog – https://github.com/httprb/http/issues/437.

janko-m:
@tycooon I was actually thinking of that the httplog gem would just do `source.inspect`, so the logs would look like this:

```rb
HTTP.post("http://example.com", body: "body")
# [httplog] Data: "body"

HTTP.post("http://example.com", json: { key: "value" })
# [httplog] Data: "{\"key\":\"value\"}"

HTTP.post("https://example.com", body: Enumerator.new { |y| ... })
# [httplog] Data: #<Enumerator: ...>

HTTP.post("https://example.com", body: File.open("/path/to/file"))
# [httplog] Data: #<File:/path/to/file>

HTTP.post("https://example.com", form: {file: HTTP::FormData::File.new("/path/to/file")})
[...]

PR #438, opened 2017-10-23
Add Request::Body#source

janko-m:
This makes it possible to access the original body object, which [httplog](https://github.com/trusche/httplog) needs to log request bodies.

Related to #437.

### DIFF
--- a/lib/http/request.rb
+++ b/lib/http/request.rb
@@ -86,7 +86,8 @@ module HTTP
       raise(UnsupportedSchemeError, "unknown scheme: #{scheme}") unless SCHEMES.include?(@scheme)
 
       @proxy   = opts[:proxy] || {}
-      @body    = request_body(opts[:body], opts)
+      @body    = Request::Body.new(opts[:body])
+      @deflate = opts[:auto_deflate]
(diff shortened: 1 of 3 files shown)


janko-m:
This is actually not necessary, and isn't applicable (e.g. when using deflating `Request#body` will not be a `Request::Body` instance). I think a different approach would be better.

janko-m:
This is now ready for review.

I modified `HTTP::Request` to create a deflated body only when it's about to stream. The reason for that change is that httplog is currently retrieving the `HTTP::Request#body`, and I wanted it to be able to call `#source` to retrieve the original body object, but that wouldn't currently work with `:auto_deflate`, because `HTTP::Request#body` wouldn't return a `HTTP::Request::Body` object but rather the `DeflatedBody` wrapper.

I think this change is beneficial in general in terms of consistency, because with this PR `HTTP::Request#body` will always return a `HTTP::Request::Body` object, even when `:auto_deflate` is turned on.

tarcieri:
Looks good to me. Any thoughts @ixti?

ixti:
LGTM as well! Thanks!

PR #446, opened 2017-12-31
Simplify setting a timeout

### LINKED ISSUE #444: More convenient timeout
As much as I absolutely appreciate to have the flexibility of connect/write/read timeouts separately, sometimes it makes sense to have a simpler form:

```ruby
HTTP.timeout(5).get
```

as a drop-in replacement for:

```ruby
Timeout.timeout(5) { YourHttpTool.get }
```

which maintains parity with the original semantics, thus enable to migrate from other tools.

Is this something we're interested?

mikegee:
- `.timeout(<Numeric>)` sets a global timeout value.
- `.timeout(connect: x, write: y, read: z)` sets per operation timeouts.

The previous style: `.timeout(:global|:per_operation, hash_of_options)` doesn't work anymore.

fixes #444

### DIFF
--- a/lib/http/chainable.rb
+++ b/lib/http/chainable.rb
@@ -77,29 +77,24 @@ module HTTP
     end
 
     # @overload timeout(options = {})
-    #   Syntax sugar for `timeout(:per_operation, options)`
-    # @overload timeout(klass, options = {})
-    #   Adds a timeout to the request.
-    #   @param [#to_sym] klass
-    #     either :null, :global, or :per_operation
+    #   Adds per operation timeouts to the request
     #   @param [Hash] options
     #   @option options [Float] :read Read timeout
     #   @option options [Float] :write Write timeout
     #   @option options [Float] :connect Connect timeout
-    def timeout(klass, options = {}) # rubocop:disable Style/OptionHash
-      if klass.is_a? Hash
-        options = klass
-        klass   = :per_operation
-      end
+    # @overload timeout(global_timeout)
+    #   Adds a global timeout to the full request
+    #   @param [Numeric] global_timeout
+    def timeout(options)
+      klass, options = case options
+                       when Numeric then [HTTP::Timeout::Global, {:global => options}]
+                       when Hash    then [HTTP::Timeout::PerOperation, options]
+                       when :null   then [HTTP::Timeout::Null, {}]
+                       else raise ArgumentError, "Use `.timeout(global_timeout_in_seconds)` or `.timeout(connect: x, write: y, read: z)`."
 
-      klass = case klass.to_sym
-              when :null          then HTTP::Timeout::Null
-              when :global        then HTTP::Timeout::Global
-              when :per_operation then HTTP::Timeout::PerOperation
-              else raise ArgumentError, "Unsupported Timeout class: #{klass}"
-              end
+                       end
 
-      %i[read write connect].each do |k|
+      %i[global read write connect].each do |k|
         next unless options.key? k
         options["#{k}_timeout".to_sym] = options.delete k
       end
--- a/lib/http/timeout/global.rb
+++ b/lib/http/timeout/global.rb
@@ -3,27 +3,25 @@
 require "timeout"
 require "io/wait"
 
-require "http/timeout/per_operation"
+require "http/timeout/null"
 
 module HTTP
   module Timeout
-    class Global < PerOperation
-      attr_reader :time_left, :total_timeout
-
+    class Global < Null
       def initialize(*args)
         super
-        reset_counter
+
+        @timeout = @time_left = options.fetch(:global_timeout)
       end
 
       # To future me: Don't remove this again, past you was smarter.
       def reset_counter
-        @time_left = connect_timeout + read_timeout + write_timeout
-        @total_timeout = time_left
+        @time_left = @timeout
       end
 
       def connect(socket_class, host, port, nodelay = false)
         reset_timer
-        ::Timeout.timeout(time_left, TimeoutError) do
+        ::Timeout.timeout(@time_left, TimeoutError) do
           @socket = socket_class.open(host, port)
           @socket.setsockopt(Socket::IPPROTO_TCP, Socket::TCP_NODELAY, 1) if nodelay
         end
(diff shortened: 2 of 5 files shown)


tarcieri:
RuboCop errors:

```
Offenses:
lib/http/chainable.rb:91:53: C: Layout/SpaceInsideHashLiteralBraces: Space inside { detected.
                       when Integer then [:global, { read: klass }]
                                                    ^
lib/http/chainable.rb:91:54: C: Style/HashSyntax: Use hash rockets syntax.
                       when Integer then [:global, { read: klass }]
                                                     ^^^^^
lib/http/chainable.rb:91:65: C: Layout/SpaceInsideHashLiteralBraces: Space inside } detected.
                       when Integer then [:global, { read: klass }]
                                                                ^
71 files inspected, 3 offenses detected
RuboCop failed!
```

janko-m:
When I read #444, I imagined that specifying an integer would mean having a timeout around the whole request, in a way that the request can time out during *any* operation, not just read. With the currently proposed implementation the request can take longer than `n` seconds since connect and write operations are not limited. If I specify 5 seconds, connecting could take 2 seconds, writing 3, and reading 4, which means the request lasted 9 seconds in total.

mikegee:
> a timeout around the whole request

Indeed, I forgot to divide by 3. Global timeout doesn't work in the least surprising way, IMO.

mikegee:
hrm. maybe setting only one to the expected global value will work as expected?

https://github.com/httprb/http/blob/70658da2af3e019b48fd4faf6551fb49fda607ad/lib/http/timeout/global.rb#L20

zanker:
Sorry about this, it's mostly a wart from how I implemented it a few years ago.

You can add a new field like `total_time` and use that if it's set, otherwise fallback to `connect + read + write`. We should deprecate the old one of adding things together, since it's not very clear.

tarcieri:
I know we're at 3.0.0 now, but it might be nice to go to 4.0.0 and redesign the timeout API a bit. That would definitely simplify moving things over to Socketry underneath

mikegee:
I created a `global_timeout` option like the `PerOperation` values. It still sums the per operation values if `global_timeout` isn't provided (explicitly or via `.timeout(Numeric)`).

I don't like that the deprecation warning requires using the connection, `HTTP.timeout(:global, read: 3)` is insufficient. I don't know if anything can be done about it or if it even matters.

I would have liked to make `Global` not inherit from `PerOperation` and simplify the code some, but backwards compatibility ...

tarcieri:
@mikegee I'd suggest rewriting the patch without regard to backwards compatibility

mikegee:
@tarcieri alrighty. I'm working on it.

mikegee:
I updated the title and description of this change. I would appreciate some review now.

lib/http/timeout/global.rb
```diff
@@ -3,22 +3,24 @@
...
-        reset_counter
+
+        @timeout = @time_left = options.fetch(:global_timeout, DEFAULT_TIMEOUT)
       end
 
       # To future me: Don't remove this again, past you was smarter.
       def reset_counter
-        @time_left = connect_timeout + read_timeout + write_timeout
-        @total_timeout = time_left
+        @time_left = timeout
```
ixti: Personal flavor, but can you please use instance variable instead of attr reader here?

lib/http/timeout/global.rb
```diff
@@ -124,7 +126,7 @@ def reset_timer
       def log_time
         @time_left -= (Time.now - @started)
         if time_left <= 0
```
ixti: Same here. I know that it was before, but please update this one too to use instance variable instead.

lib/http/chainable.rb
```diff
@@ -76,30 +76,28 @@ def request(verb, uri, options = {}) # rubocop:disable Style/OptionHash
...
-              when :per_operation then HTTP::Timeout::PerOperation
-              else raise ArgumentError, "Unsupported Timeout class: #{klass}"
-              end
-
-      %i[read write connect].each do |k|
+    # @overload timeout(global_timeout)
+    #   Adds a global timeout to the full request
+    #   @param [Numeric] global_timeout
+    def timeout(options, extra_arg = nil)
+      raise ArgumentError, TIMEOUT_USAGE if extra_arg
```
ixti: This is gonna be a breaking change, so I think it's fine to just change method signature to accept single argument only.

ixti:
Couple of changes - and I will be glad to merge it.

tarcieri:
Looks good to me too... aside from @ixti's nits

lib/http/chainable.rb
mikegee: I debated this, but thought a bit of effort to ease the transition was worth it. I'll remove it. No worries.

tarcieri: Per #447 we might end up redoing the whole thing with keyword arguments anyway, but don't worry about that for this PR

lib/http/timeout/global.rb
mikegee: Yeah, sure. I'll change `PerOperation` to match your preferred style, too.

BTW, I avoid instance variables in my code since typos can lead to subtle bugs: `@mispelled #=> nil`, whereas misspelled method calls raise a `NameError`.

mikegee:
I don't understand the build failure. It is in unrelated code, I can't duplicate it locally in 2.4.1, and when I pushed a commit focused on that spec, it passed. Is there a way to tell Travis to retry?

ixti:
Yeah, I have restarted.

spec/lib/http_spec.rb
```diff
@@ -249,15 +249,15 @@
...
     end
 
     it "fails when :pass is not given" do
-      expect { HTTP.basic_auth :user => "[USER]" }.to raise_error
+      expect { HTTP.basic_auth :user => "[USER]" }.to raise_error(KeyError)
     end
 
     it "fails when :user is not given" do
-      expect { HTTP.basic_auth :pass => "[PASS]" }.to raise_error
+      expect { HTTP.basic_auth :pass => "[PASS]" }.to raise_error(KeyError)
```
ixti: Huge thanks for fixing those!

ixti:
Thank you for your contribution!

tarcieri:
Thanks!

PR #450, opened 2018-01-16
Fix query string building for string with newlines

### LINKED ISSUE #449: Newline "\n" is encoded as "%0D%0A" in query string values
Hello 🙂

I use version 2.2.2 but the code seems to be the same in 3.0.0 as well.

I noticed that newlines `\n` are encoded as `\r\n` in query string:

```
 ~    ruby -v
ruby 2.5.0p0 (2017-12-25 revision 61468) [x86_64-darwin17]
 ~    irb
irb(main):001:0> require 'http'
=> true
irb(main):002:0> HTTP::VERSION
=> "2.2.2"
irb(main):003:0> require 'httplog'
=> true
irb(main):004:0> HTTP.get('http://example.com', params: {foo: "bar\nbaz"})
D, [2018-01-15T19:25:37.641493 #66440] DEBUG -- : [httplog] Sending: GET http://example.com/?foo=bar%0D%0Abaz
```

This is because the params are put through `Addressable::URI.form_encode`, which replaces everything with `\r\n`:

https://github.com/httprb/http/blob/c69844911b74472bd40a9979fa452bff6c179045/lib/http/client.rb#L126

I think that this behavior is incorrect. It is also unexpected. Query string values are not form values.

For example, ActiveSupport simply encodes `\n`:

```
irb(main):001:0> require 'active_support/core_ext/object/to_query'
=> true
irb(main):002:0> {foo: "bar\nbaz"}.to_query
=> "foo=bar%0Abaz"
```

mikegee:
Expose `HTTP::URI#query_values` & `HTTP::URI#query_values=` which delegate to the underlying `Addressable::URI` instead of using Addressable's `form_encode` method.

fixes #449

### DIFF
--- a/lib/http/client.rb
+++ b/lib/http/client.rb
@@ -123,7 +123,7 @@ module HTTP
       uri = HTTP::URI.parse uri
 
       if opts.params && !opts.params.empty?
-        uri.query = [uri.query, HTTP::URI.form_encode(opts.params)].compact.join("&")
+        uri.query_values = uri.query_values(Array).to_a.concat(opts.params.to_a)
       end
 
       # Some proxies (seen on WEBRick) fail if URL has
--- a/lib/http/uri.rb
+++ b/lib/http/uri.rb
@@ -15,6 +15,7 @@ module HTTP
     def_delegators :@uri, :normalized_port, :port=
     def_delegators :@uri, :path, :normalized_path, :path=
     def_delegators :@uri, :query, :normalized_query, :query=
+    def_delegators :@uri, :query_values, :query_values=
     def_delegators :@uri, :request_uri, :request_uri=
     def_delegators :@uri, :fragment, :normalized_fragment, :fragment=
     def_delegators :@uri, :omit, :join, :normalize
--- a/spec/lib/http/client_spec.rb
+++ b/spec/lib/http/client_spec.rb
@@ -156,6 +156,14 @@ RSpec.describe HTTP::Client do
 
       client.get("http://example.com/", :params => {:t => "1970-01-01T00:00:00Z"})
     end
+
+    it 'does not convert newlines into \r\n before encoding string values' do
+      expect(HTTP::Request).to receive(:new) do |opts|
+        expect(opts[:uri].query).to eq "foo=bar%0Abaz"
+      end
+
+      client.get("http://example.com/", :params => {:foo => "bar\nbaz"})
+    end
   end
 
   describe "passing multipart form data" do


spec/spec_helper.rb
```diff
@@ -72,7 +73,7 @@
   # Print the 10 slowest examples and example groups at the
   # end of the spec run, to help surface which specs are running
   # particularly slow.
-  config.profile_examples = 10
+  # config.profile_examples = 10
```
ixti: You can achieve this by:

``` sh
echo '--fail-fast --no-profile' >> .rspec-local

File: spec/lib/http/uri_spec.rb [hunk 2]
```diff
@@ -29,4 +29,18 @@
       expect(http_uri.to_s).to eq("http://example.com")
     end
   end
+
+  describe ".encode_query" do
```
ixti: It worth to add spec that covers problem explained in the issue:

``` ruby
describe "#request" do
  it "correctly encodes new line chars in query params" do
    str = "bar\nbaz"
    res = HTTP.get("http://httpbin.org/get", :params => { :foo => str })
    expect(res.parse(:json).dig("args", "foo")).to eq str
  end
end

```

lib/http/uri.rb
```diff
@@ -36,14 +36,15 @@ def self.parse(uri)
...
     #
-    # @param [#to_hash, #to_ary] form_values to encode
-    # @param [TrueClass, FalseClass] sort should key/value pairs be sorted first?
+    # @param [#to_hash, #to_ary] values to encode
     #
-    # @return [String] encoded value
-    def self.form_encode(form_values, sort = false)
-      Addressable::URI.form_encode(form_values, sort)
+    # @return [String] encoded query string
+    def self.encode_query(values)
```
ixti: Please do not remove `form_encode` method in this PR. I'm fine if we will drop this but this is a backward compatibility breakage. And I think it was added here for some reason back in days.

@tarcieri do you remember if we actually need this method to be available on `HTTP::URI` or is it OK to completly remove it?

lib/http/client.rb
```diff
@@ -123,7 +123,7 @@ def make_request_uri(uri, opts)
       uri = HTTP::URI.parse uri
 
       if opts.params && !opts.params.empty?
-        uri.query = [uri.query, HTTP::URI.form_encode(opts.params)].compact.join("&")
+        uri.query = [uri.query, HTTP::URI.encode_query(opts.params)].compact.join("&")
```
ixti: I think we should expose `:query_values` method on `HTTP::URI` (delegate it to underlying `@uri`), this in here we will be able to:

``` ruby
uri.query_values = uri.query_values(Array).concat(opts.params.to_a)
```

Thus there will be no need in exposing `encode_query`.

spec/spec_helper.rb
mikegee: Oops. I didn't mean to commit this.

ixti: I got this. :D Just saying that you can use `.rspec-local` to tune up rspec to your tastes :D

spec/lib/http/uri_spec.rb
mikegee: I think this minor string generation bug is covered sufficiently by a unit test and isn't worth a high-level integration test like that.

(I'm assuming the project has some integration tests like you described, to confirm basic functionality.)

lib/http/uri.rb
mikegee: I assumed it was only used internally. I never considered this gem providing a URI class for external usage, but that makes sense if folks use it that way.

PR #451, opened 2018-01-29
Reduce memory usage when reading response body

janko-m:
Currently for each chunk read from the socket a new string is being allocated. This is not necessary, because we're feeding that string object into the HTTP parser and never using it again.

So we initialize a buffer string and pass it in when reading chunks. This will make each chunk be read into the same buffer object, no new string will be allocated in that operation.

When profiling download of a 18MB file, before this change there was 48MB of strings being allocated, and after this change it's about 18MB. This is **more than 50%** reduction in memory allocation.

### DIFF
--- a/lib/http/connection.rb
+++ b/lib/http/connection.rb
@@ -35,6 +35,7 @@ module HTTP
       @pending_request      = false
       @pending_response     = false
       @failed_proxy_connect = false
+      @buffer               = "".b
 
       @parser = Response::Parser.new
 
@@ -210,7 +211,7 @@ module HTTP
     def read_more(size)
       return if @parser.finished?
 
-      value = @socket.readpartial(size)
+      value = @socket.readpartial(size, @buffer)
       if value == :eof
         @parser << ""
         :eof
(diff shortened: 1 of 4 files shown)


zanker:
Seems reasonable to me. We have enough specs on request states that I'd expect it to have broken if the http_parser gem was somehow reusing the string, and otherwise it does look like `@buffer` isn't reused once it's passed to the C gem.

ixti:
Looks good to me as well. Merging this in!
@janko-m thank you!

janko-m:
I know that the `master` has been switched to `4.0.0.dev`, but any chance that a new `3.0.x` patch version can be released with this change? It makes a huge difference when timeouts are used:

```rb
HTTP
  .timeout(read: 2)
  .get("http://example.com/a-50MB-file")
  .body.each { |chunk| chunk.clear }
```
```
Before: 533MB
After: 15MB
```

ixti:
@janko-m I'll work on this.

ixti:
Released as 3.2.0 (somehow I merged EVERY bugfix from master to 3-x-stable but this and released as 3.1.0, so had to release 3.2.0 immediately after that one). :D

PR #455, opened 2018-02-05
Add support for dumping SSL master keys

amasses:
To enable SSL/TLS decryption in Wireshark the master keys need to be saved to a file. This change adds an option to specify the file, and if this is present the keys will be written to file after every request.

This can be a security risk (if the premaster keys are dumped out somewhere in Production environments etc), so the default value for this is `nil` which results in the function not being executed.


tarcieri:
Just a quick note on terminology: this PR appears to be extracting the master secret, not the pre-master secret. To my knowledge PMS is inextricable, at least in OpenSSL.

amasses:
Thanks for pointing that out - this was the terminology I had copied from Wireshark, but they do use it as (pre)-master keys...

tarcieri:
Wireshark appears to support both (hence the parenthetical), as the master
secret can be deterministicaly derived from the pre-master secret.

However, it really looks to me like “Master-Key” is the master secret, and
not PMS.
-- 
Tony Arcieri

tarcieri:
I think this is an OK feature to have, but the PR needs updated and I still think the terminology is incorrect

tarcieri:
Closing this as stale. Please open a new PR with the concerns addressed if you'd still like to get this merged.

PR #456, opened 2018-02-11
Simplify HTTP::Connection#read_headers!

janko-m:
This simplifies the implementation of `HTTP::Connection#read_headers!`, the behaviour should be preserved.

### DIFF
--- a/lib/http/connection.rb
+++ b/lib/http/connection.rb
@@ -97,13 +97,9 @@ module HTTP
     # Reads data from socket up until headers are loaded
     # @return [void]
     def read_headers!
-      loop do
-        if read_more(BUFFER_SIZE) == :eof
-          raise ConnectionError, "couldn't read response headers" unless @parser.headers?
-          break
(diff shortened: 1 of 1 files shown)


ixti:
I have a feeling that there was a reason why this part was written like that. `read_more` feeds data to the parser, so it should be performed before asking parser about anything.

tarcieri:
Looks good, thanks!

tarcieri:
Oh whoops, just noticed @ixti's comment now. My bad.

janko-m:
@ixti From the functionality perspective it should be the same, as the `@headers` variable is defined only after . If you'd find it more intuitive that `@parser.headers?` is called after the first `read_more`, I could swap

```rb
until @parser.headers?
  # ...
end
```

with

```rb
begin
 # ...
end until @parser.headers?
```

ixti:
I think that's fine in fact. I might have mistook this with other place - and as we don't have specs for this it seems like more likely.

PR #458, opened 2018-02-12
Extract HTTP::Client#build_request method

tycooon:
It would sometimes be handy to have a public method that returns a request object so that we can modify it and/or pass around and then use `perform` on it. 
Please take a look and if you find the idea is OK I will add specs etc.

### DIFF
--- a/lib/http/client.rb
+++ b/lib/http/client.rb
@@ -25,13 +25,25 @@ module HTTP
 
     # Make an HTTP request
     def request(verb, uri, opts = {}) # rubocop:disable Style/OptionHash
+      opts = @default_options.merge(opts)
+      req = prepare_request(verb, uri, opts)
+      res = perform(req, opts)
+      return res unless opts.follow
+
+      Redirector.new(opts.follow).perform(req, res) do |request|
+        perform(request, opts)
+      end
+    end
+
+    # Prepare an HTTP request
+    def prepare_request(verb, uri, opts = {}) # rubocop:disable Style/OptionHash
       opts    = @default_options.merge(opts)
       uri     = make_request_uri(uri, opts)
       headers = make_request_headers(opts)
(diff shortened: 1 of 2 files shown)


ixti:
I think we will completely refactor this bit in v4.0.0; I think that after all `#request` must return `Request` instance, that will respond to `#perform`. In any case I'm fine to merge it to [v3.x branch](https://github.com/httprb/http/tree/3-x-stable) and to the master (for sake of keeping things in sync).

@tarcieri any objections?

lib/http/client.rb
```diff
@@ -25,27 +25,32 @@ def initialize(default_options = {})
...
+      res = perform(req, opts)
+      return res unless opts.follow
+
+      Redirector.new(opts.follow).perform(req, res) do |request|
+        perform(request, opts)
+      end
+    end
+
+    # Prepare an HTTP request
+    def prepare_request(verb, uri, opts = {}) # rubocop:disable Style/OptionHash
```
ixti: Please rename this to `build_request`. IMO it would better reflect what method does.

tycooon: Sure, done.

tycooon:
So, is it going to be merged?

ixti:
Yeah. I think it's good. Thank you!

tycooon:
@ixti Can we backport it to version 3 by the way?

ixti:
@tycooon Sure. It does not brings any breaking changes - so totally fine - will handle this.

ixti:
@tycooon REleased as 3.3.0

shanecav84:
What's the way to perform the built request?

shanecav84:
What I now is:

```ruby
request = HTTP.build_request(...)
client = HTTP::Client.new
client.perform(request, client.default_options)
```

Issue #389, opened 2017-01-20
HTTP::Request#headline raises InvalidURIError when path is equal to //


scarfacedeb:
I've encountered InvalidURIError error when dealing with the redirects to a location with `//` path. (e.g. `https://google.com//`).

After further investigation, I realized that HTTP doesn't allow to make a request to such urls at all.

The crux of the matter is [HTTP::Request#headline](https://github.com/httprb/http/blob/acf5dbeb5aa8c4c62a6bba01c648eb254a1dde2a/lib/http/request.rb#L144) method that uses `Addressable::URI#omit` method to get the path + query portion of the given url.

And while addressable allows to parse such urls `http://google.com//` and most of the methods work as expected, it fails with `InvalidURIError` when `uri.omit(:authority)` is called.

Relevant line in Addressable source code: https://github.com/sporkmonger/addressable/blob/8f4fee07243af0b788c625e59a9ca529489336d6/lib/addressable/uri.rb#L2427

Steps to reproduce
===

```ruby
HTTP.get("https://google.com//")
#Addressable::URI::InvalidURIError: Cannot have a path with two leading slashes without anauthority set: '//'
```

Reproducing with Addressable directly:
```ruby
# Parse url with // path
url = Addressable::URI.parse("https://google.com//")

# Check that it works as expected
url.path == "//"

# Try to remove authority
url.omit(:authority)  # raises InvalidURIError
```

----

I was going to submit this issue to the addressable repository at first, but after the consideration I think that it's primarily the bug in HTTP.

Because Addressable is following the spec that disallows path with two leading slashes without the authority.

---

Ruby version: `ruby 2.3.3p222 (2016-11-21 revision 56859) [x86_64-darwin16]`
Gem version: `http (2.1.0)` and `addressable (2.5.0)`

ixti:
Hmm... @httprb/core I guess I agree. Probably we should "fix" this on our end. Something like:

``` ruby
uri = uri.path.sub(%r{^//+}, "/") + uri.query + uri.anchor
```

PR #462, opened 2018-02-25
Fix HTTP::Request#headline to allow two leading slashes in path

scarfacedeb:
HTTP.get doesn't allow to make requests to urls with two leading slashes (e.g. `https://google.com//`).

See previous discussion at #389.

Steps to reproduce
===

```ruby
HTTP.get("https://google.com//")
#Addressable::URI::InvalidURIError: Cannot have a path with two leading slashes without anauthority set: '//'
```

Now it's possible to use such urls.


ixti:
Looks good. Thanks!

PR #464, opened 2018-03-15
Fix broken HTTP::Connection#readpartial

marshall-lee:
800c5815 broke `readpartial` method: it returns same chunk twice :(

Ref #454.

BTW it's a good example of that public accessors in behavior (non-data) classes are harmful.


janko-m:
It's a bit unfortunate that none of the current tests failed 😕

ixti:
Thank you!

PR #467, opened 2018-04-22
Drop Ruby 2.2 support for (future) 4.x+ releases

ixti:


ixti:
@httprb/core any objections?

tarcieri:
Fine by me

PR #468, opened 2018-04-24
Rewind IO-alike request body after it was consumed.

ixti:
Webmock triggers `body.each` twice upon inital request recording:

1. to perform request
2. to record request data into webmock request registry

But even without WebMock, generally Body#each must be replayable IMO.

### DIFF
--- a/lib/http/request/body.rb
+++ b/lib/http/request/body.rb
@@ -35,6 +35,7 @@ module HTTP
           yield @source
         elsif @source.respond_to?(:read)
           IO.copy_stream(@source, ProcIO.new(block))
+          @source.rewind if @source.respond_to?(:rewind)
         elsif @source.is_a?(Enumerable)
           @source.each(&block)
         end
--- a/spec/lib/http/request/body_spec.rb
+++ b/spec/lib/http/request/body_spec.rb
@@ -126,10 +126,26 @@ RSpec.describe HTTP::Request::Body do
     end
 
     context "when body is an Enumerable IO" do
-      let(:body) { StringIO.new("a" * 16 * 1024 + "b" * 10 * 1024) }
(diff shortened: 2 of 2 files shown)


ixti:
I'm gonna merge this down and release as v 3.2.1
And then will prepare this patch with extra idea for master branch

tycooon:
Would be cool to have specs for that.

ixti:
@tycooon good catch! Thank you! done.

ixti:
I'm gonna "harden" readable body source validation in master - so that if given io-alike object does not respond to size and rewind - it will fail. Not gonna add it as part of 3.x branch as it's quiet breaking change for people using non-rewindable objects and not using webmock :D

mikker:
We're using `IO.pipe` for chunked POST requests in @elastic's [APM Agent](https://github.com/elastic/apm-agent-ruby) and I'm stubbing away rewind on my reader objects because of this.

```ruby
# HTTP.rb calls #rewind the body stream which IO.pipes don't support
class ModdedIO < IO
  def self.pipe(ext_enc = nil)
    super(ext_enc).tap do |rw|
      rw[0].define_singleton_method(:rewind) { nil }
    end
  end
end
```

Just wanted to provide an actual use of IO.pipe.

ixti:
@mikker I have opened an issue with your comment. Let's continue discussion there. We ended up with not requiring IO to respond to rewind:

https://github.com/httprb/http/blob/f5ddbaebcf3b6e18810f0a6bde2f8433dfae1621/lib/http/request/body.rb#L46-L53

But, I have a feeling that the problem might come from the form data:

https://github.com/httprb/form_data/blob/02800358e4a794e33dfc9cd5308e21ff5efccd2a/lib/http/form_data/composite_io.rb#L56-L60

PR #471, opened 2018-05-10
operation-specific timeout subclasses

### LINKED ISSUE #470: Feature request: timeout subclasses
As of this time, all timeouts raise an `HTTP::TimeoutError`. I'd like to be able to handle connect timeout errors differently from read or write timeout errors. For example, I want to have a lower connect timeout which I expect to hit occasionally, but also aggressively retry. Because the error happened on connect, I know that no data was sent yet and it's therefore always safe to retry. However, if the error happened on read, I need to take special precautions to see if the server processed the request or not (naturally, the 3rd party API is not idempotent).

I think this could easily be solved, at least for the [per operation timeout](https://github.com/httprb/http/blob/master/lib/http/timeout/per_operation.rb) class. We could have subclasses of `HTTP::TimeoutError` that describe the specific type of timeout that was encountered.

Is that something you'd accept? I can take a crack at a PR if so.

bgentry:
Use separate subclasses for each type of operation that can timeout: connect, write, and read.

I also tried to add more test coverage to handle the different timeout scenarios. There are two main issues I ran into:

* Testing a connect timeout of 0 just never works locally. This could easily be done using an external server (i.e. hitting `api.github.com` as the tests do elsewhere) but that wouldn't work cleanly with the way these shared examples are used for SSL vs non-SSL. That was a bigger change than I wanted to make here.
* The only way I could reliably trigger a write timeout of 0 was with a sizable request body. However, if the write timeout is increased in this scenario, I'm seeing broken pipe connection errors. Would love to hear ideas on how to make this work.

Fixes #470.

### DIFF
--- a/lib/http/timeout/global.rb
+++ b/lib/http/timeout/global.rb
@@ -21,12 +21,12 @@ module HTTP
 
       def connect(socket_class, host, port, nodelay = false)
         reset_timer
-        ::Timeout.timeout(@time_left, TimeoutError) do
+        ::Timeout.timeout(@time_left, ConnectTimeoutError) do
           @socket = socket_class.open(host, port)
           @socket.setsockopt(Socket::IPPROTO_TCP, Socket::TCP_NODELAY, 1) if nodelay
         end
 
-        log_time
+        log_time(ConnectTimeoutError)
       end
 
       def connect_ssl
@@ -36,11 +36,11 @@ module HTTP
           @socket.connect_nonblock
         rescue IO::WaitReadable
           IO.select([@socket], nil, nil, @time_left)
-          log_time
+          log_time(ConnectTimeoutError)
           retry
         rescue IO::WaitWritable
           IO.select(nil, [@socket], nil, @time_left)
-          log_time
+          log_time(ConnectTimeoutError)
           retry
         end
       end
@@ -104,13 +104,13 @@ module HTTP
       # Wait for a socket to become readable
       def wait_readable_or_timeout
         @socket.to_io.wait_readable(@time_left)
-        log_time
+        log_time(ReadTimeoutError)
       end
 
       # Wait for a socket to become writable
       def wait_writable_or_timeout
         @socket.to_io.wait_writable(@time_left)
-        log_time
+        log_time(WriteTimeoutError)
       end
 
       # Due to the run/retry nature of nonblocking I/O, it's easier to keep track of time
@@ -119,10 +119,10 @@ module HTTP
         @started = Time.now
       end
 
-      def log_time
+      def log_time(klass)
         @time_left -= (Time.now - @started)
         if @time_left <= 0
-          raise TimeoutError, "Timed out after using the allocated #{@timeout} seconds"
+          raise klass, "Timed out after using the allocated #{@timeout} seconds"
         end
 
         reset_timer
--- a/lib/http/timeout/null.rb
+++ b/lib/http/timeout/null.rb
@@ -56,19 +56,29 @@ module HTTP
       private
 
       # Retry reading
-      def rescue_readable(timeout = read_timeout)
+      def rescue_readable(klass, timeout = read_timeout)
         yield
       rescue IO::WaitReadable
         retry if @socket.to_io.wait_readable(timeout)
-        raise TimeoutError, "Read timed out after #{timeout} seconds"
+        raise klass, "#{error_name_for_class(klass)} timed out after #{timeout} seconds"
       end
 
       # Retry writing
-      def rescue_writable(timeout = write_timeout)
+      def rescue_writable(klass, timeout = write_timeout)
         yield
       rescue IO::WaitWritable
         retry if @socket.to_io.wait_writable(timeout)
-        raise TimeoutError, "Write timed out after #{timeout} seconds"
+        raise klass, "#{error_name_for_class(klass)} timed out after #{timeout} seconds"
+      end
+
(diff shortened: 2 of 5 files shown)


lib/http/timeout/global.rb
```diff
@@ -119,10 +119,10 @@ def reset_timer
         @started = Time.now
       end
 
-      def log_time
+      def log_time(klass)
```
bgentry: Because this is a private method, I'm assuming it was ok to modify its method signature. That was the most straightforward way I could see to handle all cases.

lib/http/timeout/null.rb
```diff
@@ -56,19 +56,31 @@ def write(data)
       private
 
       # Retry reading
-      def rescue_readable(timeout = read_timeout)
+      def rescue_readable(klass, timeout = read_timeout)
```
bgentry: same comment re: method signature on private method.

lib/http/timeout/null.rb
```diff
@@ -56,19 +56,31 @@ def write(data)
...
+      def error_name_for_class(klass)
+        if klass == ConnectTimeoutError
+          "Connect"
+        elsif klass == ReadTimeoutError
+          "Read"
+        elsif klass == WriteTimeoutError
+          "Write"
+        else
+          raise "Unknown timeout error type: #{klass}"
+        end
```
bgentry: Because the method argument is the class itself (and not an instance of it) a `case` statement doesn't work here (it uses `===`). Any ideas to clean this up?

lib/http/timeout/null.rb
```diff
@@ -56,19 +56,31 @@ def write(data)
...
       end
 
       # Retry writing
-      def rescue_writable(timeout = write_timeout)
+      def rescue_writable(klass, timeout = write_timeout)
         yield
       rescue IO::WaitWritable
         retry if @socket.to_io.wait_writable(timeout)
-        raise TimeoutError, "Write timed out after #{timeout} seconds"
+        raise klass, "#{error_name_for_class(klass)} timed out after #{timeout} seconds"
```
bgentry: CodeClimate [does not like](https://codeclimate.com/github/httprb/http/pull/471) the similarity of these methods. I'm not sure abstracting out would improve the overall quality though, thoughts?

lib/http/timeout/null.rb
tycooon: You can use case on klass.name for example.

tycooon: Using strings is bad in this case, however. I would suggest using a constant hash for this mapping.

bgentry: Yeah, that's much better. Changed, thanks :v:

lib/http/timeout/per_operation.rb
```diff
@@ -20,15 +20,15 @@ def initialize(*args)
...
+        ::Timeout.timeout(@connect_timeout, ConnectTimeoutError) do
           @socket = socket_class.open(host, port)
           @socket.setsockopt(Socket::IPPROTO_TCP, Socket::TCP_NODELAY, 1) if nodelay
         end
       end
 
       def connect_ssl
-        rescue_readable(@connect_timeout) do
-          rescue_writable(@connect_timeout) do
+        rescue_readable(ConnectTimeoutError, @connect_timeout) do
```
zanker: I'd argue that this should be part of write/read timeouts, and not connect. The operations under the hood for `connect_ssl` are read/write as part of the SSL handshake. It's not the same as a connect timeout because of DNS resolution or the server isn't accepting requests, so we shouldn't mask it that way.

(A nice side effect is it makes the code simpler, since don't have to pass classes to `rescue_####`)

bgentry: yeah, I went back and forth on this. Happy to change it to the way you suggest if that's what makes the most sense.

zanker: I'd opt for consistency against the ops used personally, but @ixti or @tarcieri? Do either of you have any feelings on that.

bgentry: I think my reasoning for keeping it as-is would be to support the use case I described in #470. I mainly care about whether the request has been sent yet. As written this considers everything pre-request as a `ConnectionTimeoutError`, rather than giving read/write timeouts when the request hasn't even been sent (during the TLS handshake).

lib/http/timeout/null.rb
ixti: I think it's fine for not to ignore this code climate regression. We can think about that later. I don't see that abstracting this will help atm as well.

lib/http/timeout/per_operation.rb
ixti: [I have no strong feelings one way or the other](https://www.youtube.com/watch?v=CxK_nA2iVXw). But after couple of moments thinking and sketching I believe `ConnectTimeoutError` here is probably the best.

@zanker, as alternative we can make `connect_ssl` raise `HandshakeTimeoutError` that will be subclass of `ConnectTimeoutError` if you feel strong about that.

janko-m: Yeah, as much as I would like to simplify the implementation by using `ReadTimeoutError`, I also feel like `ConnectTimeoutError` would be more correct, so I'm ok with the current implementation.

tarcieri:
This is (was) a good PR with some good ideas. It seems we should've merged it. I know timeouts are a pain in the ass to test.

Please reopen after rebasing if you're still interested.

PR #472, opened 2018-05-11
Accept output buffer in Response::Body#readpartial

janko-m:
This makes it possible to use a `HTTP::Response::Body` object as a source argument in `IO.copy_stream`, allowing easy streaming. It also allows it to be used in the request body for another request (as `HTTP::Request::Body::Writer` uses `IO.copy_stream`), should anyone want that.

```rb
response = HTTP.get("http://example.com/download")
IO.copy_stream(response.body, destination)
```

Since `IO.copy_stream` uses an output buffer, it makes retrieving the response body much more memory efficient, because we're able to deallocate retrieved chunks. I tested with a 50MB response body, and streaming with `HTTP::Response::Body#each` allocated about 52MB of strings, whereas the `IO.copy_stream` version allocated only 3MB.

`IO.copy_stream` requires `#readpartial` to have `IO#readpartial` semantics, which among others includes raising and `EOFError` when end of file was reached. Since `IO.copy_stream` always uses an output buffer, the exception is raised only when output buffer is specified.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,7 +17,7 @@ env: JRUBY_OPTS="$JRUBY_OPTS --debug"
 
 rvm:
   # Include JRuby first because it takes the longest
-  - jruby-9.1.16.0
+  - jruby-9.2.0.0
   - 2.3
   - 2.4
   - 2.5
--- a/lib/http/connection.rb
+++ b/lib/http/connection.rb
@@ -86,6 +86,7 @@ module HTTP
     def readpartial(size = BUFFER_SIZE)
       return unless @pending_response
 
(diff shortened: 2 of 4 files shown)


janko-m:
This one needs JRuby 9.2.0.0 to pass the tests (which has the fix for https://github.com/jruby/jruby/issues/5167), but it's currently failing because of https://github.com/httprb/http/issues/475.

janko-m:
Now that I think about this more, this is probably an overkill. I don't like the idea of `HTTP::Response::Body#readpartial` partially mimicking `IO#readpartial`, it doesn't need to act as an IO object. People can just clear yielded chunks themselves if they want to save on memory.

PR #473, opened 2018-05-23
Handle early responses

janko-m:
HTTP servers typically wait until the whole request body has been received before sending the response. However, some servers might send out a response early, before they've read the whole request body. For example, they might detect the request headers were invalid, and they want to tell that to the client immediately.

For example, sending a large request to http://google.com will return an early `400 Bad Request` response. In this case http.rb will raise an `Errno::EPIPE` exception (wrapped in a `HTTP::ConnectionError`), because it tried to write to the socket but it couldn't because the server closed its end. However, the server still returned a valid response, so it's not really erroneous behaviour (e.g. curl doesn't print any errors in the same situations).

  ```rb
  # Before
  HTTP.get("http://google.com", body: "a" * 1*1024*1024)
  #~> Errno::EPIPE

  # After
  HTTP.get("http://google.com", body: "a" * 1*1024*1024)
  #=> #<HTTP::Response 400 Bad Request ...>
  ```

This commit changes `Request::Writer` to stop writing when an `Errno::EPIPE` exception occurred.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -13,6 +13,11 @@ Layout/SpaceAroundOperators:
 Layout/SpaceInsideHashLiteralBraces:
   EnforcedStyle: no_space
 
+## Lint ########################################################################
+
+Lint/HandleExceptions:
+  Enabled: false
+
 ## Metrics #####################################################################
 
 Metrics/AbcSize:
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -13,11 +13,6 @@ Layout/SpaceAroundOperators:
 Layout/SpaceInsideHashLiteralBraces:
   EnforcedStyle: no_space
 
-## Lint ########################################################################
-
-Lint/HandleExceptions:
-  Enabled: false
-
 ## Metrics #####################################################################
 
 Metrics/AbcSize:
--- a/lib/http/request/writer.rb
+++ b/lib/http/request/writer.rb
@@ -60,24 +60,32 @@ module HTTP
         @request_header.join(CRLF) + CRLF * 2
       end
 
+      # Writes HTTP request data into the socket.
       def send_request
-        # It's important to send the request in a single write call when
-        # possible in order to play nicely with Nagle's algorithm. Making
-        # two writes in a row triggers a pathological case where Nagle is
-        # expecting a third write that never happens.
+        each_chunk { |chunk| write chunk }
+      rescue Errno::EPIPE
+        # server doesn't need any more data
+        nil
+      end
+
+      # Yields chunks of request data that should be sent to the socket.
+      #
+      # It's important to send the request in a single write call when possible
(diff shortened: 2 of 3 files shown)


.rubocop.yml
```diff
@@ -13,6 +13,11 @@ Layout/SpaceAroundOperators:
 Layout/SpaceInsideHashLiteralBraces:
   EnforcedStyle: no_space
 
+## Lint ########################################################################
+
+Lint/HandleExceptions:
+  Enabled: false
```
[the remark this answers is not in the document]
janko-m: Agreed, updated.

lib/http/request/writer.rb
```diff
@@ -100,6 +111,8 @@ def write(data)
           break unless data.bytesize > length
           data = data.byteslice(length..-1)
         end
+      rescue Errno::EPIPE
+        raise # re-raise Errno::EPIPE
```
ixti: IMO this code is pretty self-explanatory and doesn't need comment.

lib/http/request/writer.rb
```diff
@@ -60,22 +60,33 @@ def join_headers
...
       def send_request
-        # It's important to send the request in a single write call when
-        # possible in order to play nicely with Nagle's algorithm. Making
-        # two writes in a row triggers a pathological case where Nagle is
-        # expecting a third write that never happens.
+        each_chunk do |chunk|
+          write(chunk)
+        end
+      rescue Errno::EPIPE # rubocop:disable Lint/HandleExceptions
+        # server doesn't need any more data
```
ixti: IMO it worth to explicitly add `nil` at the end of `rescue` block and remove rubocop disable.

ixti: Oh. It was discussed previously... Mmm... I still think we should just return `nil` from the `rescue`. Also body can be shortened:

``` ruby
def send_request
  each_chunk { |chunk| write chunk }
rescue Errno::EPIPE
  # no more data
  nil
end
```

ixti: On one hand I think that `Lint/HandleExceptions` is pretty useless and enforces developer to write more than needed. On the other hand I prefer to see:

``` ruby
begin
  # ...
rescue
  nil
end
```

than

``` ruby
begin
  # ...
rescue
end
```

IMO it's more readable that way. Easier to understand that `nil` is result in case of error.

ixti: I know that one might argue that we don't require explicit nil in case of `return if something`, but it's a bit different - it's one-liner. And reading it is pretty straight-forward to me. And when we explicitly return from non-one-liner code we add `nil` to return:

``` ruby
if something
  # ...
  return nil # <-- explicit `nil` here looks better than bare `return`
end
```

janko-m: Agreed, updated.

janko-m:
@ixti Done.

ixti:
Thanks!

PR #489, opened 2018-07-06
Fix HTTP parsing (second attempt)

fxposter:
Basically this is a https://github.com/httprb/http/pull/433 PR, but rebased against current master with 2 changed lines related to
error handling (http-parser returns a boolean for an indication of a failure, while http_parser.rb actually threw exception there), so that now all tests pass.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -5,8 +5,6 @@ ruby RUBY_VERSION
 
 gem "rake"
 
-gem "http_parser.rb", :git => "https://github.com/ixti/http_parser.rb.git"
-
 group :development do
   gem "guard-rspec", :require => false
   gem "nokogiri",    :require => false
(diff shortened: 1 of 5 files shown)


ixti:
Awesome! Thanks!

I'll take care of this.

tarcieri:
This seems simple enough. @ixti mind taking another look?

cognita:
Will this be merged in the near future?
Could we provide any kind of support to make progress on this?

ixti:
Yeah. Sorry had a pretty tough fourth quarter of 2018. Things are finally getting back in norm, so will work on this shortly. Plan is to merge this PR almost as is - with 3rd party dependency and declare HTTP::Parser as private API, and then when I will have time on thinking of good API we will make it public and slowly migrate to own FFI binding.

PR #496, opened 2018-08-08
Always encode headers as a binary string

### LINKED ISSUE #490: incompatible character encodings: UTF-8 and ASCII-8BIT
Hi, i'm having some issues sending a request with form data.
I'm getting this error on post call.

> 10: from /var/lib/gems/2.5.0/gems/http-3.3.0/lib/http/chainable.rb:27:in `post'
> 	 9: from /var/lib/gems/2.5.0/gems/http-3.3.0/lib/http/client.rb:30:in `request'
> 	 8: from /var/lib/gems/2.5.0/gems/http-3.3.0/lib/http/client.rb:70:in `perform'
> 	 7: from /var/lib/gems/2.5.0/gems/http-3.3.0/lib/http/connection.rb:76:in `send_request'
> 	 6: from /var/lib/gems/2.5.0/gems/http-3.3.0/lib/http/request.rb:115:in `stream'
> 	 5: from /var/lib/gems/2.5.0/gems/http-3.3.0/lib/http/request/writer.rb:38:in `stream'
> 	 4: from /var/lib/gems/2.5.0/gems/http-3.3.0/lib/http/request/writer.rb:70:in `send_request'
> 	 3: from /var/lib/gems/2.5.0/gems/http-3.3.0/lib/http/request/body.rb:37:in `each'
> 	 2: from /var/lib/gems/2.5.0/gems/http-3.3.0/lib/http/request/body.rb:37:in `copy_stream'
> 	 1: from /var/lib/gems/2.5.0/gems/http-3.3.0/lib/http/request/body.rb:64:in `write'
> /var/lib/gems/2.5.0/gems/http-3.3.0/lib/http/request/writer.rb:71:in `block in send_request': incompatible character encodings: UTF-8 and ASCII-8BIT (Encoding::CompatibilityError)
> 

I looked it doesnt crash when i remove a file from the form data.

This is my form data: removing file it appears to work correctly.

`{:form=>{:_id=>"5a61b17f26ef95e96dd01cd5", :docType=>"DEFAULT", :name=>"a", :clientId=>"57e40c903f9eb69974c0249e", :description=>"a", :entity=>"56d361d7478dc883f1d63bea", :isVisible=>true, :isRequired=>false, :file=>#<HTTP::FormData::File:0x0000557712865f68 @io=#<File:/home/a/a/a/lib/upload/a-_OP-201800848_a_a_a_a.pdf>, @content_type="application/octet-stream", @filename="a-_OP-201800848_a_a_a_a.pdf">}}
`
I don't know what to do now..

mikeastock:
This prevents the issue of encoding errors when trying to concat binary
bodies with this header string.

This should resolve #490

### DIFF
--- a/lib/http/request/writer.rb
+++ b/lib/http/request/writer.rb
@@ -65,7 +65,7 @@ module HTTP
         # possible in order to play nicely with Nagle's algorithm. Making
         # two writes in a row triggers a pathological case where Nagle is
         # expecting a third write that never happens.
-        data = join_headers
+        data = join_headers.force_encoding("BINARY")
 
         @body.each do |chunk|
           data << encode_chunk(chunk)
--- a/spec/lib/http/request/writer_spec.rb
+++ b/spec/lib/http/request/writer_spec.rb
@@ -47,6 +47,27 @@ RSpec.describe HTTP::Request::Writer do
       end
     end
 
+    context "when body is encoded as ASCII-8BIT and headers are encoded as UTF-8" do
+      let(:bad_string) do
+        (100..1000).to_a.each_with_object("".dup) do |char, string|
+          string << char
+        end
+      end
+
+      let(:body) { HTTP::Request::Body.new(bad_string.b) }
+
+      it "doesn't write anything to the socket and sets Content-Length" do
+        headers.set("Authorization", bad_string)
+        writer.stream
+        expect(io.string).to eq [
+          "#{headerstart}\r\n",
+          "Authorization: #{bad_string}\r\n",
+          "Content-Length: 1774\r\n\r\n",
+          bad_string
+        ].join
+      end
+    end
+
     context "when Content-Length header is set" do
       let(:headers) { HTTP::Headers.coerce "Content-Length" => "12" }
       let(:body)    { HTTP::Request::Body.new("content") }


tarcieri:
Would be good to check what the RFCs want encoding-wise, but my best guess would be this is (close to) the right behavior.

An alternative would be to reject strings that aren't `Encoding::BINARY`

mikeastock:
@tarcieri what can I do to help get this merged?

lib/http/request/writer.rb
```diff
@@ -65,7 +65,7 @@ def send_request
         # possible in order to play nicely with Nagle's algorithm. Making
         # two writes in a row triggers a pathological case where Nagle is
         # expecting a third write that never happens.
-        data = join_headers
+        data = join_headers.b
```
tarcieri: The disadvantage of `#b` is that it allocates. Since `join_headers` allocates the string anyway, it'd be more efficient to do `#force_encoding("BINARY")` here

ixti: Basically join_headers builds a string that can be forced to be binary upfront.

ixti: It can be actually even optimized a bit by pre-allocating string of needed length - it's pretty trivial to calculate that.

mikeastock: So I should move it to using `#force_encoding("BINARY")`?

tarcieri: Yes, that would be more efficient as it'd avoid duplicating the header buffer

mikeastock:
Should be good to go now

tarcieri:
Looks good to me.

@ixti any complaints?

tarcieri:
@ixti yeah definitely agree with more test cases.

Something that might help is making `charset` one of the request's `HTTP::Options`.

By default it could be `nil` i.e. don't make encoding assumptions, or essentially what it's doing today on `master`.

Setting it explicitly, e.g. `HTTP.charset(:binary)` or perhaps also `HTTP.charset(Encoding::UTF_8)` could `#force_encoding` on `data` using the specified encoding (i.e. it needs to match the specified encoding). If any of the body data does not match the specified encoding, it would raise an error.

With a `charset` option, we could even compute an `HTTP::ContentType` (e.g. add an `#encode` method in addition to `#parse`) by default unless one is specified, so if you specify an `HTTP.charset` it automatically gets added to `Content-Type: [...]; charset=[charset]`

mikeastock:
So next steps for this PR is move the encoding to an option?

tarcieri:
@mikeastock yeah, looks like that's the path forward

say8425:
Hey when will be merged this PR?

PR #506, opened 2018-10-29
Fix Instrumentation not inheriting Feature

### LINKED ISSUE #505: Instrumentation feature not working
The new instrumentation feature (https://github.com/httprb/http/pull/499) does not seem to work, using the simplest example case.

It expects an `instrumenter` object which responds to `instrument`, `start`, and `stop`:
https://github.com/httprb/http/blob/4bc1223c5bbaf85dfc142f353a30594242c7a4bb/lib/http/features/instrumentation.rb#L24-L25
which means we have to pass `ActiveSupport::Notifications.instrumenter` (which differs from the documented example).

But when I try to make a request, it appears to fail when I chain a `get` request:

```
2.5.1 :002 > require 'http'
 => true
2.5.1 :003 > require 'active_support/notifications'
 => true
2.5.1 :004 > HTTP.use(instrumentation: {instrumenter: ActiveSupport::Notifications.instrumenter}).get("https://httpbin.org/get")
       16: from (irb):4
       15: from /Users/josh/.rvm/gems/ruby-2.5.1/gems/http-4.0.0/lib/http/chainable.rb:20:in `get'
       14: from /Users/josh/.rvm/gems/ruby-2.5.1/gems/http-4.0.0/lib/http/client.rb:29:in `request'
       13: from /Users/josh/.rvm/gems/ruby-2.5.1/gems/http-4.0.0/lib/http/options.rb:165:in `merge'
       12: from /Users/josh/.rvm/gems/ruby-2.5.1/gems/http-4.0.0/lib/http/options.rb:23:in `new'
       11: from /Users/josh/.rvm/gems/ruby-2.5.1/gems/http-4.0.0/lib/http/options.rb:23:in `new'
       10: from /Users/josh/.rvm/gems/ruby-2.5.1gems/http-4.0.0/lib/http/options.rb:72:in `initialize'
        9: from /Users/josh/.rvm/gems/ruby-2.5.1/gems/http-4.0.0/lib/http/options.rb:72:in `each'
        8: from /Users/josh/.rvm/gems/ruby-2.5.1/gems/http-4.0.0/lib/http/options.rb:72:in `block in initialize'
        7: from /Users/josh/.rvm/gems/ruby-2.5.1/gems/http-4.0.0/lib/http/options.rb:188:in `[]='
        6: from /Users/josh/.rvm/gems/ruby-2.5.1/gems/http-4.0.0/lib/http/options.rb:110:in `features='
        5: from /Users/josh/.rvm/gems/ruby-2.5.1/gems/http-4.0.0/lib/http/options.rb:110:in `each_with_object'
        4: from /Users/josh/.rvm/gems/ruby-2.5.1/gems/http-4.0.0/lib/http/options.rb:110:in `each'
        3: from /Users/josh/.rvm/gems/ruby-2.5.1/gems/http-4.0.0/lib/http/options.rb:117:in `block in features='
        2: from /Users/josh/.rvm/gems/ruby-2.5.1/gems/http-4.0.0/lib/http/options.rb:117:in `new'
        1: from /Users/josh/.rvm/gems/ruby-2.5.1@/gems/http-4.0.0/lib/http/features/instrumentation.rb:18:in `initialize'
```

paul:
Fixes #505. There was a bug because the Instrumenter was getting initialized twice because the shortcut on https://github.com/httprb/http/blob/master/lib/http/options.rb#L111 wasn't hit without Instrumenter inheriting Feature.

I also fixed the docs a mentioned in the issue.

@ixti @tarcieri WDYT about adding `activesupport` as a dev dependency, so I can write a real test that exercises it being called correctly?

### DIFF
--- a/lib/http/features/instrumentation.rb
+++ b/lib/http/features/instrumentation.rb
@@ -9,10 +9,16 @@ module HTTP
     # Be sure to specify the instrumenter when enabling the feature:
     #
     #    HTTP
-    #      .use(instrumentation: {instrumenter: ActiveSupport::Notifications})
+    #      .use(instrumentation: {instrumenter: ActiveSupport::Notifications.instrumenter})
     #      .get("https://example.com/")
     #
-    class Instrumentation
+    # Emits two events on every request:
+    #
+    #  * `start_request.http` before the request is made, so you can log the reqest being started
+    #  * `request.http` after the response is recieved, and contains `start`
+    #    and `finish` so the duration of the request can be calculated.
+    #
+    class Instrumentation < Feature
       attr_reader :instrumenter, :name
 
       def initialize(instrumenter: NullInstrumenter.new, namespace: "http")
@@ -21,7 +27,9 @@ module HTTP
       end
 
       def wrap_request(request)
-        instrumenter.instrument("start_#{name}", :request => request)
+        # Emit a separate "start" event, so a logger can print the request
+        # being run without waiting for a response
+        instrumenter.instrument("start_#{name}", :request => request) {}
         instrumenter.start(name, :request => request)
         request
       end


paul:
@joshuaflanagan Can you check if this works for you?

```ruby
require "pp"
require "active_support/notifications"
ActiveSupport::Notifications.subscribe(/.*/) do |name, start, finish, id, payload|
  pp :name => name, :start => start.to_f, :finish => finish.to_f, :id => id, :payload => payload
end

HTTP.use(instrumentation: {instrumenter: ActiveSupport::Notifications.instrumenter}).get("https://httpbin.org/get")

{:name=>"start_request.http",
 :start=>1540825224.790576,
 :finish=>1540825224.7905908,
 :id=>"a1ecae467ff9e7c58964",
 :payload=>
  {:request=>
    #<HTTP::Request:0x000055e547d9bbb0
     @body=#<HTTP::Request::Body:0x000055e547d98758 @source=nil>,
     @headers=
      #<HTTP::Headers {"Connection"=>"close", "Host"=>"httpbin.org", "User-Agent"=>"http.rb/4.0.0"}>,
     @proxy={},
     @scheme=:https,
     @uri=#<HTTP::URI:0x0055e547d98de8 URI:https://httpbin.org/get>,
     @verb=:get,
     @version="1.1">}}

{:name=>"request.http",
 :start=>1540825224.792291,
 :finish=>1540825225.0600104,
 :id=>"a1ecae467ff9e7c58964",
 :payload=>
  {:response=>
    #<HTTP::Response/1.1 200 OK {"Connection"=>"close", "Server"=>"gunicorn/19.9.0", "Date"=>"Mon, 29 Oct 2018 15:00:25 GMT", "Content-Type"=>"application/json", "Content-Length"=>"195", "Access-Control-Allow-Origin"=>"*", "Access-Control-Allow-Credentials"=>"true", "Via"=>"1.1 vegur"}>}}

# => #<HTTP::Response/1.1 200 OK {"Connection"=>"close", "Server"=>"gunicorn/19.9.0", "Date"=>"Mon, 29 Oct 2018 15:00:25 GMT", "Content-Type"=>"application/json", "Content-Length"=>"195", "Access-Control-Allow-Origin"=>"*", "Access-Control-Allow-Credentials"=>"true", "Via"=>"1.1 vegur"}>
```

lib/http/features/instrumentation.rb
```diff
@@ -21,7 +27,9 @@ def initialize(instrumenter: NullInstrumenter.new, namespace: "http")
       end
 
       def wrap_request(request)
-        instrumenter.instrument("start_#{name}", :request => request)
+        # Emit a separate "start" event, so a logger can print the request
+        # being run without waiting for a response
+        instrumenter.instrument("start_#{name}", :request => request) {}
```
ixti: I dislike this part. I see reason behind this but it feels like trying to use instrumenter for something it's not supposed to be used.

ixti:
About  activesupport we actually have it as dev deps, but we were thinking towards getting rid of it.

ixti:
I would say it's safe for now to assume we have AS as dev dependency.

ixti:
We have activemodel dev dependency due to certificate of authority but we were thinking about getting rid of this. I would say that it's OK to add AS dependency for now, but I have a feeling that probably we should extract this feature into a separate gem, probably, e.g. `http-rails` or such. Have no strong feelings one way or the other thogh. @tarcieri would really like to know your thoughts.

lib/http/features/instrumentation.rb
paul: It's a common pattern in Rails itself: https://github.com/rails/rails/blob/b2eb1d1c55a59fee1e6c4cba7030d8ceb524267c/actionpack/lib/action_controller/metal/instrumentation.rb#L30-L32

I'm not a big fan of it either (I wish the `instrument` call would just automatically emit events at both the start and end of the block), but I assume this is how it was intended to be used.

joshuaflanagan: Doesn't it? https://github.com/rails/rails/blob/405aef3dede8938c8eb9789f6d6722864faddb09/activesupport/lib/active_support/notifications/instrumenter.rb#L21-L29
line 21 - start
line 29 - finish

paul: @joshuaflanagan I dug through the overly-complicated code for a few minutes, and I can't figure out why, but the subscribers don't receive any calls at the start of the block, just the end.

tarcieri:
I don't mind `activesupport` as a dev dependency. I agree extracting the functionality into a separate crate would probably be best.

lib/http/features/instrumentation.rb
joshuaflanagan: Ok, you're right. I misinterpreted that code. Line 21 is marking the start of the event, but nothing is published until the finish on line 29.

Confirmed in the docs (https://guides.rubyonrails.org/active_support_instrumentation.html#creating-custom-events):
> Simply call instrument with a name, payload and a block. The notification will be sent after the block returns.

ixti:
Merged into 4-x-stable, will be released as part of v4.0.2

PR #510, opened 2018-11-25
Use Ruby's built-in null logger

janko-m:
The `Logger` class from the Ruby standard library will create a null logger when `nil` is given as the output.

### DIFF
--- a/lib/http/features/logging.rb
+++ b/lib/http/features/logging.rb
@@ -1,5 +1,7 @@
 # frozen_string_literal: true
 
+require "logger"
+
 module HTTP
   module Features
     # Log requests and responses. Request verb and uri, and Response status are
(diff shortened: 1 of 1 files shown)


paul:
:+1:

ixti:
LGTM, but I'm not quite sure why we have default for logger in a first place. I mean IMO:

``` ruby
HTTP.use(:logging).get(url)
```

Should fail that `logger` was not specified, rather than confusingly skip logs.  I understand, that one might need to be able to disable logging, e.g.:

``` ruby
client = HTTP.use(:logging => { :logger => Logger.new })

client.use(:logging).get(url) # do not log this interaction...
```

But, seriously `client.use(:logging)` doesn't telling developer that it's actually (!) disables logging. Compare that to:

``` ruby
client.use(:logging => { :logger => Logger.new(nil) }).get(url)
```

r-obert:
agreed, it is totally unexpected for me that calling `client.use(:logging.get(..)` would disable logging.

ixti:
FWIW I would expect `HTTP.use(:logging)` to be a shorthand for `HTTP.use(:logging => { :logger => Logger.new(STDOUT) })` rather than null-logger.

Issue #512, opened 2018-12-05
Base URI


sj26:
I'm replacing a bunch of Faraday usage with the HTTP gem, it's fantastic. A common pattern I'm encountering is constructing a Faraday connection with a base URI, then doing requests using paths which are joined to the base URI.

```ruby
client = Faraday.new(url: "https://api.github.com")
response = client.get("/search/code", q: "repo:httprb/http uri")
# => #<Faraday::Response:0xabc123 @on_complete_callbacks=[], @env=#<Faraday::Env @method=:get @url=#<URI::HTTPS https://api.github.com/search/code?q=repo%3Ahttprb%2Fhttp+uri> ...>

```

I notice that the HTTP gem does something simliar when using persistent connections:

```ruby
client = HTTP.persistent("https://api.github.com")
response = http.get("/search/code", params: {q: "repo:httprb/http uri"})
# => #<HTTP::Request:0xabc123 @verb=:get, @uri=#<HTTP::URI:0xabc123 URI:https://api.github.com/code/search?q=...>, ...>
```

The Faraday version also works with path prefixes, so can use any URI as a base:

```ruby
search_client = Faraday.new(url: "https://api.github.com/search/")
response = search_client.get("code", q: "repo:httprb/http uri")
# => #<Faraday::Response:0xabc123 @on_complete_callbacks=[], @env=#<Faraday::Env @method=:get @url=#<URI::HTTPS https://api.github.com/search/code?q=repo%3Ahttprb%2Fhttp+uri> ...>
```

The same does not work with HTTP:

```ruby
client = HTTP.persistent("https://api.github.com/search/")
response = http.get("code", params: {q: "repo:httprb/http uri"})
# => HTTP::StateError (Persistence is enabled for https://api.github.com, but we got https://api.github.comcode)
```

This also requires using persistent connections, which may not always be desired.

Would a pull request be considered to add a base uri to options? Something like:

```ruby
client = HTTP.uri("https://api.github.com/search/")
client.get("code", params: {q: "repo:httprb/http uri"})
```

which sets `uri` on `HTTP::Options` so that `HTTP::Client#build_request` via `#make_request_uri` looks for a `uri` on options and if present performs a `Addressable::URI.join` on the uri from options and the uri parameter?

PR #513, opened 2018-12-05
Base URI

sj26:
Prototype of #512: Allow baking a base URI into an HTTP::Client. Super useful for creating an API client then requesting on paths alone, and making endpoints configurable:

```ruby
client = HTTP.uri("https://api.stripe.com/v1/").auth(...)
client.get("charges")
client.post("customers", json: {...})
# ...
```

Persistent gets part way there, but only allows specifying the origin excluding path prefixes, and requires opting in to persistent connections.

---
This change is [Reviewable](https://reviewable.io/reviews/httprb/http/513)

### DIFF
--- a/lib/http/chainable.rb
+++ b/lib/http/chainable.rb
@@ -245,6 +245,12 @@ module HTTP
       branch default_options.with_features(features)
     end
 
+    # Join base uri to all request uris
+    # @param uri
+    def uri(uri)
+      branch default_options.with_uri(uri)
+    end
+
     private
 
     # :nodoc:
--- a/lib/http/client.rb
+++ b/lib/http/client.rb
@@ -128,6 +128,8 @@ module HTTP
     def make_request_uri(uri, opts)
       uri = uri.to_s
 
+      uri = default_options.uri.join(uri).to_s if default_options.uri
+
       if default_options.persistent? && uri !~ HTTP_OR_HTTPS_RE
         uri = "#{default_options.persistent}#{uri}"
       end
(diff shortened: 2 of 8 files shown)


ixti:
That's somewhat duplicate of https://github.com/httprb/http/pull/493

bjeanes:
I like this interface better than mine in #493 though. Strong 👍 for merging this.

tarcieri:
#493 is definitely a different kind of API. Where it was scoped to a host (or as I suggested, `origin`) this generally operates on (potentially) merging potentially relative paths to URIs.

I think it's at least worth pointing out (not as a blocker on this PR, just as a general note) that path joining is an area that's fraught with peril, especially if the base URI is assumed to be "secure" but the joined parameter is attacker-controlled and the attacker can exploit either relative path behaviors or completely overwrite an existing path with an absolute one.

I'd personally like to see at least some tests for cases where relative and absolute paths are provided and what the behavior is, and make sure we're ok with how it behaves. I get, to a certain extent, this is testing `Addressable::URI`, but where its `URI#join` was largely an implementation detail in the past, this API puts it in the forefront, so I want to make sure we're clear on the behavior and potentially override it.

Some cases I think are important:

- URI with path component joined with absolute path: (e.g. `https://example.com/foo` and `/bar")`. Does this result in `/foo/bar` or `/bar`?
- URI with path component with absolute prefix but relative path elements (e.g. `/bar/../`)
- Paths that start with relative path elements (`../bar`)

bjeanes:
@tarcieri excellent, excellent points. _My_ goal in #493 is just about hosts but I think this API offers a bit more flexibility for other use cases. But you are 100% right that this flexibility needs to be tempered by security and usability considerations that merit further discussion.

I think we should have that discussion here before deciding to merge #493 or a derivative thereof in case there is some merit to unifying the proposals.

PR #515, opened 2018-12-27
Respect default_options for HTTP::Chainable#request and HTTP::Chainable#build_request

RickCSong:
Currently, if default_options are set, they are respected for methods such as #via or #timeout but not with a simple request.

Let me know if this would make sense, or if the current implementation is intentional!

cc @zanker HI ZANKER :)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,7 +6,7 @@ cache: bundler
 before_install:
   - gem update --system
   - gem --version
-  - gem install bundler --no-rdoc --no-ri
+  - gem install bundler
   - bundle --version
 
 install: bundle install --without development doc
--- a/lib/http/chainable.rb
+++ b/lib/http/chainable.rb
@@ -70,19 +70,15 @@ module HTTP
     end
 
     # Make an HTTP request with the given verb
-    # @param verb
-    # @param uri
-    # @option options [Hash]
-    def request(verb, uri, options = {}) # rubocop:disable Style/OptionHash
-      branch(options).request verb, uri
+    # @param (see Client#request)
+    def request(*args)
+      branch(default_options).request(*args)
     end
 
     # Prepare an HTTP request with the given verb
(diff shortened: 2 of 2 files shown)


RickCSong:
Removed some options from .travis.yml 

See error log: https://travis-ci.org/httprb/http/jobs/472558089

.travis.yml
```diff
@@ -6,7 +6,7 @@ cache: bundler
 before_install:
   - gem update --system
   - gem --version
-  - gem install bundler --no-rdoc --no-ri
+  - gem install bundler
```
ixti: Why removing those? It's better to add (or replace those with) `--no-document` in fact.

lib/http/chainable.rb
```diff
@@ -74,15 +74,15 @@ def patch(uri, options = {}) # rubocop:disable Style/OptionHash
...
+      branch(default_options.merge(options)).request verb, uri
     end
 
     # Prepare an HTTP request with the given verb
     # @param verb
     # @param uri
     # @option options [Hash]
     def build_request(verb, uri, options = {}) # rubocop:disable Style/OptionHash
-      branch(options).build_request verb, uri
+      branch(default_options.merge(options)).build_request verb, uri
```
ixti: In both cases this and above, it will be more effiecient to pass options to delegated method instead:

``` ruby
# @param (see Client#build_request)
def build_request(*args)
  branch(default_options).build_request(*args)
end
```

This way you will avoid double-merging with no good. :D

.travis.yml
ixti: After giving a thought - I guess it's fine to remove those indeed.

zanker:
👋

janko-m:
It seems that there are tests missing for this change?

ixti:
@janko-m good catch. we definitely need to do that.

Issue #511, opened 2018-11-29
Can't put cookie jar into a request


danini-the-panini:
I tried doing something like this to use the cookies from a previous response in a subsequent request:

```ruby
response = HTTP.get('http://www.example.com')
cookies = response.cookies
new_response = HTTP.get('http://www.example.com').cookies(cookies)
```

However, this does not work as expected because `response.cookies` is a `HTTP::CookieJar` and the chainable `cookies` method wants a Hash.

How do I do what I'm trying to achieve here? The documentation is sparse on this point.

tarcieri:
Reusing a `HTTP::CookieJar` (automatically) between requests was something I had originally suggested in the form of an `HTTP::Session` type: #306 

It looks like you're right though, `HTTP::Options#cookies` does not accept an `HTTP::CookieJar` as an argument, and probably should.

odinhb:
I ran into this when replacing some cURL-ing that had been done with the justification that it was easier to use cURL to maintain the session than using http.rb or other ruby http libraries. Here's the function I used:

```ruby
def cookiejar_to_hash(jar)
  hash = {}
  jar.cookies.each do |cookie|
    hash[cookie.name] = cookie.value
  end
  hash
end
```

Allowing you to
```ruby
response = HTTP.get('http://www.example.com')
cookies = cookiejar_to_hash(response.cookies)
new_response = HTTP.get('http://www.example.com').cookies(cookies)
```

HoneyryderChuck:
@odinhb there's at least [1 http ruby library supporting it](https://honeyryderchuck.gitlab.io/httpx/wiki/Cookies.html)

tarcieri:
We just landed #613 which adds support for it, but it isn't released

sferik:
`Chainable#cookies` now accepts any enumerable of `HTTP::Cookie` objects in addition to a `Hash`, so passing a cookie jar from a previous response works directly:

```ruby
response = HTTP.get("http://example.com")
HTTP.cookies(response.cookies).get("http://example.com/next")
```

This was part of a broader cookie API refactor (0dd249f3e6916e582a1d2f0c29710124a2159d28) that removed cookie management from `Options` and simplified `Chainable#cookies` to set the `Cookie` header directly from a `Hash` or `Array<HTTP::Cookie>`. Since `HTTP::CookieJar` is `Enumerable` and yields `HTTP::Cookie` objects, it should work transparently.

[...]

Issue #306, opened 2016-02-16
HTTP::Session


tarcieri:
I've looked at a lot of HTTP libraries in my day (one of the most interesting recent ones has been [proxygen](https://github.com/facebook/proxygen)) and I think the main thing this library is missing is a session abstraction:

http://docs.python-requests.org/en/master/user/advanced/#session-objects

I think this is the best way to solve #302 (cc @zanker)

janko-m:
I that the session objects might allow us to make `HTTP::Client` thread-safe (see https://github.com/socketry/async-io/issues/17). Currently I would not expect that `HTTP::Client` caches the `@connection` object without looking at it; and I still assumed that it's not, even though I'm very familiar with http.rb codebase by now. But with session objects it would be obvious that the `HTTP::Client` is (hopefully) stateless, while `HTTP::Session` uses the same connection.

janko-m:
For context, this is the example which didn't work because `HTTP::Client` is memoizing the `HTTP::Connection` object: https://github.com/socketry/async-io/issues/17

PR #517, opened 2019-01-02
Clear sensitive headers when redirecting to cross-site

### LINKED ISSUE #516: Clear sensitive headers when redirecting to cross-site
Sensitive headers (Authorization and Cookie) are sent to cross-site when redirecting.

For example, an original target URL is http://example.com and redirect to http://example.jp (cross-site). The sample code is the following.

```rb
require 'http'

HTTP.headers({Authorization: "Basic xxx", Cookie: "key=value"})
  .follow()
  .get("http://example.com")
```

This PR might be related to #264.

kyoshidajp:
Fixes #516

### DIFF
--- a/lib/http/request.rb
+++ b/lib/http/request.rb
@@ -95,13 +95,19 @@ module HTTP
     end
 
     # Returns new Request with updated uri
-    def redirect(uri, verb = @verb)
+    def redirect(uri, verb = @verb, orig_uri = nil)
       headers = self.headers.dup
       headers.delete(Headers::HOST)
 
+      uri = @uri.join(uri)
+      if !orig_uri.nil? && (uri.host != orig_uri.host)
+        headers.delete(Headers::AUTHORIZATION)
+        headers.delete(Headers::COOKIE)
+      end
+
       self.class.new(
         :verb         => verb,
-        :uri          => @uri.join(uri),
+        :uri          => uri,
         :headers      => headers,
         :proxy        => proxy,
         :body         => body.source,
--- a/lib/http/redirector.rb
+++ b/lib/http/redirector.rb
@@ -58,7 +58,7 @@ module HTTP
 
         @response.flush
 
-        @request  = redirect_to @response.headers[Headers::LOCATION]
+        @request  = redirect_to @response.headers[Headers::LOCATION], @request
         @response = yield @request
       end
 
@@ -81,7 +81,7 @@ module HTTP
 
     # Redirect policy for follow
     # @return [Request]
-    def redirect_to(uri)
+    def redirect_to(uri, orig_uri)
       raise StateError, "no Location header in redirect" unless uri
 
       verb = @request.verb
@@ -94,7 +94,7 @@ module HTTP
 
       verb = :get if !SEE_OTHER_ALLOWED_VERBS.include?(verb) && 303 == code
 
-      @request.redirect(uri, verb)
+      @request.redirect(uri, verb, orig_uri)
     end
   end
 end
--- a/spec/lib/http/redirector_spec.rb
+++ b/spec/lib/http/redirector_spec.rb
@@ -75,6 +75,41 @@ RSpec.describe HTTP::Redirector do
       expect(res.to_s).to eq "foo"
     end
 
+    context "redirect to cross-site" do
+      it "sensitive headers are not forwarded" do
+        req = HTTP::Request.new :verb => :get,
+                                :uri => "http://example.com",
+                                :headers => {"Authorization" => "Basic xxx",
+                                             "Cookie" => "key=value"}
+        res = redirect_response 300, "http://trap.example.jp"
+
+        redirector.perform(req, res) do |prev_req, _|
+          has_sensitive_headers = prev_req.headers.find do |h|
+            h.first == "Authorization" || h.first == "Cookie"
+          end
+          expect(has_sensitive_headers).to be_falsey
+          simple_response 200
+        end
+      end
+    end
+
+    context "redirect to same-site" do
+      it "sensitive headers are not forwarded" do
+        req = HTTP::Request.new :verb => :get,
+                                :uri => "http://example.com",
+                                :headers => {"Authorization" => "Basic xxx",
+                                             "Cookie" => "key=value"}
+        res = redirect_response 300, "http://example.com/redirect"
+
+        redirector.perform(req, res) do |prev_req, _|
+          has_authorization_header = prev_req.headers.find { |h| h.first == "Authorization" }
+          has_cookie_header = prev_req.headers.find { |h| h.first == "Cookie" }
+          expect(has_authorization_header && has_cookie_header).to be_truthy
+          simple_response 200
+        end
+      end
+    end
+
     context "following 300 redirect" do
       context "with strict mode" do
         let(:options) { {:strict => true} }


ixti:
Cookies should be handled in a different way. They should be filtered, and in fact I have a feeling that whole redirection thing is part of "session" alike objects, not of http client directly (as redirect can jump over multiple domains and paths, where each should be able to receive it's own cookies).

About `Authorization` header, can you point out on RFC that says that authorization header must be dropped on redirect? Probably instead of implicitly remove that header we should provide `on(:redirect)` callback that will allow to modify request.

/cc @httprb/core

tarcieri:
> I have a feeling that whole redirection thing is part of "session" alike objects, not of http client directly

Agree with this. See #306

kyoshidajp:
Thank you for your reply.

> About Authorization header, can you point out on RFC that says that authorization header must be dropped on redirect?

No. Section 6.4. "Redirection 3xx" of RFC 7231 (https://tools.ietf.org/html/rfc7231#section-6.4) doesn't explicitly explain this. But I think that client should be on the safe side.

> Probably instead of implicitly remove that header we should provide on(:redirect) callback that will allow to modify request.

You mean I should add an option to allow to modify request (default: allow) when redirecting?

HoneyryderChuck:
> They should be filtered, and in fact I have a feeling that whole redirection thing is part of "session" alike objects,

@ixti maybe so, but the redirector should be doing better here, i.e. instead of relying only on the options object cookie jar, it should take the last response cookie jar into account. And options object cookie jar usage should be filtering per request uri, which [it seems it is not doing?](https://github.com/httprb/http/blob/7c2a8613b75e79ba276c13d48d158030ac319d8d/lib/http/client.rb#L156).

tarcieri:
> ...it should take the last response cookie jar into account.

@HoneyryderChuck IMO stateful and stateless requests need different handling. What if it's the backend API client proxying different requests on behalf of individual users where the remote service is setting cookies? Now their sessions are getting mixed.

This is why it's important to have a session notion (i.e. #306) which extends beyond individual clients.

ixti:
@HoneyryderChuck 

> @ixti maybe so, but the redirector should be doing better here, i.e. instead of relying only on the options object cookie jar, it should take the last response cookie jar into account. And options object cookie jar usage should be filtering per request uri, which it seems it is not doing?.

As I said, redirector should be moved into a session object. Question of how `HTTP.cookies` API should behave is out of the scope of this question, so let's not mix it into this context. It should be discussed separately as it's actually not so trivial questions as it might see, but definitely we should improve that part.

---

@kyoshidajp 

> No. Section 6.4. "Redirection 3xx" of RFC 7231 (https://tools.ietf.org/html/rfc7231#section-6.4) doesn't explicitly explain this. But I think that client should be on the safe side.

I don't see implicit proposal on how to handle headers either. They do mention safe behaviour, but that is related to how client should behave on different HTTP verbs and responses (e.g. when POST requests gets 301 response, by default we use strict policy of raising an exception, as those are not meant to be followed without explicit user consent, thus we have `:strict => false` keyword to explicitly say that user is OK with following the redirect with changing verb from POST to GET).

> You mean I should add an option to allow to modify request (default: allow) when redirecting?

No. I'm saying that:

1. This is not responsibility of HTTP client to filter headers on redirect. I believe in all the cases where RFC does not gives explicit explanation, we should follow what `cURL` is doing (see below).

2. Dropping `Authorization` header only seems pretty weird to me. At bare minimum it should be a configurable list of headers we might want to drop (e.g. `X-Auth-Token`, `X-Api-Key`, `api_key` etc. depending on user's needs). But, it might be better to:

3. Allow user to specify modifier block for redirected request. But that will make things incredibly complicated even API-wise.

As I said before, in case of uncertainty (when RFC doe snot have clear vision on the topic) IMO we should follow what `curl`, whose manual explicitly says about `-H, --header` argument:

> **WARNING:** headers set with this option will be set in all requests - even after redirects are followed, like when told with `-L`, `--location`. This can lead to the header being  sent  to  other hosts than the original host, so sensitive headers should be used with caution combined with following redirects.

Also we can provide a keyword to instruct redirector behave similar to Firefox, which is dropping *all* headers on redirect. Or, as I said above the option can also allow specify list of headers to be stripped out upon redirect.

You can use this simple server to see how different clients behave:

``` ruby
#!/usr/bin/env ruby

require "socket"

# $ cat /etc/hosts
# ...
# 127.7.2.1 thirteen-twelve.com
# 127.7.2.1 all-colours-are-beautiful.com
# ...

server = TCPServer.new("127.7.2.1", 1312)
hosts  = %w[thirteen-twelve.com:1312 all-colours-are-beautiful.com:1312]

loop do
  socket   = server.accept
  headline = socket.gets

  STDERR.puts "> #{headline}"

  while line = socket.gets
    # stop after headers
    if line.strip.empty?
      STDERR.puts "-" * 80
      break
    end

    STDERR.puts "> #{line}"
  end

  index = headline[%r{^GET /(\S+)}, 1].to_i + 1

  if index < 3
    hostname = hosts.tap(&:rotate!).first
    socket.print "HTTP/1.1 303 See Other\r\n" \
                 "Location: http://#{hostname}/#{index}\r\n" \
                 "Content-Length: 0\r\n" \
                 "Connection: close\r\n"
  else
    socket.print "HTTP/1.1 200 OK\r\n" \
                 "Content-Length: 0\r\n" \
                 "Connection: close\r\n"
  end

  socket.close
end
```

With the above, running `curl thirteen-twelve.com:1312 -L -H "Authorization: XXX"` will show you:

```
> GET / HTTP/1.1
> Host: thirteen-twelve.com:1312
> User-Agent: curl/7.62.0
> Accept: */*
> Authorization: XXX
--------------------------------------------------------------------------------
> GET /1 HTTP/1.1
> Host: all-colours-are-beautiful.com:1312
> User-Agent: curl/7.62.0
> Accept: */*
--------------------------------------------------------------------------------
> GET /2 HTTP/1.1
> Host: thirteen-twelve.com:1312
> User-Agent: curl/7.62.0
> Accept: */*
> Authorization: XXX
```

And with Firefox you will (using edit and resend to inject headers to initial request) see:

```
> GET / HTTP/1.1
> Host: thirteen-twelve.com:1312
> User-Agent: Mozilla/5.0 (X11; Linux x86_64; rv:64.0) Gecko/20100101 Firefox/64.0
> Accept: text/html,application/xhtml+xml,application/xml;q=0.9,*/*;q=0.8
> Accept-Language: en-GB,en;q=0.5
> Accept-Encoding: gzip, deflate
> DNT: 1
> Connection: keep-alive
> Upgrade-Insecure-Requests: 1
> Pragma: no-cache
> Cache-Control: no-cache
> Authorization: XXX
> X-Api-Key: 42
> X-Auth-Token: 42
--------------------------------------------------------------------------------
> GET /1 HTTP/1.1
> Host: all-colours-are-beautiful.com:1312
> User-Agent: Mozilla/5.0 (X11; Linux x86_64; rv:64.0) Gecko/20100101 Firefox/64.0
> Accept: text/html,application/xhtml+xml,application/xml;q=0.9,*/*;q=0.8
> Accept-Language: en-GB,en;q=0.5
> Accept-Encoding: gzip, deflate
> DNT: 1
> Connection: keep-alive
> Pragma: no-cache
> Cache-Control: no-cache
--------------------------------------------------------------------------------
> GET /2 HTTP/1.1
> Host: thirteen-twelve.com:1312
> User-Agent: Mozilla/5.0 (X11; Linux x86_64; rv:64.0) Gecko/20100101 Firefox/64.0
> Accept: text/html,application/xhtml+xml,application/xml;q=0.9,*/*;q=0.8
> Accept-Language: en-GB,en;q=0.5
> Accept-Encoding: gzip, deflate
> DNT: 1
> Connection: keep-alive
> Pragma: no-cache
> Cache-Control: no-cache
```

ixti:
Bottom line - I totally agree with:

> Http::Client is already overloaded with responsibilities which makes it hard to extend/modify

In other words I would better see HTTP::Client being slimmed down to a bare minimum. No redirections, or cookies handling. The only thing that belongs to HTTP::Client from all the extras we have I believe is deflate/inflate feature :D And the rest should be extracted into a high-level client based on this one IMO.

lib/http/request.rb
```diff
@@ -95,13 +95,19 @@ def initialize(opts)
...
     # Returns new Request with updated uri
-    def redirect(uri, verb = @verb)
+    def redirect(uri, verb = @verb, orig_uri = nil)
       headers = self.headers.dup
       headers.delete(Headers::HOST)
 
+      uri = @uri.join(uri)
+      if !orig_uri.nil? && (uri.host != orig_uri.host)
+        headers.delete(Headers::AUTHORIZATION)
+        headers.delete(Headers::COOKIE)
```
tarcieri: A suggested path forward here:

Extract an `Http::Session` class, and give it a first sense of an authorization token (e.g. `auth_token`?) and cookies, then selectively decide whether or not to *add* those to the headers of particular origins. That way it avoids the need to munge these headers.

This should also make it relatively straightforward to address #511 as well.

ixti: IMO HTTP::Session should get it's behavior inspiration from user-oriented web clients (AKA Firefox and alike). And those are simply dropping ALL the headers upon any redirect (from what I observed), but naturally add Cookie header (based on host + location) of request hop.

kyoshidajp:
Sorry, it’s hard for me to do.

tarcieri:
While I think the overall idea here is good, this is not the way to implement and I would nominate closing this PR.

ixti:
Agree.

PR #532, opened 2019-02-14
Fix pipes support in bodies.

### LINKED ISSUE #531: Do not rewind unrewindable IOs
[Reported](https://github.com/httprb/http/pull/468#issuecomment-463596941) by @mikker:

---

We're using `IO.pipe` for chunked POST requests in @elastic's [APM Agent](https://github.com/elastic/apm-agent-ruby) and I'm stubbing away rewind on my reader objects because of this.

```ruby
# HTTP.rb calls #rewind the body stream which IO.pipes don't support
class ModdedIO < IO
  def self.pipe(ext_enc = nil)
    super(ext_enc).tap do |rw|
      rw[0].define_singleton_method(:rewind) { nil }
    end
  end
end
```

Just wanted to provide an actual use of IO.pipe.

ixti:
Resolves #531

### DIFF
--- a/lib/http/request/body.rb
+++ b/lib/http/request/body.rb
@@ -35,10 +35,12 @@ module HTTP
           yield @source
         elsif @source.respond_to?(:read)
           IO.copy_stream(@source, ProcIO.new(block))
-          @source.rewind if @source.respond_to?(:rewind)
+          rewind(@source)
         elsif @source.is_a?(Enumerable)
           @source.each(&block)
         end
+
+        self
       end
 
       # Request bodies are equivalent when they have the same source.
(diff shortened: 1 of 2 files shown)


ixti:
@janko I was thinking about adding the same to form_data, but not sure how to do that - as it will fail on `size` anyway ;))

janko:
@ixti We only theoretically need to do it in `HTTP::FormData::Readable#to_s`, right? From what I can see, everything else *should* be working fine. If http.rb calls `#rewind`, it will fail with the same error. `#size` will be called only if `Transfer-Encoding: chunked` is not set (which is expected, as we need `#size` otherwise). Did I miss anything?

ixti:
@janko I mean if one of parts of composable io will fail with this error - others will not be rewinded. But I guess it's an issue that should be discussed in form_data anyway, not here. :D

Issue #378, opened 2016-09-30
config URI normalization by default


sysout:
Hi,

It seems that HTTP gem do URI normalization by default and I can't find the option to change this behavior. I have an application that won't work if HTTP gem do URI normalization. For example, HTTP.get(url) will decode %2B to +
Please let me know if we can control the URI normalization or not. Thank you!

tarcieri:
I'd prefer to be able to configure this and shut it off as well. Normalization has been the cause of some past issues, e.g. #206 #246 #259 #325

ixti:
@tarcieri I agree that we should allow enable/disable normalization as well as I think it will make sense to allow use different backends for URI (Addressable::URI / ::URI), so that those who don't want to use addressable (there was an issue for that) would be able to avoid it.

mamoonraja:
@ixti @tarcieri We are running into the same issue, Is the fix coming in near feature or not?

ixti:
@mamoonraja there's no planned ETA. PRs are welcome - I'll be happy to help with that.

mamoonraja:
@ixti I am working on a PR right now, and I am approaching it in the following way:

- I am adding an option `normaliz_uri` to the request object, which will be `true` as default to keep the library backward compatible
- method `normalize_uri` will be renamed to `construct_url`, something like:

```ruby
 # @return [HTTP::URI] URI with all componentes but query being normalized.
 def construct_uri(uri, normalize_uri)
      uri = HTTP::URI.parse uri

      return uri unless normalize_uri

      HTTP::URI.new(
        :scheme     => uri.normalized_scheme,
        :authority  => uri.normalized_authority,
        :path       => uri.normalized_path,
        :query      => uri.query,
        :fragment   => uri.normalized_fragment
      )
 end
```

Do you think it's a viable solution without hurting current users?

ixti:
What if we would do this via feature instead? Like this:

``` ruby
HTTP.use(:uri_normalizer => :itself.to_proc).get(...)
```

By default normalizer will be what we have now.

mamoonraja:
I am adding a feature `uri_normalizer` and I am finishing up the pr right now. 

One more query, the way I am adding the feature is that the new feature will handle normalizing the URI, and `normalize_uri` will move out of request. Following the conventions in the library, I am handling it in `wrap_request` method inside the feature.

[...]

PR #533, opened 2019-02-21
add support for users to provide custom uri normalizer

mamoonraja:
This PR tries to tackle https://github.com/httprb/http/issues/378 while not breaking current functionality. I wanted to get the review while I spend more time cleaning up the code to answer the open questions I mentioned and add some documentation around it.

- Adds a feature `UriNormalizer`
- Provides users options to provide their own custom Uri normalizer by using `HTTP.use(:uri_normalizer => {:custom_uri_normalizer => CustomUriNormalizer.new})`
- Updated `http_spec` test to demonstrate the functionality
- By default functionality remains the same.

Open questions:
- I changed `rubocop` for now because I am adding a new condition in `request.new`, but I will work on moving that code or make client handle that.
- Instead of having `self.normalize_uri` in the request, we can just use the `DefaultUriNormalizer` in the feature itself. Probably related to the above comment.


spec/support/dummy_server/servlet.rb
```diff
@@ -148,6 +148,16 @@ def do_#{method.upcase}(req, res)
...
 
+    get "/uri_normalizer/ABC" do |_req, res|
+      res.status = 200
+      res.body   = "default normalizer"
+    end
+
+    get "/uri_normalizer/custom" do |_req, res|
+      res.status = 200
+      res.body   = "custom normalizer"
+    end
```
ixti: just add one endpoint for testing. As simple as:

``` ruby
get "/hello;world" do |_req, res|
  res.status = 200
  res.body   = "hello world"
end
```

Default normalizer will hit 404 in this case, and custom should get "hello world".

lib/http/request.rb
```diff
@@ -196,6 +200,19 @@ def inspect
       "#<#{self.class}/#{@version} #{verb.to_s.upcase} #{uri}>"
     end
 
+    # @return [HTTP::URI] URI with all components but query being normalized.
+    def self.normalize_uri(uri)
```
ixti: There's no need in duplicating default normalizer - just assign `URI::NORMALIZER` constant with lambda somethign like:

``` ruby
module HTTP
  class URI < Addressable::URI
    NORMALIZER = lambda do |uri|
      HTTP::URI.new(
        :scheme     => uri.normalized_scheme,
        # ...
      )
    end

File: lib/http/features/uri_normalizer.rb [hunk 3]
```diff
@@ -0,0 +1,32 @@
+# frozen_string_literal: true
+
+module HTTP
+  module Features
+    class UriNormalizer < Feature
+      attr_reader :custom_uri_normalizer
+
+      def initialize(custom_uri_normalizer: DefaultUriNormalizer.new)
```
ixti: make argument name `normalizer`, *custom_uri_* prefix is very redundant.

ixti:
So, I think API should look like this:

``` ruby
HTTP.use(:normalize_uri => { :normalizer => normalizer })
```

Also, normalizer object should be ANYTHING that responds to `#call` method:

``` ruby
module NormalizeMeNot
  def self.call(uri); uri; end
end

HTTP.use(:normalize_uri => { :normalizer => NormalizeMeNot })

wtf = ->(uri) { HTTP::URI.parse "https://example.com" }
HTTP.use(:normalize_uri => { :normalizer => wtf })
```

To make things easier, we can support extra options in feature, that will be passed to normalizer as second argument.

``` ruby
DEFAULT_NORMALIZER = lambda do |uri, only: [:schema, :query...]|
  URI.new(
    :schema => only.include?(:schame) ? uri.normalized_schema : uri.schema,
    # ...
```

lib/http/request.rb
```diff
@@ -66,20 +66,24 @@ class UnsupportedSchemeError < RequestError; end
     # Scheme is normalized to be a lowercase symbol e.g. :http, :https
     attr_reader :scheme
 
+    attr_reader :uri_normalizer
```
ixti: No need to expose it.

lib/http/request.rb
```diff
@@ -66,20 +66,24 @@ class UnsupportedSchemeError < RequestError; end
...
+
     # "Request URI" as per RFC 2616
     # http://www.w3.org/Protocols/rfc2616/rfc2616-sec5.html
     attr_reader :uri
     attr_reader :proxy, :body, :version
 
     # @option opts [String] :version
     # @option opts [#to_s] :verb HTTP request method
+    # @option opts [#to_s] :uri_normalizer uri normalizer
```
ixti: description is redundant. It's obvious from the key name.

spec/lib/http_spec.rb
```diff
@@ -430,6 +430,31 @@ def setsockopt(*args)
...
     end
+
+    context "with :normalize_uri" do
+      module Normalizer
+        def self.call(uri)
+          uri
+        end
+      end
+
+      it "Use the defaul Uri normalizer when user does not use uri normalizer" do
```
ixti: URI should be all caps. And topic of example should start with lowercase. Also typo in *defaul*.
In this particular example should be moved into default context, either here:

https://github.com/httprb/http/blob/1eae1557f1fb68408190345cb731fc32007754bd/spec/lib/http_spec.rb#L21

or here:

https://github.com/httprb/http/blob/1eae1557f1fb68408190345cb731fc32007754bd/spec/lib/http_spec.rb#L466

And should look like:

``` ruby
it "normalizes URI" do
  response = HTTP.get "#{dummy.endpoint}/hello world"
  expect(response.to_s).to eq("hello world")
end
```

others examples are OK to be left in here, just make them less "wordy"...

spec/lib/http_spec.rb
```diff
@@ -430,6 +430,31 @@ def setsockopt(*args)
         expect(response.to_s).to eq("#{body}-deflated")
       end
     end
+
+    context "with :normalize_uri" do
+      module Normalizer
+        def self.call(uri)
+          uri
+        end
+      end
```
ixti: Don't declare module/class dynamically. You basically don't need this at all.

spec/lib/http_spec.rb
```diff
@@ -430,6 +430,31 @@ def setsockopt(*args)
...
+
+      it "Use the defaul Uri normalizer when user does not use uri normalizer" do
+        response = HTTP.get "#{dummy.endpoint}/hello world"
+        expect(response.to_s).to eq("hello world")
+      end
+
+      it "Use the custom Uri Normalizer method" do
+        client = HTTP.use(:normalize_uri => {:normalizer => Normalizer})
```
ixti: As I said above you don't need `Normalizer` module here. Replace this line with:

```suggestion
        client = HTTP.use(:normalize_uri => {:normalizer => :itself.to_proc})
```

spec/lib/http_spec.rb
```diff
@@ -430,6 +430,31 @@ def setsockopt(*args)
...
+        response = client.get("#{dummy.endpoint}/hello world")
+        expect(response.status).to eq(400)
+      end
+
+      it "Use the default Uri normalizer when user does not specify custom uri normalizer" do
+        client = HTTP.use :normalize_uri
```
ixti: Worth to add expectation that default normalizer will be called in here:

``` ruby
expect(HTTP::URI::NORMALIZER).to receive(:call).and_call_original
```

lib/http/request.rb
```diff
@@ -66,20 +66,24 @@ class UnsupportedSchemeError < RequestError; end
...
     # @option opts [HTTP::URI, #to_s] :uri
     # @option opts [Hash] :headers
     # @option opts [Hash] :proxy
     # @option opts [String, Enumerable, IO, nil] :body
     def initialize(opts)
       @verb   = opts.fetch(:verb).to_s.downcase.to_sym
-      @uri    = normalize_uri(opts.fetch(:uri))
+      @uri_normalizer = opts[:normalize_uri] ? opts.fetch(:normalize_uri) : HTTP::URI::NORMALIZER
+      @uri = @uri_normalizer.call(opts.fetch(:uri))
```
ixti: No need to assign normalizer here if it won't be used elsewhere. Also use fetch instead of ternary here:

``` ruby
normalize = opts.fetch(:normalize_uri, HTTP::URI::NORMALIZER)
@uri = normalize.call(opts.fetch(:uri))
```

Probably worth to verify if returned value is an HTTP::URI, as we might depend on it's API. Not sure on this count though.

ixti:
@httprb/core Would like to know your opinions on whenever we should enforce custom URI normalizers to return `HTTP::URI` or just assume that user knows what he's doing and returns anything that might work later down the road (if we will depend on some of HTTP::URI methods, other implementations might be lack of).

On one hand I think we should enforce, as it will allow to avoid bug reports due to crazy setup. On the other hand, I guess one of the best things this PR can bring on is optional avoidance of HTTP::Addressable (which has been reported as slow - although I'm not buying this argument).

digitalextremist:
@ixti looking at #378 I agree with the claim that normalization ought not be the default behavior; or I'd add that at least that there ought to be a trustworthy method call which overrides normalization, not just requiring specialized configuration every time if you want to turn it off.

ixti:
@digitalextremist Mm, that's actually doable with this PR as well:

``` ruby
module MyNamespace
  HTTP = ::HTTP.use(:normalize_url => :itself.to_proc)
end
```

But I'm totally fine if we will have some sort of global defaults configuration as well, although that is a bit out of the scope of this PR IMO. Something like (just a crazy idea):

``` ruby
HTTP.defaults.uri_normalizer = ->(_) { HTTP::URI.parse "https://example.com" }
```

lib/http/request.rb
mamoonraja: I am exposing it because we are using it in `wrap_request` method in `auto_deflate`

mamoonraja:
@ixti I addressed the comments, and I agree that adding the global defaults configuration is out of scope for this PR.

ixti:
@mamoonraja thank you. I'll take a look closer to the weekend.

mamoonraja:
@ixti Thanks! Do you have an ETA on when this will be merged?

ixti:
@mamoonraja Sure. Sorry had a busy weeks. Will merge this weekend for sure.

mamoonraja:
No problem! thanks a lot!

ixti:
Merged in as https://github.com/httprb/http/commit/7fd135648f7d48b6a246d303ce9e0cc5aabfba57 and released as v4.1.0

PR #548, opened 2019-05-05
RuboCop upgrade

natesholland:
I am trying to get more involved in open source code maintenance. In order to do this I figured I would try upgrading RuboCop for a few gems I use frequently to help them stay up to date. RuboCop on this project was very out of date so I started by upgrading it by ten minor versions to `0.59.2` but I can go further if the maintainers want.

I broke this PR into different commits based on what work I was doing. The first commit is the update and auto-fixes. The commits after that are the commits where I made judgement calls on what to do.

Please let me know if there's anything I can do to help or improve.

---
This change is [Reviewable](https://reviewable.io/reviews/httprb/http/548)

### DIFF
--- a/lib/http/options.rb
+++ b/lib/http/options.rb
@@ -1,14 +1,12 @@
 # frozen_string_literal: true
 
-# rubocop:disable Metrics/ClassLength
-
 require "http/headers"
 require "openssl"
 require "socket"
 require "http/uri"
 
 module HTTP
-  class Options
+  class Options # rubocop:disable Metrics/ClassLength
     @default_socket_class     = TCPSocket
     @default_ssl_socket_class = OpenSSL::SSL::SSLSocket
     @default_timeout_class    = HTTP::Timeout::Null
--- a/lib/http/options.rb
+++ b/lib/http/options.rb
@@ -18,8 +18,9 @@ module HTTP
       attr_accessor :default_socket_class, :default_ssl_socket_class, :default_timeout_class
       attr_reader :available_features
 
-      def new(options = {}) # rubocop:disable Style/OptionHash
+      def new(options = {})
         return options if options.is_a?(self)
+
         super
       end
 
--- a/lib/http/options.rb
+++ b/lib/http/options.rb
@@ -17,9 +17,7 @@ module HTTP
       attr_reader :available_features
 
       def new(options = {})
-        return options if options.is_a?(self)
-
-        super
+        options.is_a?(self) ? options : super
       end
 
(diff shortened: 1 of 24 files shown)


spec/support/black_hole.rb
```diff
@@ -2,9 +2,11 @@
 
 module BlackHole
   class << self
+    # rubocop:disable Style/MethodMissingSuper
```
mikegee: You should be able to move this down as a suffix of the `def` and remove the `enable` comment afterwards:

```ruby
def method_missing(*) # rubocop:disable Style/MethodMissingSuper
```

lib/http/options.rb
```diff
@@ -196,3 +196,4 @@ def argument_error!(message)
     end
   end
 end
+# rubocop:enable Metrics/ClassLength
```
mikegee: Maybe you can move the `disable` comment down to the line that opens the class and we won't need this? For instance:

```ruby
module HTTP
  class Options # rubocop:disable Metrics/ClassLength
```

natesholland:
Thanks for the pointer @mikegee, I've made the recommended changes.

mikegee:
Awesome!

I didn't comment on `spec/support/dummy_server/servlet.rb` earlier, but I think the same recommendation applies there too.

natesholland:
@mikegee addressed.

As a side note, I would like to get #546 merged before mine if possible. I really want that change in place and don't want my RuboCop changes to conflict with it.

natesholland:
@ixti I am new to the Reviewable tool but I believe I have addressed all of your feedback. If you want me to revert out the Style/EvalWithLocation changes in fb6d66e please let me know and I'll do that.

ixti:
@natesholland Honestly I just wanted to try our reviewable ;))

Re Style/EvalWithLocation - I'm totally fine with the requirement of file/line. I don't like `+ 1` part only. Although I see where does it comes from, so I guess let's keep it as you did it.

ixti:
> As a side note, I would like to get #546 merged before mine if possible. I really want that change in place and don't want my RuboCop changes to conflict with it.

That PR needs more work to be done. I will get to it as soon as I will have time for that. Thus merging this prior that change.

natesholland:
@ixti if I have time, would you like me to upgrade RuboCop again in the future? The version I bumped to now is much more recent, but still a ways behind the latest version.

ixti:
@natesholland Sure, feel free to do this. I'm all for keeping rubocop to the latest.

PR #549, opened 2019-05-08
RuboCop upgrade 0.68 (latest)

natesholland:
In #548 I upgraded RuboCop by about 10 versions. In this PR I'm following up to upgrade to the latest version of RuboCop. All the changes in this PR are a result of the auto-correct flags so I didn't edit any code manually.

There are two specific items that I want to call out in this PR:
- First is that `rubocop-performance` is now its own gem that can be found here https://github.com/rubocop-hq/rubocop-performance. I suspect that we still want the performance cops to keep this project as performant as possible. Therefore I pulled the gem in and required the performance cops to keep them running. I tested to ensure that we are in fact still running performance cops.
- For Layout/AlignHash I had to make a call on which style to prefer. I personally like the table formatting, but I'm happy to change it to a different format based on what the maintainers want. The documentation for the three format options are here https://www.rubydoc.info/gems/rubocop/RuboCop/Cop/Layout/AlignHash. I like `table` and then `key` and then I think `separator is clearly the weirdest and worst.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -1,3 +1,5 @@
+require: rubocop-performance
+
 AllCops:
   TargetRubyVersion: 2.3
   DisplayCopNames: true
@@ -12,6 +14,11 @@ Layout/SpaceAroundOperators:
 
 Layout/SpaceInsideHashLiteralBraces:
   EnforcedStyle: no_space
+
+Layout/AlignHash:
+  Enabled: false
+  EnforcedColonStyle: table
+  EnforcedHashRocketStyle: table
 
 ## Metrics #####################################################################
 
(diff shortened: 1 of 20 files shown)


.rubocop.yml
```diff
@@ -13,6 +15,10 @@ Layout/SpaceAroundOperators:
 Layout/SpaceInsideHashLiteralBraces:
   EnforcedStyle: no_space
 
+Layout/AlignHash:
```
ixti: Please put this at the before `Layout/DotPosition` (so that they are alphabetically sorted.

natesholland: Done and rebased.

natesholland:
@ixti done and ready to go.

ixti:
Thank you!

Issue #547, opened 2019-04-30
Allow Features to handle ConnectionErrors


joshuaflanagan:
An obvious use case for the `Feature` extension point is for logging/instrumenting (https://github.com/httprb/http/pull/499). This allows you to log request details, and the corresponding response. However, if the connection fails (timeout, DNS lookup failure, etc), you only get a chance to log the request - never the response. Someone looking at the logs will not know what happened to the request.

I propose adding the `Feature#on_error` hook, so that logging/instrumenting extensions get a chance to log errors. It could be added in a backward compatible way, so that all existing `Feature` implementations still work.

Something like this:

```
    rescue ConnectionError => e
      options.features.each do |_name, feature|
        if feature.respond_to?(:on_error)
          feature.on_error(req, e)
        end
      end
      raise
```

The feature `on_error` callback gets access to the `Request` and the `ConnectionError` (or it could be a broader error).

I can put together a PR if we come to agreement on the behavior.

ixti:
I'm not sure I understand why existing code won't work. You can log request before it has been sent. Not sure what benefits `on_error` will provide.

joshuaflanagan:
Currently, you can configure an instance using the Logging feature, and wherever you use it, you will see a Request logged, and then the corresponding Response. The callsites dont have to know about the logging, it happens transparently by the `Feature`.
Now, if you make a request that doesn't succeed because of a connection failure, you will log the request, but your logs wont show the result of that request - because there was no response. In order to log the connection failure, you will have to directly instrument at every callsite.

You would have to do something like this, for _every_ call:

```
begin
  http.get("http://www.example.com")
rescue ConnectionError => e
  <same logger configured by HTTP Logging feature>.info("ConnectionError: #{e}"
end
```

Alternatively, you can add the `on_error` handler, and the `Logging` feature could log that error for you - without the callsite having to handle it.

ixti:
I still don't think it's a good idea to introduce the change that will be useful only for logging feature.

joshuaflanagan:
I got the impression that the `Feature` API was designed specifically for logging/instrumentation, and then made slightly more generic to allow some other middleware support. Currently, the only other `Features` are inflate/deflate. `NormalizeUri` is a strange one - I'm not sure why that is a `Feature`.
So it seems like tying the logging/instrumentation implementation to the generic `Feature` API has limited its usefulness. Would you be more open to adding first-class logging support directly in the `Client`?

paul:
[...]

PR #552, opened 2019-05-20
Allow Features to handle connection errors

joshuaflanagan:
This addresses https://github.com/httprb/http/issues/547

It allows a Feature to implement an on_error handler so that it has a chance to react to Requests that never get a Response.

### DIFF
--- a/lib/http/client.rb
+++ b/lib/http/client.rb
@@ -70,20 +70,18 @@ module HTTP
 
       @connection ||= HTTP::Connection.new(req, options)
 
-      unless @connection.failed_proxy_connect?
-        @connection.send_request(req)
-        @connection.read_headers!
+      begin
+        unless @connection.failed_proxy_connect?
+          @connection.send_request(req)
+          @connection.read_headers!
+        end
+      rescue Error => e
+        options.features.each_value do |feature|
+          feature.on_error(req, e)
+        end
+        raise
       end
-
-      res = Response.new(
-        :status        => @connection.status_code,
-        :version       => @connection.http_version,
-        :headers       => @connection.headers,
-        :proxy_headers => @connection.proxy_response_headers,
-        :connection    => @connection,
-        :encoding      => options.encoding,
-        :request       => req
-      )
+      res = build_response(req, options)
 
       res = options.features.inject(res) do |response, (_name, feature)|
         feature.wrap_response(response)
@@ -105,6 +103,18 @@ module HTTP
     end
 
     private
+
+    def build_response(req, options)
+      Response.new(
+        :status        => @connection.status_code,
+        :version       => @connection.http_version,
+        :headers       => @connection.headers,
+        :proxy_headers => @connection.proxy_response_headers,
+        :connection    => @connection,
+        :encoding      => options.encoding,
+        :request       => req
+      )
+    end
 
     # Verify our request isn't going to be made against another URI
     def verify_connection!(uri)
--- a/lib/http/feature.rb
+++ b/lib/http/feature.rb
@@ -13,6 +13,8 @@ module HTTP
     def wrap_response(response)
       response
     end
+
+    def on_error(request, error); end
   end
 end
 
--- a/spec/lib/http/client_spec.rb
+++ b/spec/lib/http/client_spec.rb
@@ -266,6 +266,61 @@ RSpec.describe HTTP::Client do
         end
       end
     end
+
+    context "Feature" do
+      let(:feature_class) do
+        Class.new(HTTP::Feature) do
+          attr_reader :captured_request, :captured_response, :captured_error
+
+          def wrap_request(request)
+            @captured_request = request
+          end
+
+          def wrap_response(response)
+            @captured_response = response
+          end
+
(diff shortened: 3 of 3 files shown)


tarcieri:
Looks like you need to fix the RuboCop errors. Otherwise conceptually this seems ok to me.

@ixti what do you think?

tarcieri:
@paul any thoughts on this approach, or suggested alternatives?

paul:
@tarcieri No, this seems fine to me. :+1:

lib/http/client.rb
```diff
@@ -70,9 +70,16 @@ def perform(req, options)
...
-      unless @connection.failed_proxy_connect?
-        @connection.send_request(req)
-        @connection.read_headers!
+      begin
+        unless @connection.failed_proxy_connect?
+          @connection.send_request(req)
+          @connection.read_headers!
+        end
+      rescue Error => e
+        options.features.each do |_name, feature|
```
tarcieri: I think you can do `each_value` here (and potentially collapse it into a 1-liner), e.g.:

`options.features.each_value { |feature| feature.on_error(req, e) }`

joshuaflanagan:
> Looks like you need to fix the RuboCop errors. Otherwise conceptually this seems ok to me.

It is difficult as an outside contributor to understand which rubocop rules a team cares about. Were each of the rules chosen because they are what the team values?  Or were they just the defaults?

For example, this project currently has the `Metrics/MethodLength` rule enabled, so this PR has this violation:

```
lib/http/client.rb:66:5: C: Metrics/MethodLength: Method has too many lines. [32/25]
    def perform(req, options)
```

Is it really appropriate for this PR to take on the task of refactoring the `Client#perform` method so that it isn't so long?

Or should I disable the rule? Or just exclude this file? Or just exclude this method?

tarcieri:
> It is difficult as an outside contributor to understand which rubocop rules a team cares about. Were each of the rules chosen because they are what the team values? Or were they just the defaults?

It's explicitly configured to 25 lines in `.rubocop.yml`. To answer your question: yes this project values smaller methods.

> Is it really appropriate for this PR to take on the task of refactoring the Client#perform method so that it isn't so long?

If you really don't want to refactor it you can disable the check for this method in particular using `rubocop:disable` and leave it for someone else to refactor. At the very least the RuboCop check needs to pass.

That said it looks like `Response.new` could benefit from some refactoring and is the source of much of the methods bulk by lines.

joshuaflanagan:
> If you really don't want to refactor it you can disable the check for this method in particular using `rubocop:disable` and leave it for someone else to refactor. At the very least the RuboCop check needs to pass.

I don't mind doing the refactor, I just thought it was out of scope for this task, and would obscure the point of this PR. I'll go ahead and disable the rule here, and leave it as a follow up issue.

ixti:
What if error will happen down the course, e.g. during response streaming?

tarcieri:
@ixti it seems like under this approach features wouldn't be able to handle errors, however it seems like they can handle the most interesting errors

joshuaflanagan:
I have rebased and addressed the rubocop issues.

ixti:
Thank you!

I have taken another look, and I think we can make API a bit more "flexible" if we will rename `on_error` to `observe`:

``` ruby
def observe(event, data); end

# ...

options.features.each_value { |f| f.observe(:request_error, :req => self, :err => e) }
```

Just an idea. As I don't like `on_error` name it's too vague, while used in very specific case only. Also, probably we should add observers in general instead of bloating this functionality into features.

tarcieri:
@ixti what about `on_event` that takes a symbol (e.g. `:error`) as an argument?

joshuaflanagan:
> @ixti what about `on_event` that takes a symbol (e.g. `:error`) as an argument?

I think that could be a good idea _if_ you know of other events that would make sense to observe.
I could imagine having `:request` and `:response` events, which could replace the need to "wrap" those objects for all use cases. I understand the wrapping approach is required for some features, but it leads to an awkward implementation for things like logging, which don't need to modify the request/response.

ixti:
Yeah. I think `on_event` makes sense to me.

And I agree, that `:request` and `:response` events will make notifications (which are actually observers) code look much better. Also after giving it another look, I guess I'm not very against `on_error` name any more, we can create a bunch of `on_{event}` methods:


``` ruby
def on_request(req:); end
def on_request_error(req:, err:); end
def on_response(res:); end
def on_response_error(res:, err:); end
```

joshuaflanagan:
Can we move forward with this PR and leave the `on_request/on_response` changes for a follow-up effort?
I'm not exactly sure how you want to differentiate between `on_request_error` and `on_response_error`, but willing to make the changes with some direction.

tarcieri:
@joshuaflanagan that sounds fine to me

joshuaflanagan:
Done

tarcieri:
@joshuaflanagan this one now has a conflict

tarcieri:
@ixti let me know if you have any qualms about landing this. We can potentially explore alternatives before cutting another release

PR #559, opened 2019-08-05
Remove unnecessary files from gemspec for smaller binary

trusche:
Hi! This PR reduces the file size of the gem binary from 76K to 62K by removing all those files from the gem that are only required for development and testing. Those are not really necessary in the packaged gem. I've left in README, CHANGES and LICENSE just in case - it can help for quick inspections via `bundle open` or `gem open` IMO. 

Please feel free to ignore this PR if the dev files were included intentionally, I just assume they weren't.

### DIFF
--- a/http.gemspec
+++ b/http.gemspec
@@ -18,11 +18,14 @@ Gem::Specification.new do |gem|
   gem.homepage      = "https://github.com/httprb/http"
   gem.licenses      = ["MIT"]
 
-  gem.executables   = `git ls-files -- bin/*`.split("\n").map { |f| File.basename(f) }
-  gem.files         = Dir["lib/**/*.rb"] +
-                      %w[http.gemspec README.md CHANGES.md LICENSE.txt]
+  Dir.chdir(__dir__) do
+    git_files = `git ls-files -z`.split("\x0")
 
-  gem.test_files    = `git ls-files -- {test,spec,features}/*`.split("\n")
+    gem.files = git_files.grep(%r{^lib/}) +
+                %w[http.gemspec README.md CHANGES.md LICENSE.txt]
+    gem.test_files = git_files.grep(%r{^spec/})
+  end
+
   gem.name          = "http"
   gem.require_paths = ["lib"]
   gem.version       = HTTP::VERSION
--- a/http.gemspec
+++ b/http.gemspec
@@ -19,7 +19,9 @@ Gem::Specification.new do |gem|
   gem.licenses      = ["MIT"]
 
   gem.executables   = `git ls-files -- bin/*`.split("\n").map { |f| File.basename(f) }
-  gem.files         = `git ls-files`.split("\n")
+  gem.files         = Dir["lib/**/*.rb"] +
+                      %w[http.gemspec README.md CHANGES.md LICENSE.txt]
+
   gem.test_files    = `git ls-files -- {test,spec,features}/*`.split("\n")
   gem.name          = "http"
   gem.require_paths = ["lib"]


http.gemspec
```diff
@@ -19,7 +19,9 @@ Gem::Specification.new do |gem|
   gem.licenses      = ["MIT"]
 
   gem.executables   = `git ls-files -- bin/*`.split("\n").map { |f| File.basename(f) }
-  gem.files         = `git ls-files`.split("\n")
+  gem.files         = Dir["lib/**/*.rb"] +
+                      %w[http.gemspec README.md CHANGES.md LICENSE.txt]
```
ixti: IMO it's better to keep git ls-files + filter instead. And we can update git ls-files command a bit:

``` ruby
  gem.files         = Dir.chdir(__dir__) do
    lib_files = `git ls-files -z`.split("\x0").
      select { |f| f.start_with?("lib/") }

    lib_files + %w[http.gemspec README.md CHANGES.md LICENSE.txt]
  end
```

http.gemspec
```diff
@@ -19,7 +19,9 @@ Gem::Specification.new do |gem|
...
+
   gem.test_files    = `git ls-files -- {test,spec,features}/*`.split("\n")
```
ixti: In fact it worth to group these two:

``` ruby
Dir.chdir(__dir__) do
  git_files = `git ls-files -z`.split("\x0")

  gem.files = git_files.grep(%r{^lib/}) +
    %w[http.gemspec README.md CHANGES.md LICENSE.txt]

  gem.test_files = git_files.grep(%r{^spec/})
end
```

http.gemspec
```diff
@@ -19,7 +19,9 @@ Gem::Specification.new do |gem|
   gem.licenses      = ["MIT"]
 
   gem.executables   = `git ls-files -- bin/*`.split("\n").map { |f| File.basename(f) }
```
ixti: and it worth to remove this one.

http.gemspec
trusche: But that would include the test files in the binary again, no? At least it does when I use this with `gem build http.gemspec`... not that it makes much difference to the size, I guess just removing `logo.png` at ~8KB from the binary makes all the difference that matters.

tarcieri: Distro maintainers who repackage RubyGems as distro-specific packages (e.g. deb, rpm) prefer the tests are included so they can run them to ensure they properly repackaged the code.

trusche:
That's good to know, thanks. Here's the improved version. Result is still 62KB, here's the file list for sanity checking (no idea how to test this unfortunately):

```
18:53 $ ls -lh http-5.0.0.pre.gem
-rw-r--r--  1 trusche  staff    62K  5 Aug 18:51 http-5.0.0.pre.gem
✔ ~/devel/private/http [master|✚ 1…77]
18:53 $ tree http-5.0.0.pre
http-5.0.0.pre
├── CHANGES.md
├── LICENSE.txt
├── README.md
├── http.gemspec
├── lib
│   ├── http
│   │   ├── chainable.rb
│   │   ├── client.rb
│   │   ├── connection.rb
│   │   ├── content_type.rb
│   │   ├── errors.rb
│   │   ├── feature.rb
│   │   ├── features
│   │   │   ├── auto_deflate.rb
│   │   │   ├── auto_inflate.rb
│   │   │   ├── instrumentation.rb
│   │   │   ├── logging.rb
│   │   │   └── normalize_uri.rb
│   │   ├── headers
│   │   │   ├── known.rb
│   │   │   └── mixin.rb
│   │   ├── headers.rb
│   │   ├── mime_type
│   │   │   ├── adapter.rb
│   │   │   └── json.rb
│   │   ├── mime_type.rb
│   │   ├── options.rb
│   │   ├── redirector.rb
│   │   ├── request
│   │   │   ├── body.rb
│   │   │   └── writer.rb
│   │   ├── request.rb
│   │   ├── response
│   │   │   ├── body.rb
│   │   │   ├── inflater.rb
│   │   │   ├── parser.rb
│   │   │   ├── status
│   │   │   │   └── reasons.rb
│   │   │   └── status.rb
│   │   ├── response.rb
│   │   ├── timeout
│   │   │   ├── global.rb
│   │   │   ├── null.rb
│   │   │   └── per_operation.rb
│   │   ├── uri.rb
│   │   └── version.rb
│   └── http.rb
└── spec
    ├── lib
    │   ├── http
    │   │   ├── client_spec.rb
    │   │   ├── connection_spec.rb
    │   │   ├── content_type_spec.rb
    │   │   ├── features
    │   │   │   ├── auto_deflate_spec.rb
    │   │   │   ├── auto_inflate_spec.rb
    │   │   │   ├── instrumentation_spec.rb
    │   │   │   └── logging_spec.rb
    │   │   ├── headers
    │   │   │   └── mixin_spec.rb
    │   │   ├── headers_spec.rb
    │   │   ├── options
    │   │   │   ├── body_spec.rb
    │   │   │   ├── features_spec.rb
    │   │   │   ├── form_spec.rb
    │   │   │   ├── headers_spec.rb
    │   │   │   ├── json_spec.rb
    │   │   │   ├── merge_spec.rb
    │   │   │   ├── new_spec.rb
    │   │   │   └── proxy_spec.rb
    │   │   ├── options_spec.rb
    │   │   ├── redirector_spec.rb
    │   │   ├── request
    │   │   │   ├── body_spec.rb
    │   │   │   └── writer_spec.rb
    │   │   ├── request_spec.rb
    │   │   ├── response
    │   │   │   ├── body_spec.rb
    │   │   │   └── status_spec.rb
    │   │   ├── response_spec.rb
    │   │   └── uri_spec.rb
    │   └── http_spec.rb
    ├── regression_specs.rb
    ├── spec_helper.rb
    └── support
        ├── black_hole.rb
        ├── capture_warning.rb
        ├── dummy_server
        │   └── servlet.rb
        ├── dummy_server.rb
        ├── fakeio.rb
        ├── http_handling_shared.rb
        ├── proxy_server.rb
        ├── servers
        │   ├── config.rb
        │   └── runner.rb
        └── ssl_helper.rb
```

ixti:
It worth to ensure at least `Gemfile` inclusion to `test_files` to make life of maintainers easier.

mikegee:
I think the `test_files` config is deprecated and the `gem test` command is not available in the later versions.

ixti:
@mikegee thanks for the info I did not knew.

I would propose to remove that setting then and move spec/ files into files. I guess I would be more happy to completely remove test files, but quick looking into Gentoo packages showed that they use gem from rubygems (rather than using github repo) for the package. And it has support for running tests before installing it in the system, so we should keep that ability.

tarcieri:
Closing this PR as stale.

Not really enthusiastic about this approach. If anything I'd like to see the file selection simplified using a glob, like `Dir["lib/**/*", "CHANGES.md", "LICENSE", "README.md"]` instead.

PR #569, opened 2019-10-21
Use Request#uri in Response

ixti:
Response initialize requires Request object now, making uri parsing absolutely redundant. And as it's a breaking change anyway - we should get rid of `:uri` option completely.

NOTE: We need to update WebMock's adapter to reflect this change prior releasing v5.0.0

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -17,6 +17,11 @@
   Fix HTTP parser.
   ([@ixti], [@fxposter])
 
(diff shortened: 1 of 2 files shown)


joshuaflanagan:
FYI - this is the commit to our fork of webmock that we've been using along with our fork of httprb https://github.com/ShippingEasy/webmock/commit/505ad0098d40e6da28cea374ac78d6daec8c9daa

Issue #524, opened 2019-02-02
Non capitalized HTTP headers


ninoseki:
Sorry for a dumb question.

I have to work with an API server which deals HTTP headers in case-sensitive manner.
(Yep I know that is a RFC violation)

Is there a good way to send non-capitalized HTTP headers?

tarcieri:
@ixti neat! SGTM 👍

brasic:
Hi!  Just came across this issue while checking to see if anyone had run into the problem we are currently dealing with.  An API we need to communicate with requires a particular header be specified with an underscore in the header name.  It's otherwise case-insensitive.  HTTP.rb makes it impossible to pass a header in an underscore because of this line, which transforms it to `-`: https://github.com/httprb/http/blob/e31d0a534cd39726ca363e1ab2e9d8e314d61581/lib/http/headers.rb#L206

I would argue that the behavior of `HTTP::Headers#normalize_header` violates [RFC 7230](http://tools.ietf.org/html/rfc7230#section-3.2), which allows the `_` character to be part of a header name.  We can monkeypatch this for now on our side but are there any thoughts on making the canonicalization algorithm spec-compliant?  I don't know of any case-sensitivity scheme that considers `_` and `-` to be equivalent.  I'd be happy to open a PR to remove `_` from the split regex but that would likely require a major version bump (looking at the specs, lots of code depends on `_` canonicalizing as `-`).

ixti:
@brasic yes - I'm working on refactoring HTTP::Headers completely so that it will llow to pass any RFC compliant header (with or without normalization).

Hendrione-Moka:
any update for this? I still got the issue.

ixti:
I had no time to work on this yet.

joshuaflanagan:
Just ran into this again, dealing with an API that has case-sensitive headers. That may be "wrong" on their side, but it is what it is, and switching http libraries is much easier than getting an external API to  change their behavior.
Do you have an in-process branch? I'd be happy to take a stab at fleshing it out. Would rather do that than switch libraries.

tarcieri:
@joshuaflanagan I think if you implemented the change @ixti suggested here it'd be accepted, and shouldn't be too difficult: https://github.com/httprb/http/issues/524#issuecomment-460014322

PR #576, opened 2019-12-15
Preserve header casing as specified

joshuaflanagan:
The original behavior was to normalize all header names so that they
were broken up into words, delimited by `-` or `_`, capitalize each word,
and then join the words together with a `-`.

This made it impossible to make a request with an underscore (`_`) in the header
name, or with a different casing (ex: all caps).

However, the normalized name made it possible to access (or delete) headers,
without having to know the exact casing.

The new behavior is based on the following rules (as specified in https://github.com/httprb/http/issues/524#issuecomment-460014322)

1) Fail if a header name is not specified as a String or Symbol
2) If the header name is specified as a Symbol, normalize it when writing it in a request.
If the header name is specified as a String, preserve it as-is when writing it in a request.
3) Allow lookup of any header using the normalized form of the name

I implemented this behavior by storing three elements for each header value:
1) normalized header name
2) header name as it will be written in a request
3) header value

Element 2 is the new addition. I considered just storing the header value
as it would be written, and only doing normalization during lookup, but
it seemed wasteful to potentially normalize the same value over and over
when searching through the list for various lookups. This way we only
normalize each name once, and can continue to use that value for lookups.
However, whenever asked for the contents (ex: via `each` or `keys`) we
return the new, non-normalized name.

Fixes: https://github.com/httprb/http/issues/524

### DIFF
--- a/lib/http/headers.rb
+++ b/lib/http/headers.rb
@@ -21,6 +21,10 @@ module HTTP
 
     # Class constructor.
     def initialize
+      # The @pile stores each header value using a three element array:
+      #  0 - the normalized header key, used for lookup
+      #  1 - the header key as it will be sent with a request
+      #  2 - the value
       @pile = []
     end
 
@@ -45,12 +49,31 @@ module HTTP
 
     # Appends header.
     #
-    # @param [#to_s] name header name
+    # @param [String, Symbol] name header name. When specified as a string, the
+    #   name is sent as-is. When specified as a symbol, the name is converted
+    #   to a string of capitalized words separated by a dash. Word boundaries
+    #   are determined by an underscore (`_`) or a dash (`-`).
+    #   Ex: `:content_type` is sent as `"Content-Type"`, and `"auth_key"` (string)
+    #   is sent as `"auth_key"`.
     # @param [Array<#to_s>, #to_s] value header value(s) to be appended
     # @return [void]
     def add(name, value)
-      name = normalize_header name.to_s
-      Array(value).each { |v| @pile << [name, validate_value(v)] }
+      lookup_name = normalize_header(name.to_s)
+      wire_name = case name
+                  when String
+                    name
+                  when Symbol
+                    lookup_name
+                  else
+                    raise HTTP::HeaderError, "HTTP header must be a String or Symbol: #{name.inspect}"
+                  end
+      Array(value).each do |v|
+        @pile << [
+          lookup_name,
+          wire_name,
+          validate_value(v)
+        ]
+      end
     end
 
     # Returns list of header values if any.
@@ -58,7 +81,7 @@ module HTTP
     # @return [Array<String>]
     def get(name)
       name = normalize_header name.to_s
-      @pile.select { |k, _| k == name }.map { |_, v| v }
+      @pile.select { |k, _| k == name }.map { |_, _, v| v }
     end
 
     # Smart version of {#get}.
@@ -96,7 +119,7 @@ module HTTP
     #
     # @return [Array<[String, String]>]
     def to_a
-      @pile.map { |pair| pair.map(&:dup) }
+      @pile.map { |item| item[1..2] }
     end
 
     # Returns human-readable representation of `self` instance.
@@ -110,7 +133,7 @@ module HTTP
     #
     # @return [Array<String>]
     def keys
-      @pile.map { |k, _| k }.uniq
+      @pile.map { |_, k, _| k }.uniq
     end
 
     # Compares headers to another Headers or Array of key/value pairs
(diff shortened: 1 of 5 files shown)


lib/http/headers.rb
```diff
@@ -49,16 +53,30 @@ def delete(name)
     # @param [Array<#to_s>, #to_s] value header value(s) to be appended
     # @return [void]
     def add(name, value)
```
ixti: Would be nice to introduce optional keyword:

``` ruby
def add(name, value, normalize: nil)
```

Naturally, when normalize is false, no normalization should be done for the wire. When true - force normalization, and when nil - type dependent normalization.

lib/http/headers.rb
```diff
@@ -49,16 +53,30 @@ def delete(name)
...
-      name = normalize_header name.to_s
-      Array(value).each { |v| @pile << [name, validate_value(v)] }
+      lookup_name = normalize_header(name.to_s)
+      wire_name = case name
+                  when String
+                    name
+                  when Symbol
+                    lookup_name
+                  else
+                    raise HTTP::HeaderError, "HTTP header must be a String or Symbol: #{name.inspect}"
```
ixti: I think it should raise only if key is no ot responding to to_s and not a Symbol. Something  like:

``` ruby
if name.is_a? Symbol
  # ...
elsif name.respond_to? :to_s
  # ...
else
  raise
end
```

ixti:
Huge thanks for the PR. I'm not against breaking api, but would like add signature to allow predictable behaviour when normalize dwarf passed

joshuaflanagan:
Thanks for the feedback. I don’t completely understand what you want to enable. Can you write me a test that would fail with the existing code, that you want to pass?

I think the most “predictable” behavior is to _not_ do any manipulation of the strings provided by the user. I do understand that is a breaking change if someone is currently specifying a “content_type” and expecting it to go out on the wire as “Content-Type”. But for the vast number of new users and new code being written, I think changing the name out from under them is more surprising.

joshuaflanagan:
I understand the normalize keyword, but not exactly why it is helpful. In one place you suggest the default is nil (current type based switching), but then another place you suggest it should be true. That suggests the default behavior would be to continue changing a specified name into something else on the wire. That seems like more surprised for future users.

lib/http/headers.rb
joshuaflanagan: I don't see how that is helpful. Almost everything in ruby responds to `to_s` (unless it derives from `BasicObject` to explicitly opt out of default behavior).

If you want to allow more than just `String` (not sure what the use case is), we could support anything that responds to `to_str`, which is used to support implicit conversion to strings.

ixti:
@joshuaflanagan honestly I'm not sure what I want. :D I want the API to be as predictable as possible. Please give me a bit of time to think. Would like to play around with your PR first.

lib/http/headers.rb
ixti: Ok. I agree. Then we need to update api doc to reflect this change.

joshuaflanagan:
Any further thoughts on this? Is the primary concern that it includes a breaking change?

I think the only breaking change is to existing code that adds a request header with a name, specified as a String, that includes an underscore. With the new code, that will go out on the wire as an underscore (`_`), whereas previous code would have (incorrectly) converted it to a dash (`-`).

Technically, it _could_ also be considered a breaking change that the casing is no longer changed from what the user specified. Header names are not supposed to be case-sensitive, so servers _shouldn't_ care that header names are no longer being converted to "capitalized" case. However, we know that not all servers are RFC compliant and _do_ care about case (which is, in fact, the motivation behind this PR). So it is possible that this change in casing will break some existing code, if the server someone is communicating with expects a specific casing _and_ the user is not currently passing the correct casing.

ixti:
After giving it more thoughts I guess I'm fine to merge it. But please, fix the API doc of `Headers#add`.

lib/http/headers.rb
joshuaflanagan: Do you want to want to keep my current strict implementation (only allow String or Symbol), or do you want to use the more relaxed "anything that supports an implicit conversion to string (ie. responds to `#to_str`)"?

I think being strict makes sense here, especially since we have different behavior based on the data type.

joshuaflanagan:
I've updated the API doc to reflect the current strict implementation that only accepts a string or symbol as the header name.

lib/http/headers.rb
ixti: I'm fine with your way. Strict Symbol or String.

uberllama:
Hey folks, we ran into this today with a partner using non-standard api headers. I noticed it still hasn't been released. Is there an ETA?

tarcieri:
See the release tracking milestone here: https://github.com/httprb/http/milestone/11

There are some showstopper bugs which would be good to resolve before cutting another release. We would appreciate help from anyone interested in getting it over the line.

Linuus:
For anyone reading this later on. This is a **breaking change**. See: https://github.com/httprb/http/issues/700

PR #577, opened 2019-12-15
Duplicate options Hash in .timeout call

antonvolkoff:
When `Hash` is passed to HTTP.timeout, it's been mutated. So when you pass
a frozen `Hash` it raises an exception.

Here is an example of the code:
```
timeout_settings = { connect: 5, write: 2, read: 10 }.freeze
HTTP.timeout(timeout_settings).get('http://example.com')
```

In this change, I have added a call to .dup on options Hash to create a
copy and avoid mutation of the original Hash.

### DIFF
--- a/spec/lib/http_spec.rb
+++ b/spec/lib/http_spec.rb
@@ -307,6 +307,15 @@ RSpec.describe HTTP do
       end
     end
 
+    context "specifying per operation timeouts as frozen hash" do
+      let(:frozen_options) { {:read => 123}.freeze }
+      subject(:client) { HTTP.timeout(frozen_options) }
+
+      it "does not raise an error" do
+        expect { client }.not_to raise_error
+      end
+    end
+
     context "specifying a global timeout" do
       subject(:client) { HTTP.timeout 123 }
 
--- a/lib/http/chainable.rb
+++ b/lib/http/chainable.rb
@@ -93,7 +93,7 @@ module HTTP
     def timeout(options)
       klass, options = case options
                        when Numeric then [HTTP::Timeout::Global, {:global => options}]
-                       when Hash    then [HTTP::Timeout::PerOperation, options]
+                       when Hash    then [HTTP::Timeout::PerOperation, options.dup]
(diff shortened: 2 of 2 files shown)


tarcieri:
Looks like it's failing due to the nonexistence of `FrozenError` on 2.4

antonvolkoff:
@tarcieri Thank you for a quick answer. There are few ways to fix the spec. I would like to run it by you to choose a proper way to handle this issue.

1) Check for version:
```ruby
let(:expected_error_klass) do
  Gem::Version.new(RUBY_VERSION) < Gem::Version.new("2.5.0") ? RuntimeError : FrozenError
end

it '...' do
  expect { client }.not_to raise_error(expected_error_klass)
end
```

2) Just check for `RuntimeError` which will work for both cases
```ruby
# because
FrozenError.superclass #=> RuntimeError

# I can just
expect { client }.not_to raise_error(RuntimeError)
```

tarcieri:
@antonvolkoff how about:

```
# TODO: remove this when support for Ruby 2.4 is dropped
skip if Gem::Version.new(RUBY_VERSION) < Gem::Version.new("2.5.0")
```

It's EOL in March 2020 so I'd say it's ok to ignore this particular test

spec/lib/http_spec.rb
```diff
@@ -307,6 +307,15 @@
       end
     end
 
+    context "specifying per operation timeouts as frozen hash" do
+      let(:frozen_options) { {:read => 123}.freeze }
+      subject(:client) { HTTP.timeout(frozen_options) }
+
+      it "does not raise an error" do
+        expect { client }.not_to raise_error(FrozenError)
```
ixti: Please don't specify error class. This code simply *should not raise*. It's generally recommended to avoid specify error class on `not_to raise_error`.

```suggestion
        expect { client }.not_to raise_error
```

antonvolkoff: Done 👍 and rebased

ixti:
Travis now being really weird. o_O Gonna force merge and figure out why later.

PR #578, opened 2019-12-16
Introduce proxy auto-detection based on ENV variables

### LINKED ISSUE #356: http_proxy and https_proxy (and probably no_proxy) not supported out of the box
I don't know if this was by design, but it took me some figuring out on how to explicitly set proxy. 

However, for portability, I'd automatically set proxy from the environment (the `http{,s}_proxy` env var), just like `net/http` does. 

It's not that hard, as URI#find_proxy is all you need to accomplish this. This also has the benefit of supporting the `no_proxy` env var.

brodock:
Auto-detection is based on standard ENV variables used in a UNIX system.
Some documentation can be seen here:
https://www.gnu.org/software/wget/manual/html_node/Proxies.html

The most common and used are "#{schema}_proxy" and no_proxy, but it also
supports other non-recommended formats covered by Ruby URI library.

Fix #356

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,7 +4,7 @@ sudo: false
 cache: bundler
 
 before_install:
-  - gem update --system
+  - yes | gem update --system
   - gem --version
   - gem install bundler
   - bundle --version
--- a/lib/http/request/proxy.rb
+++ b/lib/http/request/proxy.rb
@@ -1,0 +1,69 @@
+# frozen_string_literal: true
+
+module HTTP
+  class Request
+    class Proxy
+      def initialize(proxy_params, request_uri)
+        @proxy = proxy_params || find_proxy(request_uri)
+      end
+
+      # @return [Boolean] Is there any proxy configuration available?
+      def available?
+        @proxy && @proxy.include?(:proxy_address) && @proxy.include?(:proxy_port)
+      end
+
+      # @return [Boolean] Does the proxy include authentication credentials?
+      def authenticated?
+        @proxy && @proxy.include?(:proxy_username) && @proxy.include?(:proxy_password)
+      end
+
+      # @return [Boolean] Does the proxy include additional request headers?
+      def include_headers?
+        @proxy.key?(:proxy_headers)
+      end
+
+      # @return [String, nil] Username (authentication credential)
+      def username
+        @proxy[:proxy_username]
+      end
+
+      # @return [String, nil] Password (authentication credential)
+      def password
+        @proxy[:proxy_password]
+      end
+
+      # @return [String, nil] Address
+      def address
+        @proxy[:proxy_address]
+      end
+
+      # @return [String, Integer, nil] Port
+      def port
+        @proxy[:proxy_port]
+      end
+
+      # @return [Hash, nil] Additional headers
+      def headers
+        @proxy[:proxy_headers]
+      end
+
+      private
+
+      def find_proxy(request_uri)
+        system_proxy = find_system_proxy(request_uri)
+
+        proxy = {}
+        proxy[:proxy_username] = system_proxy.user if system_proxy&.user
+        proxy[:proxy_password] = system_proxy.password if system_proxy&.password
+        proxy[:proxy_address] = system_proxy.host if system_proxy&.host
+        proxy[:proxy_port] = system_proxy.port if system_proxy&.port
+
+        proxy
+      end
+
+      def find_system_proxy(request_uri)
+        ::URI.parse(request_uri.to_s).find_proxy
+      end
+    end
+  end
+end
--- a/lib/http/request/proxy.rb
+++ b/lib/http/request/proxy.rb
@@ -3,8 +3,24 @@
 module HTTP
   class Request
     class Proxy
-      def initialize(proxy_params, request_uri)
-        @proxy = proxy_params || find_proxy(request_uri)
+      # Try to find proxy attributes set via ENV variables matching against request_uri
+      # @see ::URI::Generic#find_proxy
+      # @param [HTTP::URI] request_uri
+      # @return [HTTP::Request::Proxy]
+      def self.auto_detect(request_uri)
+        system_proxy = ::URI.parse(request_uri.to_s).find_proxy
+
+        proxy = {}
+        proxy[:proxy_username] = system_proxy.user if system_proxy&.user
+        proxy[:proxy_password] = system_proxy.password if system_proxy&.password
+        proxy[:proxy_address] = system_proxy.host if system_proxy&.host
+        proxy[:proxy_port] = system_proxy.port if system_proxy&.port
+
+        new(proxy)
+      end
+
+      def initialize(proxy_params)
+        @proxy = proxy_params || {}
       end
 
       # @return [Boolean] Is there any proxy configuration available?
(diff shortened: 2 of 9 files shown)


lib/http/request/proxy.rb
```diff
@@ -0,0 +1,65 @@
+module HTTP
+  class Request
+    class Proxy
+      def initialize(proxy_params, request_uri)
+        @proxy = proxy_params || find_proxy(request_uri)
```
brodock: There are some caveats here that may need a decision:

- When it's user defined, it will ignore ENV defined values (maybe what we want is system to always override this?)
- If we want to define custom headers, it can't use any value from ENV

Should we try instead to merge everything? (So we can keep proxy defined by ENV but still set extra headers?)
What should be the precedence order? ENV always win?

.travis.yml
```diff
@@ -4,7 +4,7 @@ sudo: false
 cache: bundler
 
 before_install:
-  - gem update --system
+  - rvm rubygems latest
```
ixti: How does this fixes Travis? o_O Not saying it's not, just trying to understand why.

lib/http/request/proxy.rb
ixti: We definitely don't want ENV vars to take precedence over explicitly given proxy prams. At its best it will be unexpected, but overall it will be a disaster if you will need different proxies in one cases and some default in others.

lib/http/request/proxy.rb
```diff
@@ -0,0 +1,69 @@
...
+        proxy[:proxy_username] = system_proxy.user if system_proxy&.user
+        proxy[:proxy_password] = system_proxy.password if system_proxy&.password
+        proxy[:proxy_address] = system_proxy.host if system_proxy&.host
+        proxy[:proxy_port] = system_proxy.port if system_proxy&.port
+
+        proxy
+      end
+
+      def find_system_proxy(request_uri)
+        URI(request_uri.to_s).find_proxy
```
ixti: URI here, after full load of code will refer `HTTP::URI` which is `Addressable::URI` and don't have this method.

ixti:
I'm not against idea of having default proxy values in general, but I also don't really think it's a good to put this into this gems defaults. Probably it would be better to have some sort of global config instead

``` ruby
# Default: nil
HTTP.default_proxy = nil

# If you want to respect ENV vars as default proxies, add to you bootstrap code:
HTTP.default_proxy = lambda { |req| ::URI(req.uri.to_s).find_proxy }
```

There's still no "agreement" on what these env vars should be. Some use `/https?_proxy/`, some `/HTTPS?_PROXY/`, and some `/HTTPS?_PROXY/i`. And personally I find this suuuuuper frustrating. Firefox, FWIW does not use these env vars anymore.


@tarcieri would like to know your input.

tarcieri:
Per the discussion on #361, what @ixti said was where we netted out before: we'd like to support this (see also aforementioned #356), but it should be opt in.

See @ixti's suggestion of a `via_default_proxy` method here:

https://github.com/httprb/http/pull/361#issuecomment-234998621

lib/http/request/proxy.rb
brodock: I can see this in two ways (and no one being wrong):

1. System administrator (root) should be able to control how anything running on the machine should interact with the rest of the network, therefore, if ENV is set, it takes precedence
1. Developer should have total control on its own code, so if proxy is defined, it takes precedence

If we are inclined to favor the developer, I can suggest we do the following instead:

1. if at least one of the following are filled by the user/developer, ignore ENV: `proxy_address`, `proxy_port`, `proxy_username`, `proxy_password`
1. always merge `proxy_headers` (so we preserve it when using values provided by ENV)

tarcieri: @brodock see my comments here: https://github.com/httprb/http/pull/578#issuecomment-566072228

This isn't a dichotomy: many environments require both Internet access via an egress proxy, and S2S communication which doesn't go through the egress proxy.

Changing any kind of defaults is a semver breaking change that could have impacts on real-world environments where this library is currently deployed (I say this rather empirically, as this library is heavily utilized in a large enterprise company where I used to work).

.travis.yml
brodock: The issue here can be seen : https://travis-ci.org/httprb/http/jobs/625515097

```
$ gem update --system
Updating rubygems-update
Fetching rubygems-update-3.1.0.gem
Successfully installed rubygems-update-3.1.0
Installing RubyGems 3.1.0
  Successfully built RubyGem
  Name: bundler
  Version: 2.1.0.pre.3
  File: bundler-2.1.0.pre.3.gem
bundler's executable "bundle" conflicts with /home/travis/.rvm/rubies/ruby-2.4.6/bin/bundle
Overwrite the executable? [yN]  
No output has been received i
```

which is related to : https://github.com/rubygems/rubygems/issues/3030

Now looking more carefully, it seems that the change didn't had the effect I wanted, but worked because it basically did a no-op here.

brodock: There seems to be a better workaround here: https://github.com/rubygems/rubygems/issues/3036 (I will amend the commit)

lib/http/request/proxy.rb
brodock: Interesting enough it didn't break the specs... If I change to `::URI` it should work, right?

brodock:
I can explain a little bit better how it should work...
There is a lack of canonical documentation on the env variables, but the two best sources are: https://www.gnu.org/software/wget/manual/html_node/Proxies.html and https://github.com/ruby/ruby/blob/7909f06212ae8df6ba7203f8152292a190b2b33a/lib/uri/generic.rb#L1475-L1490

I tried to cover most common scenarios on our specs

lib/http/request/proxy.rb
brodock: I wasn't aware of a use-case where you would have different proxy configurations like that.

So direction is to implement this new behaviour in `via_default_proxy`? 
As it is a chainable method, what should happen if you have `HTTP.via(...).via_default_proxy...` ?

tarcieri: I believe the behavior for that sort of thing elsewhere is to overwrite the previous configuration (or rather, split/dup it and then set the new value)

brodock:
Also if it helps I can split the travis fix in its own PR, so we can fix master

lib/http/request/proxy.rb
brodock: So in my example above, assuming:

```ruby
ENV["http_proxy"] = "http://proxy1.example.com:9090"
```

```ruby
HTTP.via("proxy2.example.com", 8080).via_default_proxy()...
# proxy_address: proxy1.example.com
# proxy_port: 9090
```

but keeping credentials from `via`, wouldn't make sense, and we should unset those if default detects anything (as the credentials there are for the first proxy not the second)

```ruby
HTTP.via("proxy2.example.com", 8080, "douglas", "adams").via_default_proxy()...
```

and the third case would be to preserve the headers:

```ruby
HTTP.via("proxy2.example.com", 8080, "douglas", "adams", {"X-Custom-Header" => "some value"}).via_default_proxy()...
```

brodock: I tried to implement `via_default_proxy` with previously proposed semantics and it's not possible as we need the request_uri in advance which is not garanteed to be available during chainable definition.

I tried an alternative: `with_system_proxy`, this defines an extra proxy_param: `proxy_auto_detection: true`, which gets evaluated when building the request.

It will first try to check if user defined proxy is "available?", if that fails, it wil check whether auto_detection is set and try to use it, falling back to unset proxy when that also fails.

(still need to cover auto detection flag with specs)

tarcieri:
Closing this PR as stale.

It seems like it was going in the right direction and was close to ready once upon a time. Please rebase and reopen if you'd like to pursue it.

PR #583, opened 2020-01-09
Add digest auth feature

Silex:
This is a decent start (it works) but there's a few problems to be solved:
- Should this be a feature or a modification of HTTP::Client?
- How do I access the HTTP client that was used to make the request? Right now I used `HTTP` but we need to reuse the original one, with user options, etc.

### DIFF
--- a/lib/http/features/digest.rb
+++ b/lib/http/features/digest.rb
@@ -1,0 +1,20 @@
+# frozen_string_literal: true
+
+require "net/http/digest_auth"
+
+module HTTP
+  module Features
+    class Digest < Feature
+      def wrap_response(response)
+        if response.status.code == 401
+          digest = Net::HTTP::DigestAuth.new
+          auth = digest.auth_header(response.request.uri, response["WWW-Authenticate"], response.request.verb.to_s.upcase)
+          response = HTTP.auth(auth).get(response.request.uri)
+        end
+        response
+      end
+
+      HTTP::Options.register_feature(:digest, self)
+    end
+  end
+end


ixti:
I don't think this modification belongs to HTTP gem on its own. This feature, should be a some sort of higher-level client on top of http client itself. For this reason I'm planning to remove (extract to sessions) redirects follower as well.

Silex:
What can/should I do?

Almost all my equipments (HTTP cameras) use digest auth. I didn't find a clean way to write a reusable `digest_auth` wrapper method, because I need to know about the HTTP instance, the user, the pass, and the get/post/patch request and that *looks* ugly.

ixti:
As I said, I don't see this particular feature to be part of the HTTP core client. In your case I believe you need to simply create client class for your needs:

``` ruby
class CameraClient
  def initialize(username:, password:, http: HTTP)
    @username = username
    @password = password
    @http = http
  end

  %i[get post head].each do |verb|
    define_method(verb) { |*args, **options| request(verb, *args, **options) }
  end

  def request(verb, uri, **options)
    # your custom logic of request/response and authentication here
  end
end
```

paul:
You can also register your own features, if you want to keep your implementation:

https://github.com/httprb/http/blob/master/lib/http/options.rb#L27

```
HTTP.register_feature(:digest, MyDigestAuth)

HTTP.use(digest: {username: "AzureDiamond", password: "hunter2"}).get("http://test.example/")
```

Silex:
Thanks for the examples! :+1:

Silex:
For information, this is how what I ended up with:

```
require 'net/http/digest_auth'

module HTTP
  module Features
    class Digest < Feature
      def initialize(options = {})
        @http = options[:http] || HTTP
      end

      def wrap_response(response)
        if response.status.code == 401
          digest = Net::HTTP::DigestAuth.new
          auth = digest.auth_header(response.request.uri, response['WWW-Authenticate'], response.request.verb.to_s.upcase)
          response = @http.auth(auth).get(response.request.uri)
        end
        response
      end

      HTTP::Options.register_feature(:digest, self)
    end
  end
end

http = HTTP.timeout(5)
http = HTTP.use(digest: { http: http })
http.get(image_url).to_s
```

It's a bit ugly but it works. When I have time I'll write some DigestHTTP module that wraps HTTP, that or simply extend HTTP with a Digest module, or maybe only HTTP::Client.

tarcieri:
@Silex if we ever get around to implementing #306, an `HTTP::Session` type seems like it might be a reasonable place to put it as well.

I think what @ixti was getting at is `HTTP::Client` is presently overloaded with functionality.

Silex:
@tarcieri: right, I agree.

ixti:
Yup, that's  what I meant. I believe that HTTP client should e responsible for  a single request. No state. HTTP session  on the other hand, should be capable of redirects, digest auths, cookies handling etc.

PR #584, opened 2020-01-09
Bump http-form_data dependency to 2.2+

janko:
`http-form_data` 2.2.0 addresses Ruby 2.7 compatibility (see https://github.com/httprb/form_data/pull/28), so I think it makes sense to ensure `http` requires at least that version.

This should be the last thing needed for Ruby 2.7 compatibility.


ixti:
Oh! I'm very sorry, I did not seen your PR before I made my srcew-up commits to master:

- https://github.com/httprb/http/commit/164aad0a8a3735d88d1e31ff2a44dbd02059c1ab
- followed by https://github.com/httprb/http/commit/968e4f481fa851565c7c4f8c3db120738e15c2fc

PR #588, opened 2020-01-15
Added documentation.

marlinpierce:
Added documentation for calling parameters I had to research.

---
This change is [Reviewable](https://reviewable.io/reviews/httprb/http/588)

### DIFF
--- a/lib/http.rb
+++ b/lib/http.rb
@@ -19,13 +19,12 @@ require "http/response/parser"
 #
 # Basic Usage
 #
-# HTTP.post("https://server.com", options)
+#     HTTP.post("https://server.com", options)
+#     => #<HTTP::Response ...>
 #
-# For more detailed documentation and usage notes:
-# @see https://github.com/httprb/http/wiki the http wiki
+# @see https://github.com/httprb/http/wiki wiki
 #
 # Methods in HTTP class object generally are implemented in the Chainable module.
-#
 module HTTP
   extend Chainable
 
--- a/lib/http.rb
+++ b/lib/http.rb
@@ -15,7 +15,17 @@ require "http/response"
 require "http/response/body"
 require "http/response/parser"
 
-# HTTP should be easy
+# HTTP should be easy.
+#
+# Basic Usage
+#
+# HTTP.post("https://server.com", options)
+#
+# For more detailed documentation and usage notes:
+# @see https://github.com/httprb/http/wiki the http wiki
+#
+# Methods in HTTP class object generally are implemented in the Chainable module.
+#
 module HTTP
   extend Chainable
 
--- a/lib/http/chainable.rb
+++ b/lib/http/chainable.rb
@@ -6,42 +6,48 @@ require "http/headers"
 
 module HTTP
   module Chainable
-    # Request a get sans response body
+    # Request a get sans response body.
+    # See Client#request.
     # @param uri
     # @option options [Hash]
     def head(uri, options = {}) # rubocop:disable Style/OptionHash
       request :head, uri, options
     end
 
-    # Get a resource
+    # Get a resource.
+    # See Client#request.
     # @param uri
     # @option options [Hash]
     def get(uri, options = {}) # rubocop:disable Style/OptionHash
       request :get, uri, options
     end
 
-    # Post to a resource
+    # Post to a resource.
+    # See Client#request.
     # @param uri
     # @option options [Hash]
     def post(uri, options = {}) # rubocop:disable Style/OptionHash
       request :post, uri, options
     end
 
-    # Put to a resource
+    # Put to a resource.
+    # See Client#request.
     # @param uri
     # @option options [Hash]
     def put(uri, options = {}) # rubocop:disable Style/OptionHash
       request :put, uri, options
     end
 
-    # Delete a resource
+    # Delete a resource.
+    # See Client#request.
     # @param uri
     # @option options [Hash]
     def delete(uri, options = {}) # rubocop:disable Style/OptionHash
       request :delete, uri, options
     end
 
-    # Echo the request back to the client
+    # Echo the request back to the client.
+    # See Client#request.
     # @param uri
     # @option options [Hash]
     def trace(uri, options = {}) # rubocop:disable Style/OptionHash
@@ -62,15 +68,16 @@ module HTTP
       request :connect, uri, options
     end
 
-    # Apply partial modifications to a resource
+    # Apply partial modifications to a resource.
+    # See Client#request.
     # @param uri
     # @option options [Hash]
     def patch(uri, options = {}) # rubocop:disable Style/OptionHash
       request :patch, uri, options
     end
 
-    # Make an HTTP request with the given verb
-    # @param (see Client#request)
+    # Make an HTTP request with the given verb.
+    # See Client#request.
     def request(*args)
       branch(default_options).request(*args)
     end
@@ -224,7 +231,10 @@ module HTTP
     end
 
     # Set options for HTTP
-    # @param opts
+    #
+    # See HTTP::Options.defined_options for list of options.
+    #
+    # @param [Hash] opts
     # @return [HTTP::Options]
     def default_options=(opts)
       @default_options = HTTP::Options.new(opts)
--- a/lib/http/chainable.rb
+++ b/lib/http/chainable.rb
@@ -7,7 +7,7 @@ require "http/headers"
 module HTTP
   module Chainable
     # Request a get sans response body.
-    # See Client#request.
+    # @see Client#request
     # @param uri
     # @option options [Hash]
     def head(uri, options = {}) # rubocop:disable Style/OptionHash
@@ -15,7 +15,7 @@ module HTTP
     end
 
     # Get a resource.
-    # See Client#request.
+    # @see Client#request
     # @param uri
     # @option options [Hash]
     def get(uri, options = {}) # rubocop:disable Style/OptionHash
@@ -23,7 +23,7 @@ module HTTP
     end
 
     # Post to a resource.
-    # See Client#request.
+    # @see Client#request
     # @param uri
     # @option options [Hash]
     def post(uri, options = {}) # rubocop:disable Style/OptionHash
@@ -31,7 +31,7 @@ module HTTP
     end
 
     # Put to a resource.
-    # See Client#request.
+    # @see Client#request
     # @param uri
     # @option options [Hash]
     def put(uri, options = {}) # rubocop:disable Style/OptionHash
@@ -39,7 +39,7 @@ module HTTP
     end
 
     # Delete a resource.
-    # See Client#request.
+    # @see Client#request
     # @param uri
     # @option options [Hash]
     def delete(uri, options = {}) # rubocop:disable Style/OptionHash
@@ -47,7 +47,7 @@ module HTTP
     end
 
     # Echo the request back to the client.
-    # See Client#request.
+    # @see Client#request
     # @param uri
     # @option options [Hash]
     def trace(uri, options = {}) # rubocop:disable Style/OptionHash
@@ -69,7 +69,7 @@ module HTTP
     end
 
     # Apply partial modifications to a resource.
-    # See Client#request.
+    # @see Client#request
     # @param uri
     # @option options [Hash]
     def patch(uri, options = {}) # rubocop:disable Style/OptionHash
@@ -77,7 +77,7 @@ module HTTP
     end
 
     # Make an HTTP request with the given verb.
-    # See Client#request.
+    # @see Client#request
     def request(*args)
       branch(default_options).request(*args)
     end
@@ -231,9 +231,7 @@ module HTTP
     end
 
     # Set options for HTTP
-    #
-    # See HTTP::Options.defined_options for list of options.
-    #
+    # @see Options
     # @param [Hash] opts
     # @return [HTTP::Options]
     def default_options=(opts)
--- a/lib/http/client.rb
+++ b/lib/http/client.rb
@@ -24,7 +24,12 @@ module HTTP
       @state = :clean
     end
 
-    # Make an HTTP request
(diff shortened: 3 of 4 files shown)


lib/http.rb
```diff
@@ -15,7 +15,17 @@
...
+# HTTP should be easy.
+#
+# Basic Usage
+#
+# HTTP.post("https://server.com", options)
+#
+# For more detailed documentation and usage notes:
+# @see https://github.com/httprb/http/wiki the http wiki
+#
+# Methods in HTTP class object generally are implemented in the Chainable module.
```
ixti: I believe it's redundant by different reasons:

1. it's trying to expose implementation details
2. YARD will tell that this module is extended with Chainable module and thus has ... methods
3. It's pretty obvious from the code that one should look into Chainable for methods

lib/http.rb
```diff
@@ -15,7 +15,17 @@
...
 
-# HTTP should be easy
+# HTTP should be easy.
+#
+# Basic Usage
+#
+# HTTP.post("https://server.com", options)
+#
+# For more detailed documentation and usage notes:
+# @see https://github.com/httprb/http/wiki the http wiki
```
ixti:
```suggestion
# @see https://github.com/httprb/http/wiki wiki
```

lib/http.rb
```diff
@@ -15,7 +15,17 @@
...
 require "http/response/parser"
 
-# HTTP should be easy
+# HTTP should be easy.
+#
+# Basic Usage
+#
+# HTTP.post("https://server.com", options)
+#
+# For more detailed documentation and usage notes:
```
ixti: I think this line is redundant a bit and with `@see` tag below it will be rendered a bit bad..
```suggestion
```

lib/http.rb
```diff
@@ -15,7 +15,17 @@
...
+# @see https://github.com/httprb/http/wiki the http wiki
+#
+# Methods in HTTP class object generally are implemented in the Chainable module.
+#
```
ixti: Please avoid extra empty comment line.
```suggestion
```

lib/http.rb
```diff
@@ -15,7 +15,17 @@
 require "http/response/body"
 require "http/response/parser"
 
-# HTTP should be easy
+# HTTP should be easy.
+#
+# Basic Usage
+#
+# HTTP.post("https://server.com", options)
```
ixti: Please indent this code to make it code block in rendered docuentation:

```suggestion
#     HTTP.post("https://server.com")
#     # => #<HTTP::Response ...>
```

lib/http/chainable.rb
```diff
@@ -6,42 +6,48 @@
 
 module HTTP
   module Chainable
-    # Request a get sans response body
+    # Request a get sans response body.
+    # See Client#request.
```
ixti: Please use `@see` tag

```suggestion
    # @see Client#request
```

lib/http/chainable.rb
```diff
@@ -6,42 +6,48 @@
...
     # @param uri
     # @option options [Hash]
     def head(uri, options = {}) # rubocop:disable Style/OptionHash
       request :head, uri, options
     end
 
-    # Get a resource
+    # Get a resource.
+    # See Client#request.
```
ixti: Please use `@see` tag

```suggestion
    # @see Client#request
```

lib/http/chainable.rb
```diff
@@ -6,42 +6,48 @@
...
     # @param uri
     # @option options [Hash]
     def get(uri, options = {}) # rubocop:disable Style/OptionHash
       request :get, uri, options
     end
 
-    # Post to a resource
+    # Post to a resource.
+    # See Client#request.
```
ixti: Please use `@see` tag

```suggestion
    # @see Client#request
```

lib/http/chainable.rb
```diff
@@ -6,42 +6,48 @@
...
     # @param uri
     # @option options [Hash]
     def post(uri, options = {}) # rubocop:disable Style/OptionHash
       request :post, uri, options
     end
 
-    # Put to a resource
+    # Put to a resource.
+    # See Client#request.
```
ixti: Please use `@see` tag

```suggestion
    # @see Client#request
```

lib/http/chainable.rb
```diff
@@ -6,42 +6,48 @@
...
     # @param uri
     # @option options [Hash]
     def put(uri, options = {}) # rubocop:disable Style/OptionHash
       request :put, uri, options
     end
 
-    # Delete a resource
+    # Delete a resource.
+    # See Client#request.
```
ixti: Please use `@see` tag

```suggestion
    # @see Client#request
```

lib/http/chainable.rb
```diff
@@ -6,42 +6,48 @@
...
     # @param uri
     # @option options [Hash]
     def delete(uri, options = {}) # rubocop:disable Style/OptionHash
       request :delete, uri, options
     end
 
-    # Echo the request back to the client
+    # Echo the request back to the client.
+    # See Client#request.
```
ixti: Please use `@see` tag

```suggestion
    # @see Client#request
```

lib/http/chainable.rb
```diff
@@ -62,15 +68,16 @@ def connect(uri, options = {}) # rubocop:disable Style/OptionHash
       request :connect, uri, options
     end
 
-    # Apply partial modifications to a resource
+    # Apply partial modifications to a resource.
+    # See Client#request.
```
ixti: Please use `@see` tag

```suggestion
    # @see Client#request
```

lib/http/chainable.rb
```diff
@@ -62,15 +68,16 @@ def connect(uri, options = {}) # rubocop:disable Style/OptionHash
...
     # @param uri
     # @option options [Hash]
     def patch(uri, options = {}) # rubocop:disable Style/OptionHash
       request :patch, uri, options
     end
 
-    # Make an HTTP request with the given verb
-    # @param (see Client#request)
+    # Make an HTTP request with the given verb.
+    # See Client#request.
```
ixti: Please use `@see` tag

```suggestion
    # @see Client#request
```

lib/http/chainable.rb
```diff
@@ -224,7 +231,10 @@ def default_options
     end
 
     # Set options for HTTP
-    # @param opts
+    #
+    # See HTTP::Options.defined_options for list of options.
```
ixti: Please use `@see` tag.

``` ruby
# @see Options
```

Notice that `.defined_options` are implementation details and should not be referenced in documentation.

lib/http/client.rb
```diff
@@ -24,7 +24,12 @@ def initialize(default_options = {})
       @state = :clean
     end
 
-    # Make an HTTP request
+    # Make an HTTP request.
+    # See HTTP::Options.defined_options for list of option keys.
+    # @see https://ruby-doc.org/stdlib-2.5.1/libdoc/openssl/rdoc/OpenSSL/SSL/SSLContext.html OpenSSL::SSL::SSLContext for more information on :ssl options.
+    # @param [Symbol] verb The HTTP method.
+    # @param [String] uri The URL/URI.
+    # @option options [Hash]
```
ixti: This should be a `@param` tag. `@option` is used to declare expected keys of the Hash/Kwargs.
```suggestion
    # @param opts [Hash]
```

lib/http/client.rb
```diff
@@ -24,7 +24,12 @@ def initialize(default_options = {})
       @state = :clean
     end
 
-    # Make an HTTP request
+    # Make an HTTP request.
+    # See HTTP::Options.defined_options for list of option keys.
```
ixti: Please, replace this with:

```suggestion
    # @see Options
```

lib/http/client.rb
```diff
@@ -24,7 +24,12 @@ def initialize(default_options = {})
...
+    # @see https://ruby-doc.org/stdlib-2.5.1/libdoc/openssl/rdoc/OpenSSL/SSL/SSLContext.html OpenSSL::SSL::SSLContext for more information on :ssl options.
+    # @param [Symbol] verb The HTTP method.
```
ixti: Please delegate this documentation to `#build_request`:
```suggestion
    # @param verb (see #build_request)
```

lib/http/client.rb
```diff
@@ -24,7 +24,12 @@ def initialize(default_options = {})
...
+    # @param [String] uri The URL/URI.
```
ixti: Please delegate this documentation to `#build_request`:
```suggestion
    # @param uri (see #build_request)
```

lib/http.rb
marlinpierce: It was not obvious.  From the example in the README I went to the HTTP class documentation and mostly found a blank page, which was very confusing.  The example use the `HTTP.get` method, but the yard doc had no `get` method documented.

lib/http/options.rb
```diff
@@ -6,6 +6,19 @@
...
+  #
+  #   :headers for request headers
+  #
+  #   :body to set the body contents
+  #
+  #   :json to send JSON
+  #
+  #   :form to send form parameters, also used with multipart form using HTTP::FormData::File
+  #
+  #   :ssl for TLS/SSL options, @see https://ruby-doc.org/stdlib-2.5.1/libdoc/openssl/rdoc/OpenSSL/SSL/SSLContext.html OpenSSL::SSL::SSLContext
```
marlinpierce: Note that I had to step through the code over many attempts and many hours to discover that the key for the options to set the TLS/SSL verify mode or path to the CA file was `:ssl`.

lib/http.rb
ixti: Are we talking about the same API doc page?

https://www.rubydoc.info/gems/http/HTTP

---

image

---

lib/http.rb
```diff
@@ -15,7 +15,16 @@
 require "http/response/body"
 require "http/response/parser"
 
-# HTTP should be easy
+# HTTP should be easy.
+#
+# Basic Usage
```
ixti:
```suggestion
# @example Basic Usage
```

lib/http.rb
```diff
@@ -15,7 +15,16 @@
...
+#
+#     HTTP.post("https://server.com", options)
```
ixti: This example looks like `options` is mandatory. And does not provide any clues on what is it. So it's better to remove it, as it's not mandatory for basic usage.

lib/http/client.rb
```diff
@@ -24,7 +24,12 @@ def initialize(default_options = {})
       @state = :clean
     end
 
-    # Make an HTTP request
+    # Make an HTTP request.
+    # @see Options
+    # @see https://ruby-doc.org/stdlib-2.5.1/libdoc/openssl/rdoc/OpenSSL/SSL/SSLContext.html OpenSSL::SSL::SSLContext for more information on :ssl options.
```
ixti: Please, remove this line. `:ssl` is no better than any other options you can set with `Options`, and I see you have already added documentation there.

```suggestion
```

lib/http/options.rb
ixti: :ssl does not related to SSLContext, for that you need `:ssl_context`. I tend to think that this should be done via YARD dsl macro.

lib/http.rb
marlinpierce: Yes that looks blank to me when I'm searching for the documentation on `HTTP.get`.  It took me a while to figure out that I needed to look at the `Chainable` module.

lib/http/options.rb
marlinpierce: It took hours of stepping through the code to know what the `:ca_file` key needed to be.  Only after finding that `:ssl` is passed to that class and seeing its attributes did I discover that it needed to be `:ca_file`.

lib/http.rb
marlinpierce: OK.  However, no where else do you say that options are passed to the `get`, `post`, or `request` method.  When I was reading the doc, I had no idea how to set options.

lib/http/client.rb
marlinpierce: Where in the doc do you say that the keys for the CA file are `{ :ssl => { :ca_file => '...' } }`?

tarcieri:
@marlinpierce I'm curious... have you looked at the Wiki documentation at all?

https://github.com/httprb/http/wiki

I think it answers most of the questions you seem to be stumbling on and has lots of code examples.

It'd be nice to get those into the yardoc too.

lib/http/client.rb
ixti: That's why I said that it should be in the options documentation not here.

lib/http.rb
ixti: Ok. But that's not the reason to bloat api docs with info that is already there.

lib/http/options.rb
ixti: I'm not saying it shouldn't be documented.  I'm saying that we must document both. And don't confuse ssl option with sslcontext

lib/http.rb
ixti: Yes, we do:

---

Screenshot_20200116-190830_Firefox Preview

---

Also API docs landing page is readme, that contains examples of usage. If you want to emphasize that there's optional argument with options- that's fine. You can do that in additional `@example`:

``` ruby
# @example POST JSON data
#     HTTP.post("https://example.com", {
#       :json => { :message => "Hello, world!" }
#     })
#
# @see Options
```

lib/http.rb
tarcieri: The README pretty clearly calls out both [the Wiki](https://github.com/httprb/http/wiki) and the `Chainable` API documentation:

Screen Shot 2020-01-16 at 10 24 02 AM

Perhaps we can move the `Documentation` section higher up to the top or something. I'm curious why you might have glossed over both of those links?

tarcieri:
Closing this PR as stale.

There's some good stuff in here, though! Please rebase and reopen if you'd like to pursue it.

Issue #592, opened 2020-02-03
Support 100 continue responses


matuszewskijan:
It seems that HTTP isn't following the `100-continue` responses:
```
HTTP.headers('Content-Type' => 'application/json', 'Accept' => 'application/json').post('https://foo.bar/api', body: '{...}')
=> #<HTTP::Response/1.1 100 Continue {}>
```

when cURL does:
```
curl -v -X POST \
>   https://foo.bar/api \
>   -H 'Accept: application/json' \
>   -H 'Content-Type: application/json' \
>   -d '{...}'
...
> POST /api HTTP/1.1
> Host: foo.bar
> Accept: application/json
> Content-Type: application/json
> Content-Length: 294
> 
* upload completely sent off: 294 out of 294 bytes
< HTTP/1.1 100 Continue
< HTTP/1.1 200 OK
...
```

I would expect the HTTP response to be `<HTTP::Response/1.1 200 OK {}>`. 

Is it something what the HTTP gem should support?

woutdegeyter:
Any feedback on this one, please? 
This is currently blocking us from building out an integration as the partner API is returning these 1xx codes.
@ixti 

Thanks already.

ixti:
If I understand correctly, yes, that's something we should support. As far as I understand second response comes immediately after the 100 Continue - in other words 1 request, but 2 responses. Am I right?

matuszewskijan:
Indeed, it seems that we are receiving 2 responses from one request. I'll see to prepare a PR to support this 100 continue code.

matuszewskijan:
I've tried to add support for reading the second response and the only solution which I've found is a change in [`readpartial`](https://github.com/httprb/http/blob/master/lib/http/connection.rb#L86-L97) method:
```   
return unless @pending_response
   chunk = @parser.read(size)
   return chunk if chunk
   finished = (read_more(size) == :eof) || @parser.finished?
+
+  if finished && @buffer.include?("HTTP/1.1 100 Continue")
+    @parser.reset
+    finished = false
+  end
+
   chunk    = @parser.read(size)
   finish_response if finished
```

It seems to be very hacky for me, and the response `code` will be always `100`. 

I've spent few hours more to find anything better. Experimented with `@state.final_chunk?` in `parser.rb` but it seems to always return false for me(even when the response isn't in chunks at all).

Anyone is having any idea where we can put the logic to support this kind of scenario?

[...]

PR #593, opened 2020-02-07
Add info response support

ixti:
This patch adds support for 1XX informational responses. Those kinds of responses are sent along with normal responses, and flow looks like:

    > GET / HTTP/1.1
    > Host: example.com
    >

    < HTTP/1.1 100 Continue
    < HTTP/1.1 200 OK
    < Content-Length: 12
    <
    < Hello World!

Notice that server responds (sends to the socket) 2 HTTP responses.

---

Resolves: #592

### DIFF
--- a/lib/http/response/parser.rb
+++ b/lib/http/response/parser.rb
@@ -86,7 +86,11 @@ module HTTP
       end
 
       def on_message_complete(_response)
-        @finished[:message] = true
+        if @state.http_status < 200
+          reset
+        else
+          @finished[:message] = true
+        end
       end
 
       def reset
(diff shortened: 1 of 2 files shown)


matuszewskijan:
@ixti is it possible to release `v5.0.0pre3` version with this patch?

We already need it to support one of our clients API.

ixti:
Ok. Will do that shortly today.

matuszewskijan:
@ixti  sorry that I bother you again. Any news on the above?

ixti:
@matuszewskijan sorry, got overwhelmed at work and forgot about this request completely. Thank for a reminder - just cut pre3 release.

PR #594, opened 2020-02-10
Reflect that Ruby 2.7.x is being tested on CI

Drenmi:
Fairly self explanatory. Ruby 2.7.x is already being tested in Travis, but the `README` does not yet reflect that there is official support. 🙂


ixti:
Thank you!

Issue #595, opened 2020-02-18
Auto retry failing requests


Bertg:
We are a happy user of HTTP.rb, but the services we use are less "happy". Intermittently they will just thrown an HTTP 500 error, or time out on us. 

It would be great if this library could provide an automatic retry mechanism (similar to [Faraday's retry](https://github.com/lostisland/faraday/blob/master/lib/faraday/request/retry.rb) implementation).

We are willing to port the behaviour of retry behaviour of Faraday into this library if there is interest to have it merged. If not we would publish it as a gem that adds the functionality to HTTP.rb.

ixti:
I have this functionality implemented as a client wrapper (ugly enough for me to be shy to open-source it)... But I think it's something nice to have.

Definitely would be happy to do a code review. And I can actually provide my implementation of this for the reference might be you'll find it good enough for the basis of your work.

@httprb/core WDYT?

paul:
@ixti Seems like this might be best implemented as a "feature", to keep the main client code cleaner...

Bertg:
@paul Excuse my ignorance here, but what exactly do you mean by "Feature" in this context?
@ixti If you'd like to share your code, I'd be happy to have a look at it

ixti:
@paul I was thinking about that, but couldn't find a way to make it sane with Features unfortunately.

@Bertg at some point we introduced a thing called `Feature` which allow us to enhance behaviour of request/response. So feature is a class that accepts some initialization options and expose 2 public methods: `wrap_response` and `wrap_request`. See implementation of deflate and inflate for example:

- https://github.com/httprb/http/blob/master/lib/http/features/auto_deflate.rb
- https://github.com/httprb/http/blob/master/lib/http/features/auto_inflate.rb

PS: @paul @Bertg I'm going to prepare the code during this week and I believe it will be a good starter. I'm not married to the implementation I made, so if there will be better ways of doing it (I'm pretty sure there will be lots of things to improve as I built it long time ago) I'm totally fin with having better implementation.

Bertg:
I've been looking into introducing this feature using the ```Feature```.
I don't think that is the right way to go. 

Based on what i can read in [lib/http/client.rb:73-88](https://github.com/httprb/http/blob/master/lib/http/client.rb#L73) ```Feature``` is used to "modify" the request outgoing or the response incoming. And to "do something" on error. We don't want to modify the request or the response. And the ```on_error```  method does not allow us to change the request handling. 

[...]

PR #598, opened 2020-03-05
Add .retriable feature to Http

Bertg:
Based on the great initial POC of @ixti this should complete #595 

* Delay by default backs backs off over time
* Maximum delay time
* Exceptions to retry from
* Status codes to retry from
* Custom retry logic
* Respect Retry-After header if present
* on_retry callback

Remarks:

* RuboCop report is clean
* Tried to manually update the method comments, but I'm not sure if that is correct


Bertg:
@paul & @ixti If you have any remarks on the PR, let me know, and I'll see to them

ixti:
I had no time to do a full review, but there's one thing that I would like to be changed. Default list of statuses to retry should be 5XX. So probably for statuses it's better to provide proc.

Bertg:
@ixti Took your feedback, and made the 500+ range auto retry. 

However, I would argue that the library should not rely any status code by default. Here is why: 

When a server is throwing 500 there is a good chance that it is experiencing trouble. By retrying by default we are making the situation worse. The same logic if followed in the faraday gem.

Bertg:
I'll also wait to to comply with the code climate feedback, until the all human feedback is processed ;)

lib/http/retriable/performer.rb
```diff
@@ -168,11 +168,20 @@ def out_of_retries_error(request, response, exception)
...
+
         ->(_req, err, res, _i) {
           if err
             exception_classes.any? { |e| err.is_a?(e) }
           else
-            retry_statuses.include?(res.status.to_i)
+            response_status = res.status.to_i
+            retry_statuses.any? do |matcher|
+              case matcher
+              when Range then matcher.include?(response_status)
```
ixti: `Range#cover?` is more efficient:

```suggestion
              when Range then matcher.cover?(response_status)
```

lib/http/retriable/performer.rb
```diff
@@ -23,7 +23,7 @@ class Performer
         IOError
       ].freeze
 
-      RETRIABLE_STATUSES = [500].freeze
+      RETRIABLE_STATUSES = (500...Float::INFINITY).freeze
```
ixti: Although there are no well known 600+ statuses,  I think this range should be explicit about upper bounds:

```suggestion
      RETRIABLE_STATUSES = (500...600).freeze
```

Bertg:
> I think handling different types of retry statuses is unneeded complication and supporting proc only is more than enough - it will handle all the cases. 

The idea is to allow for a beautiful interface for the developer.

```ruby
HTTP.retriable(retry_statuses: 400..499)
HTTP.retriable(retry_statuses: [404, 500..599])
```
looks a lot better than
```ruby
HTTP.retriable(retry_statuses: ->(s) {  (400..499).cover?(s) })
HTTP.retriable((retry_statuses: ->(s) { s == 404 || (400..499).cover?(s) })
```

I could move that convienece into ```Chainable#retriable``` but I don't believe that is the best place for it.

> Also I think exception retry proc and status retry should be separated.

I get that. However one of the options a developer can pass is a proc ```should_retry``` to decide to retry or not, based on exception or status code. I'm trying to use the same logic here. 
I guess I could make a method that looks something like this:

```ruby

def perform(client, req)
  1.upto(Float::INFINITY) do |attempt| # infinite loop with index
    err, res = try_request { yield }

    if retry_request?(req, err, res, attempt)
      begin
        wait_for_retry_or_raise(req, err, res, attempt)
      ensure
        # ...
        client.close
      end
    elsif err
      client.close
      raise err
    elsif res
      return res
    end
  end
end

def retry_request?(req, err, res, attempt)
  if options[:should_retry]
    options[:should_retry].call(req, err, res, attempt)
  elsif err
    options[:exceptions].any? {|err_class| err.is_a?(err_class) }
  else
    options[: retry_statuses].cover?(req.status)
  end
end
```

Would you prefer this?

Bertg:
Made some updates!

* extracted the delay logic into its own class
* less procs & proc building overall
* stopped handling 500+ HTTP status codes by default (see previous comment for argumentation)

Bertg:
Anything I can do to help progress this PR towards merging?

rkhapre:
This is available now in Pre 5.x releases?

yashtewari:
Any updates on this? @ixti are you still in favor of adding retry functionality to Http?

tarcieri:
The 5.x release is out. If you are still interested in pursuing this, can you rebase?

czj:
I would still love to get this PR merged :-)
Is there any way we could help @levups ?

ixti:
I can try to take a look on this next week.

tomprats:
Would love to have this. I looked into major changes since it's release and besides the hashrocket one, I don't think anything changed to break this.

@ixti would you mind taking a look again?

If you want, I can rebase, but I think if I do it, it'd have to go into a new PR

ixti:
Sorry been swamped with lots of stuff. Will take a look at the rebased PR as soon as possible. And hope to make it land this month :D

danielb2:
@ixti any progress?

wilsonsilva:
This is neat! I would love to get this out of the box.

tarcieri:
I would be okay with merging this if someone fixed the merge conflicts

tomprats:
@tarcieri I fixed it on https://github.com/httprb/http/pull/775

tarcieri:
Oh hey, sorry about that. Will take a look soon.

tarcieri:
Merged in #775

Bertg:
Hurray!

It took 1608 days to get the [original work](https://github.com/httprb/http/commit/3c148d3ba9de9b000e68ce98573fb1435d7a4881) by @ixti to make it in. Not the fastest PR to ever get merged, but glad it did and to be a part of it :)

PR #599, opened 2020-03-10
Allow passing HTTP::FormData object to HTTP::Client directly

summera:
As discussed with @ixti  in https://github.com/httprb/form_data/pull/29, this allows you to pass an `HTTP::FormData` object directly instead of a `Hash` for more customization such as using a custom encoder. Example:

```ruby
custom_encoder = proc { |data| ::JSON.dump(data) }
data = { "foo[bar]" => "test" }
form_data = HTTP::FormData.create(data, encoder: custom_encoder)
HTTP.post("https://some-url.com", form: form_data)
```

Note that I decided to use the following logic to determine whether an `HTTP::FormData` object was being passed in directly:

```ruby
    def form_data(form)
      (form || {}).respond_to?(:to_h) ? HTTP::FormData.create(form) : form
    end
```

The other option would be to check if `form` `is_a?` `HTTP::FormData::Urlencoded` or a `HTTP::FormData::Multipart` but this seemed less flexible since it creates a coupling with what's returned from `HTTP::FormData.create`. However, I don't have a strong opinion here and am happy to change if another method is more desirable.

### DIFF
--- a/lib/http/client.rb
+++ b/lib/http/client.rb
@@ -175,7 +175,7 @@ module HTTP
       when opts.body
         opts.body
       when opts.form
-        form = HTTP::FormData.create opts.form
+        form = form_data(opts.form)
         headers[Headers::CONTENT_TYPE] ||= form.content_type
         form
       when opts.json
@@ -183,6 +183,10 @@ module HTTP
         headers[Headers::CONTENT_TYPE] ||= "application/json; charset=#{body.encoding.name}"
         body
       end
+    end
+
+    def form_data(form)
+      (form || {}).respond_to?(:to_h) ? HTTP::FormData.create(form) : form
     end
   end
 end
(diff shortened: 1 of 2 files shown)


summera:
@ixti provided everything here is all good, can we also backport this to `v4.x`?

summera:
@ixti just wanted to check back with you on this. Let me know what you think.

ixti:
Honestly I think it's better to check type. But I agree that checking multiple types in here is kinda ugly. So I tend to think that we can add `coerce` to `FormData` instead and use it here instead of `create`. WDYT?

summera:
@ixti thanks for the merge and apologies for not getting back to you earlier with my thoughts! Any chance of backporting this update to 4.x?

summera:
Sorry, just realized it already was! Great 👍 . Please let me know when a new release is out with the update 😄

ixti:
Released v4.4.0

PR #608, opened 2020-05-17
Add on_redirect callback option to follow.

benubois:
Hello,

I'm interested in caching permanent redirects. I was wondering if you would accept this feature that adds support for passing an `on_redirect` option to `follow`. The `on_redirect` option accepts a proc. For example:

```ruby
HTTP.follow(on_redirect: proc { |response, location| cache(location) })
  .get("http://example.com/redirect")
```

This way http can report any redirects it follows.

Thanks!


benubois:
Looks like jruby failed due to a network error:

> ERROR:  While executing gem ... (Gem::RemoteFetcher::FetchError)
> Connection reset by peer (https://rubygems.org/specs.4.8.gz)

Maybe restarting the build would fix it?

PR #618, opened 2020-08-24
Satisfy IO#readpartial API

singpolyma-shopify:
IO#readpartial allows for a second "outbuf" param which some streaming usages
expect, so support it here to allow using response bodies anywhere IO can be.


lib/http/connection.rb
```diff
@@ -83,7 +83,7 @@ def send_request(req)
     #
     # @return [String] data chunk
     # @return [nil] when no more data left
-    def readpartial(size = BUFFER_SIZE)
+    def readpartial(size = BUFFER_SIZE, outbuf = nil)
       return unless @pending_response
 
       chunk = @parser.read(size)
```
ixti, line 89: Seems like you need to write to the outbuf here as well.

spec/lib/http/connection_spec.rb
```diff
@@ -62,5 +62,13 @@
       end
       expect(buffer).to eq "1234567890"
     end
+
+    it "fill outbuf when present" do
+      connection.read_headers!
+      outbuf = String.new
+      buffer = String.new
+      buffer << outbuf while connection.readpartial(3, outbuf)
+      expect(buffer).to eq "1234567890"
```
ixti, line 71: Due to the problem above, if you will change size to `1` or `2` you will get corrupted data: `1224447790` or `12345457890` respectively.

ixti (changes requested):

ixti:
@singpolyma-shopify thank you for the PR. I really like the proposed changes, but we need to make sure they work as expected when parser's buffer contains more data than requested size :D

singpolyma-shopify:
@ixti fixed

PR #623, opened 2020-11-05
Reorder table in alphabetically order to make it easier to read

BKSpurgeon:
* Easier to scan and check a client.

A small change, but i hope it is useful.

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,22 +45,22 @@ Top three reasons:
    http.rb achieves the best performance of any Ruby HTTP library which
    implements the HTTP protocol in Ruby instead of C:
 
-  | HTTP client              | Time   | Implementation        |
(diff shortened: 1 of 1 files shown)


paul:
It was sorted by `Time` before, which was probably the original intent, and IMO more useful. 🤷

tarcieri:
The table is also ancient and should probably be removed or recreated from scratch

tarcieri:
The best way you could possibly improve the content of this section is by performing a new benchmark 😉

BKSpurgeon:
@tarcieri Tony, I appreciate the comment. I wonder about the usefulness of these performance tests -- they tend to be opinionated. I can't guarantee anything, but I can make an attempt. Please don't hold me to it.

PR #625, opened 2020-11-09
Inflater handle empty streams

### LINKED ISSUE #624: Zlib::BufError in case of empty body response with Content-Encoding and auto_inflate use
Hi,

I found edge case where making http request to one of the remote servers can raise `Zlib::BufError`.The case is about blank body returned by a server for  `204 No Content` together with header `Content-Encoding: gzip `.

Can be reproduced this way:
```ruby
require 'socket'
server = TCPServer.new 8000
loop do

client = server.accept
client.puts <<-EOT
HTTP/1.1 204 No Content
Content-Encoding: gzip
Content-Type: application/json;charset=utf-8

EOT
client.close
end
```

Call like:
```ruby
HTTP.use(:auto_inflate, logging: { logger: Logger.new(STDOUT) }).get('http://0.0.0.0:8000')
```
raise an error `... /gems/http-5.0.0.pre3/lib/http/response/inflater.rb:19:in 'finish': buffer error (Zlib::BufError)`.

LukaszMaslej:
Hello,
This prevents the issue of Zlib::BufError errors raised when reading
by Inflater empty streams (should fix #624).

Thanks!

### DIFF
--- a/lib/http/response/inflater.rb
+++ b/lib/http/response/inflater.rb
@@ -16,7 +16,7 @@ module HTTP
         if chunk
           chunk = zstream.inflate(chunk)
         elsif !zstream.closed?
-          zstream.finish
+          zstream.finish if zstream.total_in.positive?
           zstream.close
         end
         chunk
--- a/spec/support/dummy_server/servlet.rb
+++ b/spec/support/dummy_server/servlet.rb
@@ -172,5 +172,17 @@ class DummyServer < WEBrick::HTTPServer
                    "#{req.body}-raw"
                  end
     end
+
+    post "/no-content-204" do |req, res|
+      res.status = 204
+      res.body   = ""
+
+      case req["Accept-Encoding"]
+      when "gzip" then
+        res["Content-Encoding"] = "gzip"
+      when "deflate" then
+        res["Content-Encoding"] = "deflate"
+      end
+    end
   end
 end
--- a/spec/lib/http_spec.rb
+++ b/spec/lib/http_spec.rb
@@ -438,6 +438,22 @@ RSpec.describe HTTP do
 
         expect(response.to_s).to eq("#{body}-deflated")
       end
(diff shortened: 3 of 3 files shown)


tarcieri:
@LukaszMaslej seems like a reasonable enough workaround, although there are some RuboCop failures

lib/http/response/inflater.rb
```diff
@@ -16,7 +16,7 @@ def readpartial(*args)
         if chunk
           chunk = zstream.inflate(chunk)
         elsif !zstream.closed?
-          zstream.finish
+          zstream.finish if zstream.total_in > 0
```
tarcieri, line 19: RuboCop suggests the following here:

```suggestion
          zstream.finish if zstream.total_in.positive?
```

spec/support/dummy_server/servlet.rb
```diff
@@ -172,5 +172,20 @@ def do_#{method.upcase}(req, res)
...
+      res.body = case req["Accept-Encoding"]
+                 when "gzip" then
+                   res["Content-Encoding"] = "gzip"
+                   ""
+                 when "deflate" then
+                   res["Content-Encoding"] = "deflate"
+                   ""
+                 else
+                   ""
+                 end
```
tarcieri, line 188: RuboCop suggests moving the `""` out of the conditional since it's the same for all branches

```suggestion
      res.body = ""
      case req["Accept-Encoding"]
      when "gzip" then
         res["Content-Encoding"] = "gzip"
       when "deflate" then
         res["Content-Encoding"] = "deflate"
        end
```

LukaszMaslej: @tarcieri fair suggestions, fixed.

tarcieri:
The test failures appear to be unrelated. Thanks for the fix!

Issue #633, opened 2020-12-17
Migrate to GitHub Actions


tarcieri:
[Travis CI has changed their pricing model](https://blog.travis-ci.com/2020-11-02-travis-ci-new-billing) and it's [no longer usable for OSS projects](https://www.jeffgeerling.com/blog/2020/travis-cis-new-pricing-plan-threw-wrench-my-open-source-works).

I would suggest migrating CI to GitHub Actions.

ixti:
@tarcieri :+1:

tarcieri:
See also #627 for more CI problems

tarcieri:
Here's an example template: https://github.com/postmodern/digest-crc/blob/master/.github/workflows/ruby.yml

ixti:
@tarcieri thanks, started work on this, but ssl issues are kinda make it not ok to be merged yet ;))

Issue #627, opened 2020-11-19
OpenSSL::SSL::SSLErrors in test suite


tarcieri:
CI on `master` is presently broken, with SSL-related tests failing with the following error:

https://travis-ci.org/github/httprb/http/jobs/744656845

```
#<OpenSSL::SSL::SSLError: SSL_connect returned=1 errno=0 state=error: sslv3 alert unsupported certificate>
```

PR #634, opened 2020-12-20
Switch to GH-Actions

ixti:
Resolves: #633


ixti:
Tests are failing due to SSL errors, which are caused by invalid config and/or incorrectly generated certificates. As we do some hackety hackery to mimic different ssl weirdiness, it might be the problem of misconfiguration, but I can't figure out what exactly isn't correct.

ixti:
I tend to disable failing specs temporarily and address those as a separate pr (help is super welcomed).

@tarcieri wdyt?

tarcieri (approved):

tarcieri (approved):
There's an open issue for the SSL-related failures: #627. I'm also not sure what's up there... I'm guessing something changed related to either accepted algorithms or hostnames.

Otherwise, this looks good!

lib/http/timeout/null.rb
```diff
@@ -36,8 +36,9 @@ def start_tls(host, ssl_socket_class, ssl_context)
         connect_ssl
 
         return unless ssl_context.verify_mode == OpenSSL::SSL::VERIFY_PEER
+        return unless ssl_context.respond_to?(:verify_hostname) && ssl_context.verify_hostname
```
tarcieri, line 39: In hindsight, this should probably have been:

```ruby
return if ssl_context.respond_to?(:verify_hostname) && !ssl_context.verify_hostname
```

...so as to only support explicitly *disabling* hostname verification. As is, it seems hostname verification won't be performed on JRuby at all.

tarcieri: Opened #636 as a follow-up to this comment.

ixti:
Thanks for the fix

PR #636, opened 2020-12-28
Make SSL hostname verification default to on

tarcieri:
Following up on this comment:

https://github.com/httprb/http/pull/634#pullrequestreview-559142562

The previous logic skipped hostname verification entirely if the `verify_hostname` method is not defined for `OpenSSL::SSL::SSLContext`, which is currently the case for JRuby.

This commit changes the logic so if that method is undefined, hostname verification is still performed. Otherwise, hostname verification would always be skipped on Rubies which don't define a `verify_hostname` method.

Note that this was *just* introduced in #634 which was merged 10 hours ago, so I think this was caught quickly enough simply correcting it suffices and there isn't additional security-related followup here (e.g. CVE)

### DIFF
--- a/lib/http/timeout/null.rb
+++ b/lib/http/timeout/null.rb
@@ -36,7 +36,7 @@ module HTTP
         connect_ssl
 
         return unless ssl_context.verify_mode == OpenSSL::SSL::VERIFY_PEER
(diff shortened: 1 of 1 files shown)


ixti:
Oh. Thanks @tarcieri a lot. I've totally overlooked this.

PR #638, opened 2020-12-29
DNS failover handling

midnight-wonderer:
Hello there,

I want to add DNS failover support for the client.  
This implementation inclined with Nginx's behavior when used as a reverse proxy.

I need help adding the required specs; if someone wants to help, please send a pull request to my branch.

For some context, some services use [A records](https://en.wikipedia.org/wiki/List_of_DNS_record_types) with multiple IP addresses, unless you are large enough to handles failover at the [BGP](https://en.wikipedia.org/wiki/Border_Gateway_Protocol) level, this is the only practical way to eliminate a single point of failure.  
But the failover technique needs client support, hence the pull request.

### DIFF
--- a/lib/http/timeout/global.rb
+++ b/lib/http/timeout/global.rb
@@ -1,7 +1,8 @@
 # frozen_string_literal: true
 
-require "timeout"
 require "io/wait"
+require "resolv"
+require "timeout"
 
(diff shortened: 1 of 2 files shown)


tarcieri:
Thank you!

Issue #630, opened 2020-12-13
Not working on Apple Silicon (M1) - libhttp-parser-ext.bundle: mach-o, but wrong architecture


rgaufman:
I have this simple script:

```ruby
require 'bundler/inline'

gemfile do
  source 'https://rubygems.org'
  gem 'http'
end

response = HTTP.post('http://example.com', json: nil)
```

I also tried with gem 'http', github: 'httprb/http'

I get this output:

```
Romans-MacBook-Air.local ➜  scripts git:(master) ✗ ruby test-http.rb
Traceback (most recent call last):
	24: from test-http-arse.rb:6:in `<main>'
	23: from /opt/homebrew/Cellar/ruby/2.7.2/lib/ruby/2.7.0/bundler/inline.rb:54:in `gemfile'
	22: from /opt/homebrew/Cellar/ruby/2.7.2/lib/ruby/2.7.0/bundler/settings.rb:124:in `temporary'
	21: from /opt/homebrew/Cellar/ruby/2.7.2/lib/ruby/2.7.0/bundler/inline.rb:70:in `block in gemfile'
	20: from /opt/homebrew/Cellar/ruby/2.7.2/lib/ruby/2.7.0/bundler/runtime.rb:58:in `require'
	19: from /opt/homebrew/Cellar/ruby/2.7.2/lib/ruby/2.7.0/bundler/runtime.rb:58:in `each'
	18: from /opt/homebrew/Cellar/ruby/2.7.2/lib/ruby/2.7.0/bundler/runtime.rb:69:in `block in require'
	17: from /opt/homebrew/Cellar/ruby/2.7.2/lib/ruby/2.7.0/bundler/runtime.rb:69:in `each'
	16: from /opt/homebrew/Cellar/ruby/2.7.2/lib/ruby/2.7.0/bundler/runtime.rb:74:in `block (2 levels) in require'
	15: from /opt/homebrew/Cellar/ruby/2.7.2/lib/ruby/2.7.0/bundler/runtime.rb:74:in `require'
	14: from /opt/homebrew/lib/ruby/gems/2.7.0/gems/http-4.4.1/lib/http.rb:8:in `<top (required)>'
	13: from /opt/homebrew/lib/ruby/gems/2.7.0/gems/http-4.4.1/lib/http.rb:8:in `require'
	12: from /opt/homebrew/lib/ruby/gems/2.7.0/gems/http-4.4.1/lib/http/client.rb:9:in `<top (required)>'
[...]

Issue #622, opened 2020-10-26
Bring own FFI Binding for llhttp instead of using http-parser


Jcambass:
Hi there, I know there [has been talk about dropping http-parser](https://github.com/httprb/http/issues/604) because of it's somewhat staled maintenance state. As I commented on this issue, we experienced segfaults with ruby 2.7 which forced us to dig a bit into the library and dependencies. One issue has already been [fixed in ffi](https://github.com/ffi/ffi/commit/6662bccd0bff2f64b79380e23850548af76637b7) and the [requirements have been bumped on http-parser](https://github.com/cotag/http-parser/pull/7), however we still see segfaults even with this new version.

I would love to bring http-parser back to life but even the node community seams to have given up on it:
>http-parser is not actively maintained. New projects and projects looking to migrate should consider llhttp.

As stated on the [llhttp repository](https://github.com/nodejs/llhttp):

>Let's face it, http_parser is practically unmaintainable. Even introduction of a single new method results in a significant code churn.

Instead of continuing flogging a dead horse, we should probably invest into taking another route. One possible option would be to create a new ffi binding for llhttp. From a quick search I even found a quite recent[ gem that implements ruby bindings for llhttp](https://github.com/metabahn/llhttp/), maybe this could be a good candidate to use in the future.

I'm sure that we can come up with some resources to work on this, but I would love to hear your opinion on this topic.

tarcieri:
I agree this has been an ongoing problem. If someone would like to prototype switching to a maintained alternative, that's certainly something we can consider.

sandstrom:
Maybe it would be possible to collaborate with e.g. excon or some of the other ruby http libraries on a shared http parser?

tarcieri:
`llparser` is probably the most promising way forward here. See #639.

As for FFI, it looks like @bryanp has suggested it's possible to support with `llhttp`: https://github.com/metabahn/llhttp/pull/4

bryanp:
Yes, it's possible. I need to find a couple of hours to finish the dual-implementation approach summarized here: https://github.com/metabahn/llhttp/pull/4#issuecomment-753037962. Hoping to get to that in the next couple of weeks.

bryanp:
It took a bit longer to get to, but `llhttp-ffi` is published. Read more here: https://github.com/httprb/http/pull/639#issuecomment-790293187

ixti:
Thanks @bryanp!

Issue #604, opened 2020-03-29
Time to drop http-parser?


utkarsh2102:
Hi,

https://github.com/httprb/http/blob/8b432a79072235ec445ecc3e2763b6efca64490b/http.gemspec#L33

Can we patch http to not use http-parser because it looks like it is not being maintained actively; last commit was in October 2018.
Or is there any alternative that we could possibly use?

ixti:
We had plans of switching to own ffi binding, but we don't have enough bandwidth to do so. Meanwhile, you should not worry about http-parser as long as it works fine.

Jcambass:
We run into random segfaults when we upgraded from `4.1.1` to `4.4.1`. After some digging we found that this was [caused by the new Compacting GC in ruby 2.7 and the `ffi` library ](https://github.com/ffi/ffi/issues/742)used by `http-parser`. `ffi` has [already issued a fix](https://github.com/ffi/ffi/commit/6662bccd0bff2f64b79380e23850548af76637b7) for this.

For now we fixed this by manually including `ffi 1.13.1` in our gemfile to include the fix. I also [opened a PR](https://github.com/cotag/http-parser/pull/7) to require at least ffi 1.13.1. 

However, this still makes me wonder if this project should either bring it's own ffi binding (if there are resources) or maybe more realistic, try to take over (if really needed, fork) `http-parser`.

PR #639, opened 2020-12-29
Initial implementation for llparser binding

midnight-wonderer:
Hello,  
And ping https://github.com/httprb/http/issues/604, https://github.com/httprb/http/issues/630, https://github.com/httprb/http/issues/622.

I tried hooking the llparser in; it is actually not that hard.  
I almost got every test passed; only one failed.

The test that failed is in `client_spec.rb`. The context is: "with broken body (too early closed connection)".
The failure seems to be related to the error raised in the old `parser.rb` in `add` method.  
I have no idea how this thing works, maybe a leaky abstraction?

On the implementation details
I used https://github.com/metabahn/llhttp, which is not quite production-ready.
The gem did not handle the compilation properly and required some patch to work with `http.rb`.
The patch in question is minimal, just:
~~~c
VALUE rb_llhttp_http_major(VALUE self) {
  llhttp_t *parser;

  Data_Get_Struct(self, llhttp_t, parser);

  return UINT2NUM(parser->http_major);
}

VALUE rb_llhttp_http_minor(VALUE self) {
  llhttp_t *parser;

  Data_Get_Struct(self, llhttp_t, parser);

  return UINT2NUM(parser->http_minor);
}
~~~
and
~~~c
  rb_define_method(cParser, "http_major", rb_llhttp_http_major, 0);
  rb_define_method(cParser, "http_minor", rb_llhttp_http_minor, 0);
~~~
For testing, you have to compile the extension manually after patching.

In summary, replacing the gem should not take that long, and this PR can be a starting point.

### DIFF
--- a/http.gemspec
+++ b/http.gemspec
@@ -31,6 +31,7 @@ Gem::Specification.new do |gem|
   gem.add_runtime_dependency "http-cookie",    "~> 1.0"
   gem.add_runtime_dependency "http-form_data", "~> 2.2"
   gem.add_runtime_dependency "http-parser",    "~> 1.2.0"
+  gem.add_runtime_dependency "llhttp"
 
   gem.add_development_dependency "bundler", "~> 2.0"
 
--- a/lib/http/connection.rb
+++ b/lib/http/connection.rb
@@ -3,7 +3,7 @@
 require "forwardable"
 
 require "http/headers"
-require "http/response/parser"
+require "http/response/llparser"
 
 module HTTP
   # A connection to the HTTP server
@@ -37,7 +37,7 @@ module HTTP
       @failed_proxy_connect = false
       @buffer               = "".b
 
-      @parser = Response::Parser.new
+      @parser = Response::LLParser.new
 
       @socket = options.timeout_class.new(options.timeout_options)
       @socket.connect(options.socket_class, req.socket_host, req.socket_port, options.nodelay)
--- a/lib/http/response/llparser.rb
+++ b/lib/http/response/llparser.rb
@@ -1,0 +1,124 @@
+# frozen_string_literal: true
+
+require "llhttp"
+
+module HTTP
+  class Response
+    # @api private
+
+    class LLParsingHandler < ::LLHttp::Delegate
+      def initialize(target)
+        @target = target
+        super()
+        reset
+      end
+
+      def reset
+        @header_field = nil
+      end
+
+      def on_header_field(field)
+        @header_field = field
+      end
+
+      def on_header_value(value)
+        return unless @header_field
+        @target.add_header(@header_field, value)
+        @header_field = nil
+      end
+
+      def on_headers_complete
+        @target.status_code = @target.parser.status_code
+        @target.http_version = "#{@target.parser.http_major}.#{@target.parser.http_minor}"
+        @target.mark_header_finished
+      end
+
+      def on_body(body)
+        @target.add_body(body)
+      end
+
+      def on_message_complete
+        @target.mark_message_finished
+      end
+    end
+
+    class LLParser
+      attr_reader \
+        :parser,
+        :headers
+      attr_accessor \
+        :status_code,
+        :http_version
+
+      def initialize
+        @parsing_handler = LLParsingHandler.new(self)
+        @parser = ::LLHttp::Parser.new(
+          @parsing_handler,
+          type: :response
+        )
+        reset
+      end
+
+      def reset
+        @parsing_handler.reset
+        @header_finished = false
+        @message_finished = false
+        @headers = Headers.new
+        @body_buffer&.close
+        @body_buffer&.clear
+        @body_buffer = ::Queue.new
+        @read_buffer = ""
+      end
+
+      # @return [self]
+      def add(data)
+        (parser << data).tap do |success|
+          raise IOError, "Could not parse data" unless success
+          break self
+        end
+      end
+
+      alias << add
+
+      def mark_header_finished
+        @header_finished = true
+      end
+
+      def headers?
+        @header_finished
+      end
+
+      def add_header(name, value)
+        @headers.add(name, value)
+      end
+
+      def mark_message_finished
+        @message_finished = true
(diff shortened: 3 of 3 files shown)


midnight-wonderer:
Digging further into this, I am not sure if the test is correct.  
Here what the [spec](https://tools.ietf.org/html/rfc2616#section-4.4) says:
> If a Transfer-Encoding header field (section 14.41) is present and
     has any value other than "identity", then the transfer-length is
     defined by use of the "chunked" transfer-coding (section 3.6),
     __unless the message is terminated by closing the connection__.

So, if I were to interpret it verbatim, I would not raise an error there; even if the chunked transfer seems unfinished, the closing connection should precede it in this case.

tarcieri:
Related issue: https://github.com/httprb/http/pull/424

bryanp:
@midnight-wonderer What compilation issues did you hit with `llhtttp`?

midnight-wonderer:
The llhttp gem just shipped with `.bundle` to rubygems.org, no compilation on install.
My workstation runs Linux, and it looks for `.so`, which is not there.

So an auto-deployed ruby project can't have the gem as the dependency yet.

bryanp:
@midnight-wonderer It shouldn't include `.bundle` at all, that's a mistake. But it installs fine for me on ubuntu:

```
root@ubuntu-s-1vcpu-1gb-nyc3-01:~# apt-get -y update && apt-get -y install ruby ruby-dev build-essential
...
root@ubuntu-s-1vcpu-1gb-nyc3-01:~# gem install llhttp
Fetching llhttp-0.0.2.gem
Building native extensions. This could take a while...
Successfully installed llhttp-0.0.2
Parsing documentation for llhttp-0.0.2
Installing ri documentation for llhttp-0.0.2
Done installing documentation for llhttp after 0 seconds
1 gem installed
root@ubuntu-s-1vcpu-1gb-nyc3-01:~#
```

Please open an issue at https://github.com/metabahn/llhttp if you want to discuss the issues you're hitting.

tarcieri:
Just to chime in here, I think moving to https://github.com/metabahn/llhttp sound great, but I think it'd be a lot more painless if the C extension shipped the source code as opposed to requiring a .so native dependency be installed on target systems or else installation fails.

Shipping the source code with the gem doesn't preclude dynamic linking for those who prefer it: it can still link with a dynamic library if detected. But failing to install at all without a dynamic library present is a bad user experience, IMO.

bryanp:
@tarcieri I agree with you, but `llhttp` does ship the source code.

tarcieri:
Aah, great! So it sounds like the installation issues are a bug?

bryanp:
Yeah, it seems that way. I haven't been able to reproduce so hopefully @midnight-wonderer can provide more details.

tarcieri:
Another thing to consider is JRuby.

Given https://github.com/metabahn/llhttp is written as an MRI C extension, we'd need to disable it on JRuby, but then the question is what to use instead. Perhaps we could include a pure Ruby alternative parser, or find a Java parser with a JRuby extension.

FFI would allow for using a common native extension on both platforms, which certainly simplifies things as there's only one parser to worry about.

bryanp:
I'm certainly open to moving `llhttp` to FFI.

Btw it turns out I _can_ reproduce @midnight-wonderer's issue. Requiring `llhttp` fails in the ubuntu example I posted above:

```
irb(main):001:0> require "llhttp"
       12: from /usr/bin/irb:23:in `<main>'
       11: from /usr/bin/irb:23:in `load'
       10: from /usr/lib/ruby/gems/2.7.0/gems/irb-1.2.1/exe/irb:11:in `<top (required)>'
        9: from (irb):1
        8: from /usr/lib/ruby/2.7.0/rubygems/core_ext/kernel_require.rb:156:in `require'
        7: from /usr/lib/ruby/2.7.0/rubygems/core_ext/kernel_require.rb:168:in `rescue in require'
        6: from /usr/lib/ruby/2.7.0/rubygems/core_ext/kernel_require.rb:168:in `require'
        5: from /var/lib/gems/2.7.0/gems/llhttp-0.0.2/lib/llhttp.rb:3:in `<top (required)>'
        4: from /var/lib/gems/2.7.0/gems/llhttp-0.0.2/lib/llhttp.rb:6:in `<module:LLHttp>'
        3: from /var/lib/gems/2.7.0/gems/llhttp-0.0.2/lib/llhttp.rb:6:in `require_relative'
        2: from /var/lib/gems/2.7.0/gems/llhttp-0.0.2/lib/llhttp/parser.rb:29:in `<top (required)>'
        1: from /var/lib/gems/2.7.0/gems/llhttp-0.0.2/lib/llhttp/parser.rb:29:in `require_relative'
LoadError (cannot load such file -- /var/lib/gems/2.7.0/gems/llhttp-0.0.2/lib/llhttp/llhttp_ext)
```

I'll dig in later to see what's going on.

bryanp:
I took a couple of hours to reimplement `llhttp` with FFI: https://github.com/metabahn/llhttp/pull/4. It seems to work, but it needs more testing and I'd like to do some benchmarking against the MRI C extension before committing to FFI. Open to your feedback!

ixti:
:+1: for the move to llhttp, but as @tarcieri said we should ensure we support jRuby as well. Thus if llhttp will use ffi - that will be awesome!

bryanp:
I just pushed a new release of [`llhttp`](https://github.com/metabahn/llhttp/tree/main/mri) along with a new [`llhttp-ffi`](https://github.com/metabahn/llhttp/tree/main/ffi) gem for compatibility outside of MRI (see the changelogs for specific changes). The MRI implementation is ~4x more performant than the FFI implementation, so generally speaking `llhttp-ffi` should only be used to provide backwards-compatibility to a project.

Please let me know how I can help bring this across the finish line—happy to support the effort.

tarcieri:
@bryanp perhaps you could make a new PR? (unless @midnight-wonderer wants to update this one)

bryanp:
@tarcieri done: https://github.com/httprb/http/pull/651

ixti:
#651 was merged to master. Huge thanks to everyone involved!

PR #643, opened 2021-01-19
resolve #642 - Add Pattern Matching

### LINKED ISSUE #642: [Feature] Pattern Matching extensions
I'd like to propose adding pattern matching to some of the core classes of `http`, primarily `deconstruct_keys`, to allow for matching against things like Responses and Requests.

Consider:

```ruby
response in {
  status: 200..299,
  body: /Header I want/
}

case response
in { status: 200..299, body: } then Success(body)
in { status: 300.., body: } then Failure(body)
else Failure('unknown')
end
```

This is just a theoretical example and may not be 100% accurate to the API, but I wanted to run things by the core team before implementing this as it ends up being a decent amount of additional code.

It should be noted that adding the interfaces (`deconstruct` and `deconstruct_keys`) will be backward compatible, but testing it with pattern matching will not be.

baweaver:
Resolves feature request #642 requesting the addition of Pattern
Matching hooks for Ruby 2.7+ by introducing `to_h`, `deconstruct`, and
`deconstruct_keys` to core classes.

This change also addresses spec changes by gating pattern matching specs
on Ruby 2.7+ to prevent failures in older versions. All specs have also
been given a pattern matching spec matching their implementation to
demonstrate potential usages.

While demonstrational usages could dive into nested classes that are in
`HTTP` this was avoided to keep tests isolated from eachother.

### DIFF
--- a/lib/http/headers.rb
+++ b/lib/http/headers.rb
@@ -115,12 +115,31 @@ module HTTP
     end
     alias to_hash to_h
 
+    # Pattern matching interface
+    #
+    # @param keys [Array[Symbol]]
+    #   Keys to extract
+    #
+    # @return [Hash[Symbol, Any]]
+    def deconstruct_keys(keys)
+      underscored_keys_map = underscored_keys_mapping
+
+      self
+        .to_h
+        .map { |k, v| [underscored_keys_map[k], v] }
+        .to_h
+        .slice(*keys)
+    end
+
     # Returns headers key/value pairs.
     #
     # @return [Array<[String, String]>]
     def to_a
       @pile.map { |item| item[1..2] }
     end
+
+    # Adds pattern matching interface using `to_a` as a base
+    alias_method :deconstruct, :to_a
 
     # Returns human-readable representation of `self` instance.
     #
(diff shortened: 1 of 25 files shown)


baweaver:
It appears that conditional gating for Ruby 2.7+ isn't isolating these specs. I'll have to follow up with an additional commit some time tomorrow when I have time to address that.

lib/http/headers.rb
```diff
@@ -219,6 +238,14 @@ def coerce(object)
 
     private
 
+    # Underscored version of HTTP Header keys for
+    # Pattern Matching
+    #
+    # @return [Hash[String, Symbol]]
+    def underscored_keys_mapping
+      Hash[keys.map { |k| [k, k.tr('A-Z-', 'a-z_').to_sym] }]
+    end
```
baweaver, line 247: This may be unnecessary if an alternate method is available that does the same.

baweaver:
The only way I can get around "invalid" syntax that doesn't exist until 2.7 that I can think of or get to work is unfortunately very hacky:

```ruby
if RUBY_VERSION >= '2.7'
  eval <<~RUBY
    [1,2,3] in [1, *]
  RUBY
end
```

Would love to implement this for newer versions of Ruby but can't think of anything better than that hack for the moment

baweaver:
It appears that Ruby core uses a similar hack to get around old versions:

https://github.com/ruby/ruby/commit/f9696ca6cbfe827b7ce7199e511fa7431f9333b1

baweaver:
Regarding CodeClimate I can reduce one of the methods out, but the other class not as easily. Granted my opinion is 20 methods / class is a bit arbitrary, but will defer to standards of the repo.

sferik:
Thanks for this PR! I had to rework it quite a bit, since the `main` branch has changed rather substantially since 2021. Implemented in https://github.com/httprb/http/commit/38402b6b889c82742e353dac0ae9c2ba9a650d8d.

PR #647, opened 2021-02-04
Add support for the MKCALENDAR HTTP verb

meanphil:
This is to support the RFC4791 Calendaring Extensions to WebDAV (aka, CalDAV), allowing a client to create a new calendar on a CalDAV server.


tarcieri:
Thanks!

PR #651, opened 2021-03-04
Replace http-parser with llhttp

bryanp:
Builds on https://github.com/httprb/http/pull/639 to finish updating the response parser to use `llhttp` (using the ffi implementation for broader compatibility). 

I updated one spec as skipped as `llhttp` does not return an error like `http-parser` does. Best I can tell `llhttp` exhibits the correct behavior, which was also @midnight-wonderer's take here: https://github.com/httprb/http/pull/639#issuecomment-752325117. Let me know if you disagree and I'll start a discussion with the authors of the `llhttp` c library.

Resolves https://github.com/httprb/http/issues/604, https://github.com/httprb/http/issues/622.


bryanp:
@ixti or @tarcieri: how do you feel about dropping Ruby 2.4 support with this change? Ruby 2.4 is no longer maintained (2.5 will EOL soon as well). IMO this seems like a good opportunity to drop support for old rubies.

Right now https://github.com/metabahn/llhttp only supports Ruby 2.5 and up (explaining the failing checks above). I don't think there's any reason it _couldn't_ support Ruby 2.4, but I'd prefer to only support maintained rubies. Let me know your thoughts.

tarcieri:
I'm fine with bumping the minimum Ruby version to 2.6 or newer.

ixti:
Yeah with 5.x release I was hoping to bump the min ruby version.

ixti:
I'm gonna merge https://github.com/httprb/http/pull/652 and then will rebase&merge this PR

ixti:
Need to support Ruby 2.5 due to jRuby :((

bryanp:
Speaking of jRuby, I'm trying to debug the failure above but can't reproduce locally. Have you seen this before?

```
exec: jrake: not found
```

Not sure why it's trying to use `jrake`—I don't see a reference to it in `ffi-compiler`.

ixti:
It's not related to ffi-compiler. My branch with no other changes but ruby drop - fails on jruby as well o_O

tarcieri:
@enebo @headius any idea what might be causing these `jrake: not found` errors which ostensibly appear to be coming out of `ffi-compiler` (even though that doesn't appear to directly reference `jrake` anywhere in its source code?)

ixti:
@tarcieri I propose to temporarily exclude jRuby from the CI and deal with it later

ixti:
@bryanp please rebase on master and I'll merge you PR. Will deal with jRuby later

tarcieri:
```
lib/http/response/parser.rb:13:48: C: [Correctable] Style/HashSyntax: Use hash rockets syntax.
        @parser = LLHttp::Parser.new(@handler, type: :response)
```

@ixti it might be time to consider migrating to using symbols as keys

bryanp:
@ixti done

headius:
> what might be causing these jrake: not found errors

That is a new one to me... we have not shipped a "jrake" in years and years. There used to be a push (by others, not by us) to prefix every command installed in JRuby with a "j" like "jgem", "jrake", "jrails" and so on, so perhaps ffi-compiler got caught up in that?

tarcieri:
@headius the confusing thing is I don't see `jrake` referenced anywhere in the `ffi-compiler` source code, nor do any of its dependencies (of which it only has a couple) stick out at me in particular as potential culprits

ixti:
Thank you!

bryanp:
@ixti Any idea when the next release might be pushed out?

tarcieri:
See #659 and #660

PR #656, opened 2021-04-03
Handle connection timeouts in Features

semenyukdmitry:
The change from 609f07b3454bde2f48ca5e1589cda169f7a77187 allows Features to handle errors, but it isn’t a complete implementation, because while it does handle read/write timeouts, it does not work for connection timeouts.

This pull request resolves this by moving connection establishment to the exception-handling block introduced in that commit.

### DIFF
--- a/lib/http/client.rb
+++ b/lib/http/client.rb
@@ -68,9 +68,9 @@ module HTTP
 
       @state = :dirty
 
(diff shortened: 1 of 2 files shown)


tarcieri (approved):
Looks good to me. @ixti?

rubocop failure looks unrelated

ixti:
Thank you!

PR #664, opened 2021-05-13
Bump llhttp-ffi to 0.3.0

bryanp:
Updates the `llhttp-ffi` dependency to `v0.3.0`. The only difference is that it now uses `lhttp v6.0.1`, the latest release.

### DIFF
--- a/http.gemspec
+++ b/http.gemspec
@@ -30,7 +30,7 @@ Gem::Specification.new do |gem|
   gem.add_runtime_dependency "addressable",    "~> 2.3"
   gem.add_runtime_dependency "http-cookie",    "~> 1.0"
   gem.add_runtime_dependency "http-form_data", "~> 2.2"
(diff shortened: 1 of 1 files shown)


tarcieri:
@bryanp do you think I should cut another release with this?

bryanp:
@tarcieri I don't think it's absolutely critical. None of the new features in `llhttp` are used by http.rb and the bug fixes are unlikely to affect http.rb users. `llhttp-ffi v0.0.1` (the version used by the current version of http.rb) uses `llhttp v4.0.0` which addressed CVE-2020-8287—looks like no more security related changes since.

bryanp:
@tarcieri @ixti It would be good to cut a release that includes this PR. See https://github.com/metabahn/llhttp/pull/22 (just released to `0.3.1`).

tarcieri:
Released in v5.0.1

Issue #665, opened 2021-05-17
4.x add deprecation warning to response.parse()


ronald:
In Version 5 Http.get(url).parse throws ArgumentError "wrong number of arguments (given 0, expected 1)"
https://github.com/httprb/http/blob/53b7b91ac3012dbdfb43fa65947fd8d9c2311c7c/lib/http/response.rb#L159 introduced with #538.

I suggest adding a deprecation warning in Version 4.

tarcieri:
That's a good idea.

ixti:
https://github.com/httprb/http/commit/fa542332fd625a26480bb8677be16c66660bd507

DannyBen:
Actually.... I am not sure I understand why this API was changed at all? This change makes the `#parse` method much less usable. If I know what the content type is, I do not need the HTTP gem to parse it for me - I can `JSON.parse` it myself.... I was assuming that the intention of the `#parse` method is to do "best effort" to parse it.

In addition, the response object already knows what the mime type is, so why not use it to parse?

The below code feels wrong:

```ruby
require 'http'

url = "https://jsonplaceholder.typicode.com/posts/1"
response = HTTP.get url

# Why do we have to do this...
parsed_response = response.parse(response.mime_type)

# ...instead of this?
# parsed_response = response.parse

p parsed_response
```

To me, the beauty of the HTTP gem is that it "just works". 
Can't the `#parse` method be changed to have its `type` argument optional?

```ruby
def parse(type = nil)
  type ||= mime_type
  MimeType[type].decode to_s
end
```

What am I missing?

ixti:
@DannyBen

Even if you pass `Accept: application/json` header, server might respond with plain text, or html, etc. So implicit mime type inferring from the response might lead into a confusing experience IMO:

``` ruby
HTTP.headers(:accept => "application/json").get("https://httpbin.org/html").parse
# HTTP::Error: Unknown MIME type: text/html
```

Let's say you define HTML decoder for that:

``` ruby
module HTTP
  module MimeType
    class HTML < Adapter
      def decode(str)
        ::Nokogiri::HTML str
      end
    end

    register_adapter "text/html", HTML
    register_alias   "text/html", :html
  end
end
```

And let's say you expect result to be a JSON, but it comes back as HTML (error or such):

``` ruby
payload = HTTP.headers(:accept => "application/json").get("https://httpbin.org/html").parse
```

[...]

PR #670, opened 2021-05-26
Revert `Response#parse` behavior introduced in #540

DannyBen:
Following the discussion in #665, this PR reverts the behavior of `Response#parse` to how it behaved in `4.x`.

### DIFF
--- a/lib/http/response.rb
+++ b/lib/http/response.rb
@@ -156,8 +156,8 @@ module HTTP
     # @param type [#to_s] Parse as given MIME type.
     # @raise (see MimeType.[])
     # @return [Object]
-    def parse(type)
-      MimeType[type].decode to_s
+    def parse(type = nil)
+      MimeType[type || mime_type].decode to_s
     end
 
     # Inspect a response
--- a/spec/lib/http/response_spec.rb
+++ b/spec/lib/http/response_spec.rb
@@ -87,19 +87,32 @@ RSpec.describe HTTP::Response do
   end
 
   describe "#parse" do
-    let(:headers)   { {"Content-Type" => "application/json"} }
+    let(:headers)   { {"Content-Type" => content_type} }
     let(:body)      { '{"foo":"bar"}' }
 
-    it "fails if MIME type decoder is not found" do
-      expect { response.parse "text/html" }.to raise_error(HTTP::Error)
+    context "with known content type" do
+      let(:content_type) { "application/json" }
+      it "returns parsed body" do
+        expect(response.parse).to eq "foo" => "bar"
+      end
     end
 
-    it "uses decoder found by given MIME type" do
-      expect(response.parse("application/json")).to eq("foo" => "bar")
+    context "with unknown content type" do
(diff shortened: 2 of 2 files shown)


tarcieri (approved):

paul:
I also agree with this change. If I'm getting a response back that says its "application/json", but its not, then I have to handle that *somewhere*. If I do any of `JSON(HTTP.get())` or `HTTP.get().parse` or `HTTP.get().parse(:json)` doesn't matter, they're all going to fail exactly the same way. 

If I'm writing a quick script, then letting it just fail is fine. If I'm doing something more production-worthy, I have to do the manual labor to either rescue the exception, or not use `#parse`, or just let it go to the exception tracker. All of this is the same if `#parse`  does the easy thing, or forces me to add more boilerplate, so I vote we do the easy thing by default.

lib/http/response.rb
```diff
@@ -156,8 +156,8 @@ def chunked?
     # @param type [#to_s] Parse as given MIME type.
     # @raise (see MimeType.[])
     # @return [Object]
-    def parse(type)
-      MimeType[type].decode to_s
+    def parse(type = nil)
+      MimeType[type || mime_type].decode to_s
```
ixti, line 160: How about making default to be mime-type:

```suggestion
    def parse(as = mime_type)
      MimeType[as].decode to_s
```

ixti: /cc @tarcieri @DannyBen

ixti:
> If I do any of `JSON(HTTP.get())` or `HTTP.get().parse` or `HTTP.get().parse(:json)` doesn't matter, they're all going to fail exactly the same way.

My argument is not that JSON parsing will fail, but that if one will add custom MIME type parser, `#parse` behaviour will become pretty confusing IMO. That was the main reason I had when I proposed to get rid of implicit argument. I guess I always thought of `#parse` as some sort of simple helper over `JSON.parse`.

But I have no problems with bringing back implicit mime type guessing.

lib/http/response.rb
ixti: I find this pattern to be bad, but don't have strong feelings either way :D

ixti:
Thanks for the quick PR. :D

DannyBen:
> I guess I always thought of #parse as some sort of simple helper over JSON.parse.

I completely agree with this. I for one, feel that the argument should be removed. I feel that `#parse` should do it if it can, and raise an error (a friendly one) if it can't.

tarcieri:
@ixti perhaps between this and #664 it'd be good to cut a point release

DannyBen:
> Thanks for the quick PR. :D

Thanks for being flexible and quick to merge.

ixti:
> > I guess I always thought of #parse as some sort of simple helper over JSON.parse.
> I completely agree with this. I for one, feel that the argument should be removed.

That's a good point. I like this idea.

ixti:
> @ixti perhaps between this and #664 it'd be good to cut a point release

Sure. Will tackle that ASAP

PR #676, opened 2021-07-13
Update addressable because of CVE-2021-32740

matheussilvasantos:
Hi, can we have the addressable gem updated?

Thanks!


tarcieri:
Looks like an unrelated RuboCop failure 😢

matheussilvasantos:
Yes. I just fixed it too.

tarcieri:
Thank you!

PR #678, opened 2021-07-15
feature: Restore HTTP::Response :uri option for backwards compatibility.

schwern:
This should allow folks to update to 5.x without breaking code, in my case
test factories, which make HTTP::Responses.

If nothing else, the current code has the potential for `response.uri` and `response.request.uri` to be different and that seems Bad.

* Put :uri back, but its deprecated. Remove it in 6.x?
* :uri and :request now conflict to avoid having uri being one thing and request.uri another.
    * Could allow it if uri and request.uri.to_s match?
    * Is supplying both useful for forward compatibility?
* Remove redundant internal uses of uri.

### DIFF
--- a/lib/http/response.rb
+++ b/lib/http/response.rb
@@ -40,10 +40,11 @@ module HTTP
     # @option opts [HTTP::Connection] :connection
     # @option opts [String] :encoding Encoding to use when reading body
     # @option opts [String] :body
-    # @option opts [HTTP::Request] request
+    # @option opts [HTTP::Request] request The request this is in response to.
+    # @option opts [String] :uri (DEPRECATED) used to populate a missing request
     def initialize(opts)
       @version       = opts.fetch(:version)
-      @request       = opts.fetch(:request)
+      @request       = init_request(opts)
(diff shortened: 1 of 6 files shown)


tarcieri (approved):
This looks ok to me.

@ixti WDYT?

lib/http/response.rb
```diff
@@ -164,5 +165,22 @@ def parse(type = nil)
...
+    # Initialize an HTTP::Request from options.
+    #
+    # @return [HTTP::Request]
+    def init_request(opts)
+      raise ArgumentError, ":uri is for backwards compatibilty and conflicts with :request" \
+        if opts[:request] && opts[:uri]
+
+      # For backwards compatibilty
+      if opts[:uri]
+        HTTP::Request.new(:uri => opts[:uri], :verb => :get)
```
ixti, line 180: We should probably add warning here.

ixti (approved):
This is great. I wrote that we may want to add warnings, but I've already changed my mind, and I think this should be merged as is! Thank you!

PR #679, opened 2021-08-08
Use features on redirected requests

nomis:
When following redirects, features are not used for subsequent requests.

This makes it impossible to use automatic deflation after a redirect or to log and instrument redirected requests.

Wrap requests with configured features when following redirects.


tarcieri:
Thank you!

Issue #487, opened 2018-06-26
Remove Content-Length: 0 from GET request


zt2:
When I find a GET request, there is always a `Content-Length: 0` field in the HTTP header like

```
GET / HTTP/1.1
Connection: close
Host: localhost:8080
User-Agent: http.rb/3.3.0
Content-Length: 0

```

It happened in `lib/http/request/writer.rb` `add_body_type_headers`:

```ruby
      # Adds the headers to the header array for the given request body we are working
      # with
      def add_body_type_headers
        return if @headers[Headers::CONTENT_LENGTH] || chunked?

        @request_header << "#{Headers::CONTENT_LENGTH}: #{@body.size}"
      end
```

In this case, I think the header Content-Length should only be included when you are sending a message-body which body size > 0.

This behavior trigger some bugs, for example:

```ruby
require 'http'

res = HTTP.get 'https://registry.npmjs.org'
puts res.to_s
```

will cause an error:

```
[ztz@localhost tmp]$ ruby http.rb
No frontdoor hosts available
```

by using `typhoeus`:

```ruby
require 'typhoeus'

res = Typhoeus.get 'https://registry.npmjs.org'
puts res.to_s
```

will output:

```
[ztz@localhost tmp]$ ruby typhoeus.rb 
{"db_name":"registry","doc_count":922224,"doc_del_count":342,"update_seq":11768451,"purge_seq":0,"compact_running":false,"disk_size":8918630986,"data_size":6431280597,"instance_start_time":"1528907394828029","disk_format_version":6,"committed_update_seq":11768449}
```

You can test by using commands below:

```
curl -I -H "Content-Length: 0" https://registry.npmjs.org
curl -I https://registry.npmjs.org
```

What you expected to happen: 

GET request without `Content-Length: 0`
```
GET / HTTP/1.1
Connection: close
Host: localhost:8080
User-Agent: http.rb/3.3.0

```

What actually happened: 

GET request with `Content-Length: 0`
```
GET / HTTP/1.1
Connection: close
Host: localhost:8080
User-Agent: http.rb/3.3.0
Content-Length: 0

```

Version of gem or commit ref you are using:
http 3.3.0

Version of ruby you are using:
ruby 2.5.1p57 (2018-03-29 revision 63029) [x86_64-linux]

tarcieri:
Per the HTTP/1.1 RFC including a `Content-Length: 0` is perfectly valid and recommended (`SHOULD`):

[...]

PR #684, opened 2021-08-20
Don't set Content-Length for GET, HEAD, DELETE, or CONNECT requests without a BODY

jyn514:
The HTTP spec says (https://tools.ietf.org/html/rfc7230#section-3.3.2):

> A user agent SHOULD NOT send a Content-Length header field when the request message does not contain a payload body and the method semantics do not anticipate such a body

Fixes https://github.com/httprb/http/issues/487.

This is my first time writing Ruby so hopefully the code isn't too bad :sweat_smile: happy to make changes.

### DIFF
--- a/lib/http/request/writer.rb
+++ b/lib/http/request/writer.rb
@@ -47,7 +47,11 @@ module HTTP
       # Adds the headers to the header array for the given request body we are working
       # with
       def add_body_type_headers
-        return if @headers[Headers::CONTENT_LENGTH] || chunked?
+        return if @headers[Headers::CONTENT_LENGTH] || chunked? || (
+          @body.source.nil? && %w[GET HEAD DELETE CONNECT].any? do |method|
+            @request_header[0].start_with?("#{method} ")
+          end
+        )
 
         @request_header << "#{Headers::CONTENT_LENGTH}: #{@body.size}"
       end
--- a/spec/lib/http/request/writer_spec.rb
+++ b/spec/lib/http/request/writer_spec.rb
@@ -47,8 +47,19 @@ RSpec.describe HTTP::Request::Writer do
       end
     end
 
-    context "when body is empty" do
+    context "when body is not set" do
       let(:body) { HTTP::Request::Body.new(nil) }
+
+      it "doesn't write anything to the socket and doesn't set Content-Length" do
+        writer.stream
+        expect(io.string).to eq [
+          "#{headerstart}\r\n\r\n"
+        ].join
+      end
+    end
+
+    context "when body is empty" do
+      let(:body) { HTTP::Request::Body.new("") }
 
       it "doesn't write anything to the socket and sets Content-Length" do
         writer.stream


spec/lib/http/request/writer_spec.rb
```diff
@@ -50,11 +50,10 @@
     context "when body is empty" do
       let(:body) { HTTP::Request::Body.new(nil) }
```
harrishancock, line 51: [Disclaimer: I don't know Ruby.]

Hmm, I wonder if this is distinct from the actually-empty-body case. That is, I would expect that `HTTP::Request::Body.new("")` should still produce a Content-Length: 0. Maybe worth adding a test?

lib/http/request/writer.rb
```diff
@@ -47,7 +47,9 @@ def connect_through_proxy
       # Adds the headers to the header array for the given request body we are working
       # with
       def add_body_type_headers
-        return if @headers[Headers::CONTENT_LENGTH] || chunked?
+        return if @headers[Headers::CONTENT_LENGTH] || chunked? || (
+          @request_header[0].start_with?("GET ") && @body.source.nil?
```
harrishancock, line 51: I suspect this should also apply to [HEAD](https://datatracker.ietf.org/doc/html/rfc7231#section-4.3.2), [DELETE](https://datatracker.ietf.org/doc/html/rfc7231#section-4.3.5), and [CONNECT](https://datatracker.ietf.org/doc/html/rfc7231#section-4.3.6) requests, as they all have the same wording in RFC7231 about having no defined payload semantics.

jyn514:
For context, https://github.com/node-fetch/node-fetch/issues/1249 was mistakenly filed against node-fetch instead of cloudflare - we don't allow constructing GET requests with a content-length in Cloudflare Workers. This change would allow workers to proxy GET requests that originally came from httprb.

(Technically, we allow proxying those requests, but not constructing them; but it's common to do `new Request(request.body, ...)`.)

spec/lib/http/request/writer_spec.rb
jyn514: Sure, I added a test. `Body.new("")` still adds a Content-Length: 0.

harrishancock: Nice.

bdw429s:
> we don't allow constructing GET requests with a content-length in Cloudflare Workers.

@jyn514 Thank you for clarifying that.  I've been stuck in the middle working with a software that sends HTTP requests with this library that are being rejected by Cloudflare workers written by another 3rd party who aren't helping me at all.  The assumption that the workers were using Node Fetch was on my part due to the difficulty in getting support.  (When I reached out to CloudFlare, they also refused to help me since I don't own the workers in question).

As for what the RFC says, you can read more here: https://github.com/httprb/http/issues/487

But basically, "SHOULD NOT" and "MUST NOT" are different in RFC-speak.  httprb is "bending" the spec by doing something not recommended, but it is not "breaking" the spec.  And while I've made a case in the link above for httprb to stop "bending" the spec and avoid doing things which are not recommended, I would also make a case to you that CloudFlare is wrong in outright blocking a behavior which is not explicitly forbidden in the spec.  Sending a request body is forbidden, but sending a content=length header is not.  httprb isn't the only HTTP client I've found which includes a `Content-Length: 0` header when sending GET requests.

tarcieri:
Thanks!

lib/http/request/writer.rb
```diff
@@ -47,7 +47,11 @@ def connect_through_proxy
       # Adds the headers to the header array for the given request body we are working
       # with
       def add_body_type_headers
-        return if @headers[Headers::CONTENT_LENGTH] || chunked?
+        return if @headers[Headers::CONTENT_LENGTH] || chunked? || (
+          @body.source.nil? && %w[GET HEAD DELETE CONNECT].any? do |method|
+            @request_header[0].start_with?("#{method} ")
+          end
+        )
```
ixti, line 54: I guess I'm too late, but I think this would look easier to read:

``` ruby
return if @headers[Headers::CONTENT_LENGTH] || chunked?

content_length = @body.size
return if content_length.zero? && (
  %w[GET HEAD DELETE CONNECT].any? do |method|
    @request_header[0].start_with?("#{method} ")
  end
)

```

Also, seem like it will be more efficient to pass request object into writer for this o-O

tarcieri: @ixti yeah, I thought the logic here could be better, although I didn't consider that and it seems nice.

It hasn't been released yet so....

jyn514: Checking @body.size is not correct, it won't send a Content-Length when the body is an empty string. The other changes seem fine though :)

PR #689, opened 2021-09-14
WIP: Improve implicit Content-Length guesser

ixti:
* Suggest setting `Content-Length` header manually in case we can't guess body size.

See-Also: https://github.com/httprb/http/pull/560


Skandiakiwi13 (approved):
A

PR #690, opened 2021-09-23
Remove note about 5.x only being a dev branch now that its been released

sandersch:
Small update to the README to remove note about 4.x now that 5.x is the stable branch.

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,10 +7,6 @@
 
 [Documentation]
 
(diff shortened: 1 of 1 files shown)


tarcieri:
Seems like we may not have the rubocop version pinned? /cc @ixti

tarcieri:
I opened #691 to pin the rubocop version to the latest and fix the current offenses

tarcieri:
Thanks!

Issue #693, opened 2021-09-29
Cookies extraction is not working properly in 5.x


flosacca:
In 5.0.2, `HTTP::Response#cookies` does not return cookies set by Set-Cookie header fields in a non-canonical form (e.g. `set-cookie` rather than `Set-Cookie`). It seems due to #576. The behavior of `HTTP::Headers#each` changes to yield the original field name, but `HTTP::Response#cookies` iterates over header fields by calling `HTTP::Headers#each` and only finds fields with name exactly equal to `Set-Cookie`.

tarcieri:
It would seem this would impact all 5.x releases, not just 5.0.2?

tarcieri:
Thanks for confirming. Seems like an easy fix if anyone would like to open a PR.

PR #694, opened 2021-09-30
Fix cookies extraction

flosacca:
Follows #693.

### DIFF
--- a/lib/http/response.rb
+++ b/lib/http/response.rb
@@ -138,7 +138,7 @@ module HTTP
     def_delegator :content_type, :charset
 
     def cookies
-      @cookies ||= headers[Headers::SET_COOKIE].each_with_object CookieJar.new do |v, jar|
+      @cookies ||= headers.get(Headers::SET_COOKIE).each_with_object CookieJar.new do |v, jar|
         jar.parse(v, uri)
       end
     end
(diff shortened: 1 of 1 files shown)


lib/http/response.rb
```diff
@@ -138,8 +138,8 @@ def content_type
     def_delegator :content_type, :charset
 
     def cookies
-      @cookies ||= headers.each_with_object CookieJar.new do |(k, v), jar|
-        jar.parse(v, uri) if k == Headers::SET_COOKIE
+      @cookies ||= headers[Headers::SET_COOKIE].each_with_object CookieJar.new do |v, jar|
+        jar.parse(v, uri)
```
tarcieri, line 142: `each_with_object` doesn't work because it's returning a string. How about something like:

```suggestion
      @cookies ||= CookieJar.new do |jar|
          jar.parse(headers[Headers::SET_COOKIE], url) if headers.has_key?(Headers::SET_COOKIE)
      end
```

flosacca:
Sorry, I only tested the case where more than one cookies were present. The problem can be fixed by simply replacing `Headers#[]` with `Headers#get` since the latter always returns an array. I have passed CI on another branch this time.

tarcieri:
Thanks!

PR #707, opened 2022-02-04
Distinguish connection timeouts

YuLeven:
HTTP.rb offers us a way of configuring open timeouts, but there is no way to tell which timeout is what from the caller's point of view.

In some scenarios such as systems prone to occasional network intermittencies, singling out a short connection timeout to retry specifically for that case is often desirable.  A catch-all retry strategy like we have today makes it hard to decide when to attempt a cheap and quick network intermittence retry from a long, target endpoint-health related write or read retry. 

I'm proposing we introduce `ConnectionTimeoutError` as a subclass of `TimeoutError` so it becomes possible for the user to selectively catch open-time specific timeouts while maintaining compatibility with catch-all code currently using `TimeoutError`.

### DIFF
--- a/lib/http/errors.rb
+++ b/lib/http/errors.rb
@@ -19,6 +19,9 @@ module HTTP
   # Generic Timeout error
   class TimeoutError < Error; end
 
+  # Timeout when first establishing the conncetion
+  class ConnectionTimeoutError < TimeoutError; end
+
   # Header value is of unexpected format (similar to Net::HTTPHeaderSyntaxError)
   class HeaderError < Error; end
 end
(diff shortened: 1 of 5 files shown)


lib/http/errors.rb
```diff
@@ -19,6 +19,9 @@ class StateError < ResponseError; end
   # Generic Timeout error
   class TimeoutError < Error; end
 
+  # Timeout when first establishing the conncetion
+  class ConnectionTimeoutError < TimeoutError; end
```
YuLeven, line 23: Wondering if `OpenTimeoutError` might read better as it alines with what `Net::HTTP` has in place. Collective memory and all.

YuLeven: Neat. I like this. Makes it more intuitive for folks with the underlying syscall on mind.

tarcieri:
Thank you!

PR #712, opened 2022-04-22
Recognize cookies set by redirect

tkellogg:
When a redirect response arrives, the browser will recognize
`Set-Cookie` headers the same as any other request/response. This PR
enables the Redirector to correctly follow the cookie hand-offs.

I encountered this situation out in the wild. This PR, as-is, seems to
fix the issue. I haven't bothered to find the relevant spec.

### DIFF
--- a/lib/http/redirector.rb
+++ b/lib/http/redirector.rb
@@ -59,11 +59,42 @@ module HTTP
         @response.flush
 
         # XXX(ixti): using `Array#inject` to return `nil` if no Location header.
-        @request  = redirect_to(@response.headers.get(Headers::LOCATION).inject(:+))
+        @request = redirect_to(@response.headers.get(Headers::LOCATION).inject(:+))
+        self.class.update_cookies(@response, @request)
         @response = yield @request
       end
 
       @response
+    end
+
+    class << self
+      # Used internally to update cookies between redirects. If a redirct response contains
+      # a Set-Cookie header(s), the following request should have that cookie set.
+      #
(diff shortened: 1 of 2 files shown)


tkellogg:
Ah, there's still a problem with this — the cookies set by redirects still aren't set in the final response object. Gonna need a bit more work

tarcieri:
A bit of sleuthing shows this behavior is implemented by all major browsers, so it seems like we should support it: https://blog.dubbelboer.com/2012/11/25/302-cookie.html

tarcieri:
Looks like some unrelated rubocop failures. I thought we had rubocop pinned to prevent that...

tkellogg:
@tarcieri This is ready now, as far as I'm concerned. Let me know if anything needs fixing

tarcieri:
@tkellogg some rubocop failures

tarcieri:
I think these are fixed by the rubocop update?

tarcieri:
Aah nope, oh well. #718 fixes them

ixti:
I have some plans of removing redirector as is, and introducing session object instead. Session will be a http client wrapper that will hold cookies. This PR has these lines:

``` ruby
      # I wish we could just do @response.cookes = cookie_jar
      cookie_jar.each do |cookie|
        @response.cookies.add(cookie)
      end
```

I believe this is wrong and comes from the idea of using redirector as a session. The response should not hold cookie jar of previous requests.

tarcieri:
A `Session` type to encapsulate this state would be great. See also #306.

Issue #714, opened 2022-06-13
Should responses with content-type application/json have UTF-8 encoding by default?


drwl:
Currently, responses without a specified charset encoding in the header are stored internally with binary/ascii 8bit encoding. According to the spec, the specified charset _should_ be redundant. Should the default for json responses by utf-8 instead?

From [RFC8259](https://datatracker.ietf.org/doc/html/rfc8259#section-8.1):
> JSON text exchanged between systems that are not part of a closed
>    ecosystem MUST be encoded using UTF-8 [[RFC3629](https://datatracker.ietf.org/doc/html/rfc3629)].
> 
>    Previous specifications of JSON have not required the use of UTF-8
>    when transmitting JSON text.  However, the vast majority of JSON-
>    based software implementations have chosen to use the UTF-8 encoding,
>    to the extent that it is the only encoding that achieves
>    interoperability.

Also in RFC8259, at the end of Section 11 (or just above of Section 12):
>    Note:  No "charset" parameter is defined for this registration.
>       Adding one really has no effect on compliant recipients.

tarcieri:
That seems like a reasonable default, yes

drwl:
> That seems like a reasonable default, yes

I can try to put up a PR. Where do you suggest this code change take place?

My initial thought is to change HTTP::Response initializer:
```ruby
        encoding   = opts[:encoding] || charset

        if encoding.nil?
          if mime_type == 'application/json'
            encoding = Encoding::UTF_8
          else
            encoding = Encoding::BINARY
          end
        end
```

thoughts?

ixti:
@tarcieri I like the idea. Reviewed the PR.

ixti:
I can cut release layer today

PR #715, opened 2022-06-14
Set default encoding to UTF-8 for json

drwl:
This PR changes the default encoding for response with the type `application/json` to be UTF-8 by default, instead of binary.

Implements https://github.com/httprb/http/issues/714.

### DIFF
--- a/lib/http/response.rb
+++ b/lib/http/response.rb
@@ -53,7 +53,7 @@ module HTTP
         @body = opts.fetch(:body)
       else
         connection = opts.fetch(:connection)
-        encoding   = fetch_or_default_encoding(opts)
+        encoding = opts[:encoding] || charset || default_encoding
 
         @body = Response::Body.new(connection, :encoding => encoding)
       end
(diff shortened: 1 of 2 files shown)


lib/http/response.rb
```diff
@@ -53,7 +53,7 @@ def initialize(opts)
         @body = opts.fetch(:body)
       else
         connection = opts.fetch(:connection)
-        encoding   = opts[:encoding] || charset || Encoding::BINARY
+        encoding   = fetch_or_default_encoding(opts)
```
ixti, line 56:
```suggestion
        encoding   = opts[:encoding] || charset || default_encoding
```

lib/http/response.rb
```diff
@@ -168,6 +168,17 @@ def inspect
...
+    def fetch_or_default_encoding(opts)
+      return opts[:encoding] if opts[:encoding]
+      return charset if charset
+
+      if mime_type == "application/json"
+        Encoding::UTF_8
+      else
+        Encoding::BINARY
+      end
+    end
```
ixti, line 180:
```suggestion
    def default_encoding
      return Encoding::UTF_8 if mime_type == "application/json"

      Encoding::BINARY
    end
```

ixti (changes requested):
Let's add `default_charset` method that will handle mime-type based logic only.

ixti:
Rubocop failures are unrelated - I'll fix them separately.

ixti:
Merged as https://github.com/httprb/http/commit/09848c262992d9b87ae64089a62b8ae1122b6ba2

PR #720, opened 2022-06-20
Add JRUBY to CI

ixti:


ixti:
@tarcieri should we maintain JRuby support? It's kinda holing us from bumping min ruby version to 2.6

tarcieri:
@ixti you could always do something like:

```ruby
gem.required_ruby_version = defined?(JRUBY_VERSION) ? "2.5" : "2.6"
```

tarcieri:
Aah. I imagine there are a number of JRuby users of this library still (indeed that's probably still true of a few of my former employers).

Unless there's a big benefit to dropping support, I suggest we continue to do so.

ixti:
There are some benefits, but most of them are in my head then real world. ))) If jruby is used by the http gem consumers then let's keep it )))

ixti:
Will need to fix CI then ))) or mark some tests as non jruby friendly )))

ixti:
@tarcieri I've added JRuby into "flaky specs group" for now. because re-running individual failing specs works just fine. Which means there might be some state related issue thuogh :((

Issue #535, opened 2019-02-27
Encoding is messed up with AutoInflate


ixti:
When requesting a resource with auto_inflate feature on, encoding of body most certainly will be messed up.

To reproduce:

``` ruby
HTTP.use(:auto_inflate).headers("Accept-Encoding" => "gzip").get(url)
```

`url` in the above is some URL that returns response with headers like this:

```
Content-Type: application/json; charset=utf-8
Content-Encoding: gzip
```

in this example, encoding of the body will be `Encoding::BINARY`. Even worse, if you pass `:encoding => Encoding::UTF_8` explicitly when calling `#get` it won't be respected at the end.

PR #721, opened 2022-08-04
Fix AutoInflate with encoding other than Encoding::BINARY.

benubois:
Fix for #535

### DIFF
--- a/lib/http/features/auto_inflate.rb
+++ b/lib/http/features/auto_inflate.rb
@@ -17,15 +17,15 @@ module HTTP
           :headers       => response.headers,
           :proxy_headers => response.proxy_headers,
           :connection    => response.connection,
-          :body          => stream_for(response.connection),
+          :body          => stream_for(response.connection, response.body.encoding),
           :request       => response.request
         }
 
         Response.new(options)
       end
 
-      def stream_for(connection)
-        Response::Body.new(Response::Inflater.new(connection))
+      def stream_for(connection, encoding)
+        Response::Body.new(Response::Inflater.new(connection), :encoding => encoding)
       end
 
       private
--- a/spec/lib/http_spec.rb
+++ b/spec/lib/http_spec.rb
@@ -456,6 +456,15 @@ RSpec.describe HTTP do
 
         expect(response.to_s).to eq("")
       end
+      
+      it "returns decoded body with the correct charset" do
+        encoding = Encoding::Shift_JIS
+        client   = HTTP.use(:auto_inflate).headers("Accept-Encoding" => "gzip").encoding(encoding)
+        body     = "もしもし!".encode(encoding, Encoding::UTF_8)
+        response = client.post("#{dummy.endpoint}/encoded-body", :body => body)
+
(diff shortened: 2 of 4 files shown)


lib/http/features/auto_inflate.rb
```diff
@@ -17,15 +17,15 @@ def wrap_response(response)
...
+          :body          => stream_for(response.connection, response.body.encoding),
           :request       => response.request
         }
 
         Response.new(options)
       end
 
-      def stream_for(connection)
-        Response::Body.new(Response::Inflater.new(connection))
+      def stream_for(connection, encoding)
```
tarcieri, line 27: This would be a breaking change to a public API, however you could do:

```suggestion
      def stream_for(connection, encoding: Encoding::BINARY)
```

benubois: Fixed!

spec/lib/http_spec.rb
```diff
@@ -456,6 +456,15 @@ def setsockopt(*args)
 
         expect(response.to_s).to eq("")
       end
+
+      it "returns decoded body with the correct charset" do
+        encoding = Encoding::Shift_JIS
+        client   = HTTP.use(:auto_inflate).headers("Accept-Encoding" => "gzip").encoding(encoding)
+        body     = "もしもし!".encode(encoding, Encoding::UTF_8)
+        response = client.post("#{dummy.endpoint}/encoded-body", :body => body)
```
ixti, line 464: I'm struggling to understand this test.

``` ruby
client   = HTTP.use(:auto_inflate).headers("Accept-Encoding" => "gzip").encoding(encoding)
```
^^^ `encoding(...)` enforces encoding of the response.

lib/http/features/auto_inflate.rb
```diff
@@ -17,15 +17,15 @@ def wrap_response(response)
...
           :request       => response.request
         }
 
         Response.new(options)
       end
 
-      def stream_for(connection)
-        Response::Body.new(Response::Inflater.new(connection))
+      def stream_for(connection, encoding: Encoding::BINARY)
+        Response::Body.new(Response::Inflater.new(connection), :encoding => encoding)
```
ixti, line 28: `stream_for` can and should be private, and `encoding:` seems like don't need default value.

lib/http/features/auto_inflate.rb
```diff
@@ -17,15 +17,15 @@ def wrap_response(response)
           :headers       => response.headers,
           :proxy_headers => response.proxy_headers,
           :connection    => response.connection,
-          :body          => stream_for(response.connection),
+          :body          => stream_for(response.connection, :encoding => response.body.encoding),
```
ixti, line 20: Let's move logic to `stream_for`:

```suggestion
          :body          => wrap_body(response),
```

Then move and rename `stream_for` to be `private` `wrap_body`:

``` ruby
private

def wrap_body(response)
  stream = Response::Inflater.new(response.connection)

  Response::Body.new(stream, :encoding => response.body.encoding)
end
```

ixti (changes requested):

PR #723, opened 2022-08-13
raise HTTP::Request::InvalidURIError for wrong url's

### LINKED ISSUE #565: Better error messages for invalid URLs
```
HTTP.get(nil)
HTTP::Request::UnsupportedSchemeError (unknown scheme: )

HTTP.get('')
HTTP::Request::UnsupportedSchemeError (unknown scheme: )

HTTP.get(':')
Addressable::URI::InvalidURIError (Cannot assemble URI string with ambiguous path: ':')
```

Ideally it should return InvalidURI exceptions if the URL is invalid instead of UnsupportedSchemeError, as it is less confusing

skatkov:
Really enjoy using your gem, so decided to study internal of it better. 

is there any better way to understand internals, when contributing to gem directly? I guess, not.

This PR addressed an issue that is assigned for 5.0 milestone

closes #565

### DIFF
--- a/lib/http/client.rb
+++ b/lib/http/client.rb
@@ -154,6 +154,8 @@ module HTTP
       uri.path = "/" if uri.path.empty?
 
       uri
+    rescue Addressable::URI::InvalidURIError => e
+      raise HTTP::Request::InvalidURIError, e.message
     end
 
     # Creates request headers with cookies (if any) merged in
--- a/lib/http/request.rb
+++ b/lib/http/request.rb
@@ -23,6 +23,9 @@ module HTTP
     # The scheme of given URI was not understood
     class UnsupportedSchemeError < RequestError; end
 
+    # Provided url is missing or wrong
+    class InvalidURIError < RequestError; end
+
     # Default User-Agent header value
     USER_AGENT = "http.rb/#{HTTP::VERSION}"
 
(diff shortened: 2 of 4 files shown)


lib/http/client.rb
```diff
@@ -154,6 +154,8 @@ def make_request_uri(uri, opts)
       uri.path = "/" if uri.path.empty?
 
       uri
+    rescue Addressable::URI::InvalidURIError => e
+      raise HTTP::Request::InvalidURIError, e.message
```
ixti, line 158: I wonder if we should move this to `HTTP::URI.parse` instead? And probably call it `HTTP::URI::InvalidError`

lib/http/request.rb
```diff
@@ -222,6 +224,12 @@ def inspect
     #   @return [String]
     def_delegator :@uri, :host
 
+    def check_errors
+      raise(InvalidURIError, "empty uri provided") if @uri.host.nil?
+      raise(UnsupportedMethodError, "unknown method: #{verb}") unless METHODS.include?(@verb)
+      raise(UnsupportedSchemeError, "unknown scheme: #{scheme}") unless SCHEMES.include?(@scheme)
```
ixti, line 230: IMO, all of the above should raise `ArgumentError`. Please make this method private. Furthermore, what are your thoughts on renaming it to `validate!`?

PR #731, opened 2022-11-30
Strip brackets from IPv6 addresses before passing them to a socket.

### LINKED ISSUE #730: Requests to an IPv6 address result in: HTTP::ConnectionError (failed to connect: getaddrinfo: Name or service not known)
I'm trying to use http.rb to make GET requests to IPv6 addresses and always get the following error:

> HTTP::ConnectionError (failed to connect: getaddrinfo: Name or service not known)

I've tried with and without brackets around the IP address and the results are the same. (I'd expect the version without brackets to fail, so that by itself isn't a problem. More detail on why I'm even trying a bracket-less version below.)

Interestingly, making a request to the exact same URL with httparty succeeds when using the bracketed form.

Here is a demonstration of what I see using the IPv6 address for example.com (2606:2800:220:1:248:1893:25c8:1946):

http.rb without brackets:
```
irb(main):177:0> HTTP.get("http://2606:2800:220:1:248:1893:25c8:1946/foo?bar=baz")
       15: from script/rails:6:in `<main>'
       14: from script/rails:6:in `require'
       13: from /redacted/vendor/bundle/ruby/2.6.0/gems/railties-3.2.22.30/lib/rails/commands.rb:41:in `<top (required)>'
       12: from /redacted/vendor/bundle/ruby/2.6.0/gems/railties-3.2.22.30/lib/rails/commands/console.rb:8:in `start'
       11: from /redacted/vendor/bundle/ruby/2.6.0/gems/railties-3.2.22.30/lib/rails/commands/console.rb:47:in `start'
       10: from (irb):177
        9: from /redacted/vendor/bundle/ruby/2.6.0/gems/http-3.3.0/lib/http/chainable.rb:20:in `get'
        8: from /redacted/vendor/bundle/ruby/2.6.0/gems/http-3.3.0/lib/http/chainable.rb:77:in `request'
        7: from /redacted/vendor/bundle/ruby/2.6.0/gems/http-3.3.0/lib/http/client.rb:30:in `request'
        6: from /redacted/vendor/bundle/ruby/2.6.0/gems/http-3.3.0/lib/http/client.rb:67:in `perform'
        5: from /redacted/vendor/bundle/ruby/2.6.0/gems/http-3.3.0/lib/http/client.rb:67:in `new'
        4: from /redacted/vendor/bundle/ruby/2.6.0/gems/http-3.3.0/lib/http/connection.rb:43:in `initialize'
        3: from /redacted/vendor/bundle/ruby/2.6.0/gems/http-3.3.0/lib/http/timeout/null.rb:21:in `connect'
        2: from /redacted/vendor/bundle/ruby/2.6.0/gems/http-3.3.0/lib/http/timeout/null.rb:21:in `open'
        1: from /redacted/vendor/bundle/ruby/2.6.0/gems/http-3.3.0/lib/http/timeout/null.rb:21:in `initialize'
HTTP::ConnectionError (failed to connect: getaddrinfo: Name or service not known)
```
http.rb with brackets:
```
irb(main):178:0> HTTP.get("http://[2606:2800:220:1:248:1893:25c8:1946]/foo?bar=baz")
       16: from script/rails:6:in `<main>'
       15: from script/rails:6:in `require'
       14: from /redacted/vendor/bundle/ruby/2.6.0/gems/railties-3.2.22.30/lib/rails/commands.rb:41:in `<top (required)>'
       13: from /redacted/vendor/bundle/ruby/2.6.0/gems/railties-3.2.22.30/lib/rails/commands/console.rb:8:in `start'
       12: from /redacted/vendor/bundle/ruby/2.6.0/gems/railties-3.2.22.30/lib/rails/commands/console.rb:47:in `start'
       11: from (irb):178
       10: from (irb):178:in `rescue in irb_binding'
        9: from /redacted/vendor/bundle/ruby/2.6.0/gems/http-3.3.0/lib/http/chainable.rb:20:in `get'
        8: from /redacted/vendor/bundle/ruby/2.6.0/gems/http-3.3.0/lib/http/chainable.rb:77:in `request'
        7: from /redacted/vendor/bundle/ruby/2.6.0/gems/http-3.3.0/lib/http/client.rb:30:in `request'
        6: from /redacted/vendor/bundle/ruby/2.6.0/gems/http-3.3.0/lib/http/client.rb:67:in `perform'
        5: from /redacted/vendor/bundle/ruby/2.6.0/gems/http-3.3.0/lib/http/client.rb:67:in `new'
        4: from /redacted/vendor/bundle/ruby/2.6.0/gems/http-3.3.0/lib/http/connection.rb:43:in `initialize'
        3: from /redacted/vendor/bundle/ruby/2.6.0/gems/http-3.3.0/lib/http/timeout/null.rb:21:in `connect'
        2: from /redacted/vendor/bundle/ruby/2.6.0/gems/http-3.3.0/lib/http/timeout/null.rb:21:in `open'
        1: from /redacted/vendor/bundle/ruby/2.6.0/gems/http-3.3.0/lib/http/timeout/null.rb:21:in `initialize'
HTTP::ConnectionError (failed to connect: getaddrinfo: Name or service not known)
```
httparty without brackets:
```
irb(main):179:0> HTTParty.get("http://2606:2800:220:1:248:1893:25c8:1946/foo?bar=baz")
       16: from script/rails:6:in `<main>'
       15: from script/rails:6:in `require'
       14: from /redacted/vendor/bundle/ruby/2.6.0/gems/railties-3.2.22.30/lib/rails/commands.rb:41:in `<top (required)>'
       13: from /redacted/vendor/bundle/ruby/2.6.0/gems/railties-3.2.22.30/lib/rails/commands/console.rb:8:in `start'
       12: from /redacted/vendor/bundle/ruby/2.6.0/gems/railties-3.2.22.30/lib/rails/commands/console.rb:47:in `start'
       11: from (irb):179
       10: from (irb):179:in `rescue in irb_binding'
        9: from /redacted/vendor/bundle/ruby/2.6.0/gems/httparty-0.16.2/lib/httparty.rb:601:in `get'
        8: from /redacted/vendor/bundle/ruby/2.6.0/gems/httparty-0.16.2/lib/httparty.rb:489:in `get'
        7: from /redacted/vendor/bundle/ruby/2.6.0/gems/httparty-0.16.2/lib/httparty.rb:563:in `perform_request'
        6: from /redacted/vendor/bundle/ruby/2.6.0/gems/httparty-0.16.2/lib/httparty.rb:563:in `new'
        5: from /redacted/vendor/bundle/ruby/2.6.0/gems/httparty-0.16.2/lib/httparty/request.rb:63:in `initialize'
        4: from /redacted/vendor/bundle/ruby/2.6.0/gems/httparty-0.16.2/lib/httparty/request.rb:73:in `path='
        3: from /opt/rubies/ruby-2.6/lib/ruby/2.6.0/uri/common.rb:234:in `parse'
        2: from /opt/rubies/ruby-2.6/lib/ruby/2.6.0/uri/rfc3986_parser.rb:73:in `parse'
        1: from /opt/rubies/ruby-2.6/lib/ruby/2.6.0/uri/rfc3986_parser.rb:67:in `split'
URI::InvalidURIError (bad URI(is not URI?): "http://2606:2800:220:1:248:1893:25c8:1946/foo?bar=baz")
```
httparty with brackets:
```
irb(main):180:0> HTTParty.get("http://[2606:2800:220:1:248:1893:25c8:1946]/foo?bar=baz")
=> #<HTTParty::Response:0x7f257239b888 parsed_response="<?xml version=\"1.0\" encoding=\"iso-8859-1\"?>\n<!DOCTYPE html PUBLIC \"-//W3C//DTD XHTML 1.0 Transitional//EN\"\n         \"http://www.w3.org/TR/xhtml1/DTD/xhtml1-transitional.dtd\">\n<html xmlns=\"http://www.w3.org/1999/xhtml\" xml:lang=\"en\" lang=\"en\">\n\t<head>\n\t\t<title>404 - Not Found</title>\n\t</head>\n\t<body>\n\t\t<h1>404 - Not Found</h1>\n\t</body>\n</html>\n", @response=#<Net::HTTPNotFound 404 Not Found readbody=true>, @headers={"content-type"=>["text/html"], "date"=>["Sat, 19 Nov 2022 01:56:17 GMT"], "server"=>["ECS (dab/DC47)"], "content-length"=>["345"], "connection"=>["close"]}>
```

I realize I'm using old versions of both HTTP libraries as well as Rails (working with a large corporate codebase that's very outdated), but I've also tried this with http.rb v5 in a Pry session outside of the context of this project, and the same error occurs.

Normally I would just use httparty, because it works, but unfortunately I can't because of a bad interaction between httparty and the newrelic_rpm gem. Apparently `Net::HTTP` needs brackets removed from IPv6 addresses, so httparty has some code that intentionally removes the brackets before handing off the request to `Net::HTTP`. At this point, New Relic instrumentation hooks into the request, attempts to re-parse the URL using the addressable gem, and raises an exception because it requires brackets around IPv6 addresses. None of this is http.rb's fault, I'm just mentioning it for the context as to why I'm trying use http.rb in the first place. I wanted to try something that doesn't use `Net::HTTP` under the hood so that there's no conflict in support for brackets around IPv6 addresses. I see that both http.rb and newrelic_rpm use the addressable gem for URLs, so they should be compatible.

Any idea why I'm getting this error from http.rb? I wasn't able to find documentation or issues that specifically mention IPv6 to see examples of how it's supposed to work or how to troubleshoot.

Thank you for your time.

jeraki:
Fixes #730.

I took my best guess at how you'd want this change implemented and tested, but I'm happy to revise it based on feedback. Thanks!

### DIFF
--- a/lib/http/uri.rb
+++ b/lib/http/uri.rb
@@ -19,6 +19,18 @@ module HTTP
     def_delegators :@uri, :fragment, :normalized_fragment, :fragment=
     def_delegators :@uri, :omit, :join, :normalize
 
+    # Host, either a domain name or IP address. If the host is an IPv6 address, it will be returned
+    # without brackets surrounding it.
+    #
+    # @return [String] The host of the URI
+    attr_reader :host
+
+    # Normalized host, either a domain name or IP address. If the host is an IPv6 address, it will
+    # be returned without brackets surrounding it.
+    #
+    # @return [String] The normalized host of the URI
+    attr_reader :normalized_host
+
     # @private
     HTTP_SCHEME = "http"
 
@@ -82,6 +94,9 @@ module HTTP
       else
         raise TypeError, "expected Hash for options, got #{options_or_uri.class}"
       end
+
+      @host = process_ipv6_brackets(@uri.host)
+      @normalized_host = process_ipv6_brackets(@uri.normalized_host)
     end
 
     # Are these URI objects equal? Normalizes both URIs prior to comparison
@@ -109,52 +124,15 @@ module HTTP
       @hash ||= to_s.hash * -1
     end
 
-    # Host, either a domain name or IP address. If the host is an IPv6 address, it will be returned
-    # without brackets surrounding it.
-    #
-    # @return [String] The host of the URI
-    def host
-      @host ||= begin
-        ip = IPAddr.new(@uri.host)
-
-        if ip.ipv6?
-          ip.to_s
-        else
-          @uri.host
-        end
-      rescue IPAddr::Error
-        @uri.host
-      end
-    end
-
-    # Normalized host, either a domain name or IP address. If the host is an IPv6 address, it will
-    # be returned without brackets surrounding it.
-    #
-    # @return [String] The normalized host of the URI
-    def normalized_host
-      @normalized_host ||= begin
-        ip = IPAddr.new(@uri.normalized_host)
-
-        if ip.ipv6?
-          ip.to_s
-        else
-          @uri.normalized_host
-        end
-      rescue IPAddr::Error
-        @uri.normalized_host
-      end
-    end
-
     # Sets the host component for the URI.
     #
     # @param [String, #to_str] new_host The new host component.
     # @return [void]
     def host=(new_host)
-      @uri.host = new_host
+      @uri.host = process_ipv6_brackets(new_host, :brackets => true)
 
-      # Reset dependent values
-      remove_instance_variable(:@host) if defined?(@host)
-      remove_instance_variable(:@normalized_host) if defined?(@normalized_host)
+      @host = process_ipv6_brackets(@uri.host)
+      @normalized_host = process_ipv6_brackets(@uri.normalized_host)
     end
 
     # Port number, either as specified or the default if unspecified
@@ -192,6 +170,26 @@ module HTTP
     # @return [String] human-readable representation of URI
     def inspect
       format("#<%s:0x%014x URI:%s>", self.class.name, object_id << 1, to_s)
+    end
+
+    private
+
+    # Process a URI host, adding or removing surrounding brackets if the host is an IPv6 address.
+    #
+    # @param [Boolean] brackets When true, brackets will be added to IPv6 addresses if missing. When
+    #   false, they will be removed if present.
+    #
+    # @return [String] Host with IPv6 address brackets added or removed
+    def process_ipv6_brackets(raw_host, brackets: false)
+      ip = IPAddr.new(raw_host)
+
+      if ip.ipv6?
+        brackets ? "[#{ip}]" : ip.to_s
+      else
+        raw_host
+      end
+    rescue IPAddr::Error
+      raw_host
     end
   end
 end
--- a/spec/lib/http/uri_spec.rb
+++ b/spec/lib/http/uri_spec.rb
@@ -1,9 +1,11 @@
 # frozen_string_literal: true
 
 RSpec.describe HTTP::URI do
+  let(:example_ipv6_address) { "2606:2800:220:1:248:1893:25c8:1946" }
+
   let(:example_http_uri_string)  { "http://example.com" }
   let(:example_https_uri_string) { "https://example.com" }
-  let(:example_ipv6_uri_string) { "https://[2606:2800:220:1:248:1893:25c8:1946]" }
+  let(:example_ipv6_uri_string) { "https://[#{example_ipv6_address}]" }
 
   subject(:http_uri)  { described_class.parse(example_http_uri_string) }
   subject(:https_uri) { described_class.parse(example_https_uri_string) }
@@ -35,14 +37,25 @@ RSpec.describe HTTP::URI do
   end
 
   describe "#host=" do
-    it "resets cached values for #host and #normalized_host" do
+    it "updates cached values for #host and #normalized_host" do
       expect(http_uri.host).to eq("example.com")
       expect(http_uri.normalized_host).to eq("example.com")
 
-      http_uri.host = "[2606:2800:220:1:248:1893:25c8:1946]"
+      http_uri.host = "[#{example_ipv6_address}]"
 
-      expect(http_uri.host).to eq("2606:2800:220:1:248:1893:25c8:1946")
-      expect(http_uri.normalized_host).to eq("2606:2800:220:1:248:1893:25c8:1946")
+      expect(http_uri.host).to eq(example_ipv6_address)
+      expect(http_uri.normalized_host).to eq(example_ipv6_address)
+    end
+
+    it "ensures IPv6 addresses are bracketed in the inner Addressable::URI" do
+      expect(http_uri.host).to eq("example.com")
+      expect(http_uri.normalized_host).to eq("example.com")
+
+      http_uri.host = example_ipv6_address
+
+      expect(http_uri.host).to eq(example_ipv6_address)
+      expect(http_uri.normalized_host).to eq(example_ipv6_address)
+      expect(http_uri.instance_variable_get(:@uri).host).to eq("[#{example_ipv6_address}]")
     end
   end
 


tarcieri:
A few thoughts on this:

1. It'd be good to find the spec for these sorts of bracketed addresses in URIs, so we know we're implementing them correctly. I've never seen them before and I'm not sure what the behavior is supposed to be (e.g. are they IPv6 only or do they also work with IPv4? Do DNS names work?)
2. The regex should match what the spec says about what's allowed inside the brackets, i.e. the inner capture group should be for an IPv6 address, and not just anything
3. It'd be good to do this handling at the time the URI is processed (i.e. as early as possible) so we don't have bracketed `host` strings flying around that need to be stripped in various places. Perhaps inside of the constructor for `lib/http/uri.rb`?

tarcieri:
Apparently the bracketed hosts are defined in [RFC3896 § 3.2.2](https://www.rfc-editor.org/rfc/rfc3986#section-3.2.2):

> A host identified by an Internet Protocol literal address, version 6
> [[RFC3513](https://www.rfc-editor.org/rfc/rfc3513)] or later, is distinguished by enclosing the IP literal
> within square brackets ("[" and "]").  This is the only place where
> square bracket characters are allowed in the URI syntax.

tarcieri:
@jeraki I would suggest changing `HTTP::URI` to eagerly strip the brackets from IPv6 addresses and caching it as a `@host` variable which the accessor methods can be changed to use

tarcieri:
I think it's fine to change the behavior. The current behavior is broken per #730, so changing it is a bugfix.

As to how to approach the setter and whether it should mutate the underlying `Addressable::URI` is a tricky question. I guess the setter should add brackets if `host=` receives a valid IPv6 address?

lib/http/uri.rb
```diff
@@ -110,6 +109,54 @@ def hash
...
+    #
+    # @param [String, #to_str] new_host The new host component.
+    # @return [void]
+    def host=(new_host)
+      @uri.host = new_host
+
+      # Reset dependent values
+      remove_instance_variable(:@host) if defined?(@host)
+      remove_instance_variable(:@normalized_host) if defined?(@normalized_host)
+    end
```
tarcieri, line 158: It seems like all of this could be done eagerly in the constructor rather than lazily which would simplify the implementation.

Also I'm not sure I see the value of having two methods. I'd suggest just always having `host` return the unbracketed version.

tarcieri: Aah, sorry, my bad. Offhand I don’t know what `normalized_host` does either

tarcieri:
@jeraki I mentioned [that particular issue earlier](https://github.com/httprb/http/pull/731#issuecomment-1333109827)

I suggested the setters need to use the bracketed form of IPv6 addresses with `Addressable::URI` so the URLs serialize correctly, while internally caching the host with the brackets stripped.

That's another thing I think would be easier with eagerly updating instance variables, rather than trying to do it lazily.

I think if you have one private helper method, called from both the constructor and the setter, which extracts the instance variables out of `Addressable::URI`, you can have fairly clean logic for this, and also DRY out the logic for checking if an address is an IPv6 address.

tarcieri:
Looks good now, thanks!

tarcieri:
Sure

cben:
Conceptually, it's weird having brackets in `host`.  They are not part of an [IPv6 address](https://www.rfc-editor.org/rfc/rfc3513) by itself; the brackets are merely part of the URL syntax for IPv6 (needed to distinguish the colons inside the address from :port number).

However, the Addressable gem makes the distinction that [`#hostname`](https://www.rubydoc.info/gems/addressable/Addressable%2FURI:hostname) is the address by itself vs. [`#host`](https://www.rubydoc.info/gems/addressable/Addressable%2FURI:host) is the bracketed-if-IPv6 part of the [`#authority`](https://www.rubydoc.info/gems/addressable/Addressable%2FURI:authority) portion of the URL.  
With that context, this is consistent :+1:.  
(But if you can switch to dealing with `hostname` (?) it might simplify things)

tarcieri:
@cben that sounds like a great point. Could one of you open a PR to switch to revert this and switch to `#hostname`?

PR #738, opened 2023-02-14
Fixed incorrect link in CHANGES.md

daniel-sabourin:
Thanks everyone to their contributions to this gem. Really appreciate all the effort put in.

I've been in the middle of upgrading from 4.x to 5.x and I noticed an incorrect link in the change log. Fixed so it points to the correct link


tarcieri:
Thanks!

PR #742, opened 2023-03-24
Run CI against Ruby 3.2

koheisg:


spec/lib/http/options/headers_spec.rb
```diff
@@ -14,7 +14,11 @@
   end
 
   it "accepts any object that respond to :to_hash" do
-    x = Struct.new(:to_hash).new("accept" => "json")
+    x = if RUBY_VERSION >= "3.2.0"
+          Data.define(:to_hash).new(:to_hash => { "accept" => "json" })
+        else
+          Struct.new(:to_hash).new({ "accept" => "json" })
+        end
```
koheisg, line 21: `Struct.new(:to_hash, keyword_init: false).new("accept" => "json")`

If this is the case, the code will not branch by version, but keyword_init: false will cause rubocop to say `Use hash rockets syntax.

koheisg:
```suggestion
    Struct.new(:to_hash, keyword_init: false).new("accept" => "json")
```

`Struct.new(:to_hash, keyword_init: false).new("accept" => "json")`

If this is the case, the code will not branch by version, but keyword_init: false will cause rubocop to say `Use hash rockets syntax.

tarcieri:
@koheisg a separate issue, but it's probably about time we changed from hashrockets to keyword args in general

1mmt-kohei-sugi:
@tarcieri OK ruby 3.2 support first! Thanks.

koheisg:
@tarcieri OK ruby 3.2 support first! Thanks.

spec/lib/http/options/headers_spec.rb
tarcieri: I think we should migrate from hashrockets to keyword args in general, and that's long overdue.

I would suggest disabling the rubocop lint for hashrockets.

PR #743, opened 2023-03-26
Add `Connection#finished_request?`

c960657:
When using persistent connections there is no easy way to check if the entire body of the previous request has been consumed before sending a new request, e.g. if an exception was raised while reading the body.

Add `Connection#finished?` which does the same check as the guards in `Connection#send_request`.

https://github.com/httprb/http/blob/32a2ae68a1650d267f0687381d58880cf1b60315/lib/http/connection.rb#L69-L76

### DIFF
--- a/lib/http/connection.rb
+++ b/lib/http/connection.rb
@@ -132,7 +132,7 @@ module HTTP
       @pending_request  = false
     end
 
-    def finished?
+    def finished_request?
       !@pending_request && !@pending_response
     end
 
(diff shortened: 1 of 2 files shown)


lib/http/connection.rb
```diff
@@ -132,6 +132,10 @@ def close
       @pending_request  = false
     end
 
+    def finished?
```
tarcieri, line 135: I think `finished?` is a little ambiguous, since it could mean the connection itself rather than a request. How about...
```suggestion
    def finished_request?
```

tarcieri:
Thanks!

PR #744, opened 2023-03-29
Bump llhttp-ffi (0.5.0)

bryanp:
Updates the `llhttp-ffi` dependency to `v0.5.0`. The only difference is that it now uses `lhttp v8.1.0`, the latest release.


bryanp:
Note I am seeing one test failure on this branch but also see it in `main`:

```
1) HTTP::Options headers accepts any object that respond to :to_hash
     Failure/Error: x = Struct.new(:to_hash).new("accept" => "json")

     ArgumentError:
       unknown keywords: accept
     # ./spec/lib/http/options/headers_spec.rb:17:in `initialize'
     # ./spec/lib/http/options/headers_spec.rb:17:in `new'
     # ./spec/lib/http/options/headers_spec.rb:17:in `block (2 levels) in <top (required)>'
```

PR #754, opened 2023-06-08
Stricter timeout options parsing

### LINKED ISSUE #752: raise error when invalid hash is passed to timeout
Hi, love httprb.

I would like to suggest to add some validation of the arguments to timeout. I can see that there are already some but we managed to pass `{timeout: 5}` as the argument

This is the related code. 
https://github.com/httprb/http/blob/462d711040735be3661068e9e52eb14a4269f1f4/lib/http/chainable.rb#L93-L112

I can open a PR of you would want that?

stoivo:
fix #752


lib/http/timeout/per_operation.rb
```diff
@@ -19,6 +20,38 @@ def initialize(*args)
...
+          end
+
+          raise ArgumentError, "at least one option" if options.empty?
+
+          options
+        end
+
+        private
+
+        def undo_short_names(options)
```
ixti, line 40: WDYT about naming it: `expand_names`?

ixti (changes requested):

lib/http/timeout/per_operation.rb
stoivo: 🤦 very smart.

lib/http/chainable.rb
```diff
@@ -90,21 +90,18 @@ def build_request(*args)
     # @overload timeout(global_timeout)
     #   Adds a global timeout to the full request
     #   @param [Numeric] global_timeout
+    PER_OPERATION_KEYS = Set.new(%i[read write connect])
```
tycooon, line 93: This looks unused :)

stoivo: moved in into lib/http/timeout/per_operation.rb but didn't remove it from here

stoivo:
What are we waiting for?

tarcieri:
@stoivo the tests are red, for one thing

stoivo:
Pushed with no changed and now spec are green

lib/http/timeout/per_operation.rb
```diff
@@ -11,6 +11,40 @@ class PerOperation < Null
...
+        end
+
+        private
+
+        def expand_names(options)
+          %i[read write connect].each do |k|
+            next unless options.key? k
+
+            if options.key?("#{k}_timeout".to_sym)
+              raise ArgumentError, "can't pass both #{k} and #{"#{k}_timeout".to_sym}"
```
ixti, line 38:
```suggestion
              raise ArgumentError, "can't pass both #{k} and #{k}_timeout"
```

lib/http/timeout/per_operation.rb
```diff
@@ -11,6 +11,40 @@ class PerOperation < Null
       WRITE_TIMEOUT = 0.25
       READ_TIMEOUT = 0.25
 
+      SETTINGS = Set.new(%i[read_timeout write_timeout connect_timeout])
+
+      class << self
+        def parse_options(options)
+          options = options.dup.then { |opts| expand_names(opts) }
```
ixti, line 18:
```suggestion
          options = expand_names(options.dup)
```

lib/http/timeout/per_operation.rb
```diff
@@ -11,6 +11,40 @@ class PerOperation < Null
...
+          options.each do |key, value|
+            unless SETTINGS.member?(key) && value.is_a?(Numeric)
+              raise ArgumentError, "invalid option #{key.inspect}, must be numeric " \
+                                   "`.timeout(connect: x, write: y, read: z)`."
+            end
+          end
+
+          raise ArgumentError, "at least one option" if options.empty?
+
+          options
```
ixti, line 28: I believe we also should fail if given options are mistyped or totally wrong ones

ixti (changes requested):
After taking a closer look, I think we should do parsing a bit differently:

``` ruby
KEYS = %i[read write connect].to_h { |k| [k, :"#{k}_timeout"] }.freeze

class << self
  def normalize_options(options)
    normalized = {}
    original   = options.dup

    KEYS.each do |short, long|
      if original.key?(short) && original.key(long)
        raise ArgumentError, "can't pass both #{short} and #{long}"
      end

      normalized[long] = original.key?(long) ? original.delete(long) : original.delete(short)
      raise ArgumentError, "#{long} must be numeric", unless normalized[long].is_a?(Numeric)
    end

    if original.size.positive?
      raise ArgumentError, "unknown timeout options: #{original.keys.join(', ')}"
    end

    if normalized.empty?
      raise ArgumentError, "no timeout options given"
    end

    normalized
  end
end
```

stoivo:
Very nice suggestion. So much more strait forward. It looks like it sneaked in a `,` in `raise ArgumentError, "#{long} must be numeric", unless normalized[long].is_a?(Numeric)` 
> After taking a closer look, I think we should do parsing a bit differently:
> 
> ```ruby
> KEYS = %i[read write connect].to_h { |k| [k, :"#{k}_timeout"] }.freeze
> 
> class << self
>   def normalize_options(options)
>     normalized = {}
>     original   = options.dup
> 
>     KEYS.each do |short, long|
>       if original.key?(short) && original.key(long)
>         raise ArgumentError, "can't pass both #{short} and #{long}"
>       end
> 
>       normalized[long] = original.key?(long) ? original.delete(long) : original.delete(short)
>       raise ArgumentError, "#{long} must be numeric", unless normalized[long].is_a?(Numeric)
>     end
> 
>     if original.size.positive?
>       raise ArgumentError, "unknown timeout options: #{original.keys.join(', ')}"
>     end
> 
>     if normalized.empty?
>       raise ArgumentError, "no timeout options given"
>     end
> 
>     normalized
>   end
> end
> ```

You should be able to only set one of the actions callbacks like ```HTTP.timeout :read => 125```?

stoivo:
I like how it looks but rubocop this it is too high complexity. I think it's nice
```
normalize_options instead of parse_options

rubocop offences
be rubocop -a                                                                                                                                                                                                               15:37:07
lib/http/timeout/per_operation.rb:17:9: C: Metrics/AbcSize: Assignment Branch Condition size for normalize_options is too high. [<5, 23, 9> 25.2/17]
        def normalize_options(options) ...
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
lib/http/timeout/per_operation.rb:17:9: C: Metrics/CyclomaticComplexity: Cyclomatic complexity for normalize_options is too high. [10/7]
        def normalize_options(options) ...
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
lib/http/timeout/per_operation.rb:17:9: C: Metrics/MethodLength: Method has too many lines. [11/10]
        def normalize_options(options) ...
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
lib/http/timeout/per_operation.rb:17:9: C: Metrics/PerceivedComplexity: Perceived complexity for normalize_options is too high. [10/8]
        def normalize_options(options) ...
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 80/80 files 
80 files inspected, 4 offenses detected

```

stoivo:
@ixti ping

ixti:
oh. sorry, totally forgot. I'm afk till Monday, will merge as soon as get back

ixti:
I've added a commit that disables inline some cops for that method, but I believe I found a regression. Right now it's possible to set global timeout with:

``` ruby
HTTP.timeout({ global: 161 })
```

And now that I'm looking at all of this again, I think next major we should add breaking change and:

* change `#timeout` signature to `#timeout(type, options = nil)`
* `type` will be one of: `:none`, `:global`, `:per_operation`
* signature of `Global#initialize` should be `(value)`
* signature of `PerOperation#initialize` should be `(connect: 0, write: 0, read: 0)`

@tarcieri WDYT?

tarcieri:
Sounds fine to me although it would be a good time to consider potential other breaking changes too

misalcedo:
I was comparing this HTTP client to others in Ruby and one thing I would like to add for timeouts is that global and per-operation are not mutually exclusive. I may want a request to time out after 1 minute, but spend no more than 150ms on the TCP connection portion or spend no more than 10 seconds idle on a read or write operation.

An alternative interpretation is what HTTPX does, which is to treat read as a timeout reading the response, write as a timeout writing the request and global as a overall timeout.

In my experience, except for streaming, HTTP clients care about total request timeout and maybe connect timeout, but definitely not per-operation timeouts. Streaming cares about operation timeouts and only optionally cares about a total timeout.

Often clients measure time to first byte from the server as well as overall latency, so timeouts on both of those could be helpful.

I am happy to help implement this type of behavior for 6.X if that is the target version.

tarcieri:
@misalcedo I opened #773 to track a potential redesign of the timeout subsystem for `http` 6.x

tarcieri:
@stoivo any chance you could rebase this and get the tests green? We'd like to get it merged before a v6.0.0 release

stoivo:
I will try to take a look at this this week

stoivo:
I am sorry. I was motivated to fix this when I opened the MR. I have since moved on to use httpx and faraday. I don't have the extra time to look into this.

PR #760, opened 2023-08-28
Allow for copy paste to work

hasandiwan:
Eliminate the "$" executing bundle

### DIFF
--- a/README.md
+++ b/README.md
@@ -38,7 +38,7 @@ gem "http"
 
 And then execute:
 ```bash
-$ bundle
+bundle
 ```
 
 Or install it yourself as:


tarcieri:
The README contains a mixture of both shell and Ruby code examples. Including the prompts helps visually disambiguate which ones are which, even if it makes copy/paste slightly harder.

Also, there are multiple shell examples and this proposal only covers one of them, so at best it's inconsistent.

ixti:
I find it's better to have `$` prefix for shell commands. That follows common practice: `$ some_command` means running `some_command` in shell as user, and `# some_command` as root.

This section is also pretty standard, as it's the default README of gem skeleton provided by bundler.

With that being said, I do think it's better to let shell commands to standout from ruby code. Removing `$` will eventually lead to somebody asking to bring it back because they thought it was supposed to be copy-pasted to IRB, or Gemfile, or such.

ixti:
We can update it to be smaller though, I believe people working with Ruby know what Gemfile is and how to modify it manually, so we can follow latest bundler's template:

``` md
## Installation

Install the gem and add to the application's Gemfile by executing:

    $ bundle add http

If bundler is not being used to manage dependencies, install the gem by executing:

    $ gem install http
```

PR #765, opened 2023-09-27
Prevent CRLF injection due to broken URL normalizer

c960657:
If a custom URL normaliser does not filter out white-space, raise an exception to prevent CRLF injection. Whitespace, in particular `\n`, is never allowed in path, query or hostname.

This is not a security fix as such but an extra line of defence.

### DIFF
--- a/lib/http/request.rb
+++ b/lib/http/request.rb
@@ -169,10 +169,12 @@ module HTTP
     def headline
       request_uri =
         if using_proxy? && !uri.https?
-          uri.omit(:fragment)
+          uri.omit(:fragment).to_s
         else
-          uri.request_uri
+          uri.request_uri.to_s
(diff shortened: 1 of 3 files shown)


lib/http/request.rb
```diff
@@ -169,11 +169,13 @@ def connect_using_proxy(socket)
     def headline
       request_uri =
         if using_proxy? && !uri.https?
-          uri.omit(:fragment)
+          uri.omit(:fragment).to_s
         else
-          uri.request_uri
+          uri.request_uri.to_s
         end
```
tarcieri, line 175: It seems like this could be DRY'd out by calling `.to_s` on `end`

ixti (approved):
LGTM

PR #766, opened 2023-10-06
 Handle responses in the reverse order from the requests

souk4711:
Resolve https://github.com/httprb/http/issues/767

BREAKING CHANGE:

* multiple features execution depends on order, e.g. `HTTP.use(:auto_inflate, :logging)`
* [replay function in webmock](https://github.com/bblimke/webmock/blob/63940ac95345968cf435820a62102d913441862c/lib/webmock/http_lib_adapters/http_rb/webmock.rb#L46).

### DIFF
--- a/lib/http/client.rb
+++ b/lib/http/client.rb
@@ -81,7 +81,7 @@ module HTTP
       end
       res = build_response(req, options)
 
-      res = options.features.inject(res) do |response, (_name, feature)|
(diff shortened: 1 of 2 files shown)


ixti:
@tarcieri we have a bunch of breaking changes that we can ship as major release.

tarcieri:
Yeah that sounds fine

PR #768, opened 2023-10-14
Start 6.0.0 version

ixti:
* Drop Ruby 2.6 support
* Drop Ruby 2.7 support
* Update rubocop


ixti:
I don't know why rubocop fails - it passes locally o_O

ixti:
I think we can switch away from hash rocket (even though I prefer it, lately I learned to love new Hash style), and switch to kwargs instead of options Hashes with major bump.

PR #769, opened 2023-10-18
Goodbye Hash rockets...

ixti:
I love Hash rockets. And think that it helps distinguish Hashes from keyword arguments, but:

1. we should avoid using options Hashes in arguments
2. lately I find it harder to deal with conditional logic of how Ruby differentiates keyword args from Hashes - so IMO it's better to move towards style that will imply we're using keyword arguments


paul (approved):
:heart:

tarcieri (approved):

PR #778, opened 2024-02-08
Drop depenency on base64

Earlopain:
#759 added `base64` as a depenency in order to suppress a warning/error on Ruby 3.3/3.4

Since `base64` is mostly just a wrapper around the `pack`/`unpack` methods it's easy to just replace.

Other gems have taken a similar approach:
* https://github.com/rubocop/rubocop/pull/12313
* https://github.com/rack/rack/pull/2110
* https://github.com/elastic/elasticsearch-ruby/pull/2295
* https://github.com/socketry/protocol-http/pull/51
* etc.


lib/http/chainable.rb
```diff
@@ -213,9 +211,9 @@ def auth(value)
     def basic_auth(opts)
       user  = opts.fetch(:user)
       pass  = opts.fetch(:pass)
-      creds = "#{user}:#{pass}"
+      strict_base64_encoded_creds = ["#{user}:#{pass}"].pack("m0")
```
ixti, line 214: We probably can introduce `HTTP::Utils` (don't like `Utils` modules in general, but it seems like fitting in this case)` with `encode_base64` method to avoid repeating `pack`. That should make code a bit better.

tarcieri: Or how about `HTTP::Encoding`, where there are probably a few other things you could stick there to provide a nicer facade for various bits in the standard library

ixti:
Oh. Yeah. Encoding seems better.

lib/http/chainable.rb
ixti: Oh, yeah! `HTTP::Encoding` is a much better namespace.

Earlopain: While trying to implement this suggestion I stumbled upon a conflict with the build-in `Encoding` constant. Each usage of the building encoding would need to be prefixed with `::` to make ruby look in the global namespace, in addition to required changes in other gems of the ecosystem like `http-form-data`.

I've went with `Encoder` instead which doesn't have such issues.

lib/http/encoder.rb
```diff
@@ -0,0 +1,10 @@
+# frozen_string_literal: true
+
+module HTTP
+  module Encoder
+    # Equivalent to Base64.strict_encode64
+    def self.to_base64(input)
+      [input].pack("m0")
+    end
+  end
```
ixti, line 9:
```suggestion
  module Encoding
    module_function
    
    # @param input [String]
    # @return [String] RFC 4648 compliant Base64 encoded string
    def encode_base64(input)
      [input].pack("m0")
    end
  end
```

ixti (changes requested):
Please name the module `Encoding`, make functions `module_function` (so that it can be included and used directly), and name the method `encode_base64`.

lib/http/chainable.rb
ixti: Just saw this comment. Hm, I don't like `Encoder` name though. Probably we can go with `Base64` LOL:

``` ruby
module HTTP
  module Base64
    module_function

    def encode64(input)
      [input].pack("m0")
    end
  end

  class Request
    include Base64

    ...

      digest = encode64("#{proxy[:proxy_username]}:#{proxy[:proxy_password]}")

    ...
  end
end


```

ixti: @tarcieri WDYT?

Earlopain: This does get the tests all passing, however still suffer from the same (though now a bit more theoretical) problem. Other gems that extend the HTTP namespace would now get a different constant when acessing plain `Base64`.

Luckily the test suite exposed the problem with using `Encoding`. I can't say if that would be an issue with `Base64`.

tarcieri (approved):
This now looks to be working aside from the rubocop lints which I fixed in #788

Issue #674, opened 2021-06-22
Incorrect response headers give a generic message instead of indicating a problem parsing the headers


fhoeben:
We are sending a request to a server that is sending incomplete response headers (i.e. last line of headers does not contain a `:`).

```
HTTP/1.1 200 OK
Access-Control-Allow-Origin: *
Access-Control-Allow-Headers: Authorization, Content-Type, Accept, X-User-Email, X-Auth-Token
Allow: GET, POST, OPTIONS
Content-Type: text/html;charset=utf-8
Content-Length: 0
X-XSS-Protection: 1; mode=block
X-Content-Type-Options: nosniff
X-Frame-Options: SAMEORIGIN
Connection: keep-alive
ver


```

As a result we get a `ConnectionError` telling us 'error reading from socket: Could not parse data' (from `Connection.read_more()`). It would have been very helpful if we would get a message stating that the response headers were incorrect (and `Connection.read_headers!` which calls `read_more` does seem to indicate that in some cases).

tarcieri:
For clarity, you're saying in this case that the `\r\n\r\n` delimiter between the headers and body is there, it's just there's one malformed header at the end? (`ver`)

fhoeben:
Yes. (and the actual body is empty)

fhoeben:
I forgot to mention we are on version 4.4.1

tarcieri:
Aah, we completely changed parsers in v5.0, so I'd be curious if this problem persists

fhoeben:
I just tried that, but my complaints about their broken headers are being taken seriously. They are trying to fix but currently the server is not responding at all...

I'll let you know if I can get a result with 5.0.0

PR #783, opened 2024-05-07
Add more granularity to `ConnectionError` through more specific error classes

omkarmoghe:
- would love to have more granularity into what kind of Connection error is being thrown
- this PR changes 3/4 `ConnectionError` throws to use subclasses specific to that failure mode
- this is backwards compatible because they are subclasses of `ConnectionError` so this should not affect any existing `rescue` logic
- open to feedback!


omkarmoghe:
somewhat related to #674

tarcieri (approved):

PR #789, opened 2024-08-23
Cache header normalization to reduce object allocation

alexcwatt:
I'm using this gem to make a lot of requests in a multi-threaded environment. I've noticed that header normalization contributes to a large percentage of the overall objects allocated.

I expect that most services will come across some of the same headers again and again. In my case, I haven't counted the different headers that are encountered in production, but I only issue traffic to the same service so I am confident that I'm parsing the same headers many times.

This PR introduces a cache for header normalization to reduce object allocation. Here's a benchmark script to validate the improvement (you can put this in the root of the repository as bench.rb, then run with `bundle exec ruby bench.rb` for example):

```ruby
require 'bundler/inline'

gemfile do
  source 'https://rubygems.org'
  gem "benchmark-ips"
  gem "benchmark-memory"
  gem "http", path: "."
end

require "http"

require "benchmark/ips"
require "benchmark/memory"

puts "== Benchmark speed"
Benchmark.ips do |x|
  x.report("HTTP::Headers#add") {
    headers = HTTP::Headers.new
    headers.add("content_type", "application/json")
    headers.add("some-other-header", "some-value")
    headers.add("another-header", "another-value")
    headers.add("yet-another-header", "yet-another-value")
  }
end

puts "\n== Benchmark memory"
Benchmark.memory do |x|
  x.report("HTTP::Headers#add same header") {
    headers = HTTP::Headers.new
    headers.add("content_type", "application/json")
    headers.add("content_type", "application/json")
    headers.add("content_type", "application/json")
    headers.add("content_type", "application/json")
  }

  x.report("HTTP::Headers#add different header") {
    headers = HTTP::Headers.new
    headers.add("content_type", "application/json")
    headers.add("some-other-header", "some-value")
    headers.add("another-header", "another-value")
    headers.add("yet-another-header", "yet-another-value")
  }
end
```

### Results on main

```
== Benchmark speed
ruby 3.2.1 (2023-02-08 revision 31819e82c8) [x86_64-darwin22]
Warming up --------------------------------------
   HTTP::Headers#add     7.672k i/100ms
Calculating -------------------------------------
   HTTP::Headers#add     76.731k (± 6.1%) i/s -    383.600k in   5.024653s

== Benchmark memory
Calculating -------------------------------------
HTTP::Headers#add same header
                         2.832k memsize (     0.000  retained)
                        50.000  objects (     0.000  retained)
                         5.000  strings (     0.000  retained)
HTTP::Headers#add different header
                         3.648k memsize (    80.000  retained)
                        58.000  objects (     1.000  retained)
                        19.000  strings (     1.000  retained)
```

### Results on this branch

```
== Benchmark speed
ruby 3.2.1 (2023-02-08 revision 31819e82c8) [x86_64-darwin22]
Warming up --------------------------------------
   HTTP::Headers#add    30.330k i/100ms
Calculating -------------------------------------
   HTTP::Headers#add    299.935k (± 2.7%) i/s -      1.516M in   5.059825s

== Benchmark memory
Calculating -------------------------------------
HTTP::Headers#add same header
                         1.040k memsize (     0.000  retained)
                        18.000  objects (     0.000  retained)
                         2.000  strings (     0.000  retained)
HTTP::Headers#add different header
                         1.040k memsize (    80.000  retained)
                        18.000  objects (     1.000  retained)
                         8.000  strings (     1.000  retained)
```

### Analysis

I wanted to use `Benchmark.ips` to make sure what I'm adding doesn't regress performance. This PR improves performance on the benchmark because a single `HeaderNormalizer` instance is memoized in the `Header` class and reused for all the benchmarks, so once we've normalized a header once, there is no additional work the next time.

However, what I really care about is the object allocations. We can see that this PR increases the retained memory, as expected, since we are now persisting header normalized values in a cache. But the win here is the reduction in objects allocated.

### Follow up

I just wanted to get this out for your feedback! Please let me know what you think and whether you're open to something like this in the gem. If not, perhaps we can explore a pattern that would make it easier for me to extend the gem...


lib/http/header_normalizer.rb
```diff
@@ -0,0 +1,87 @@
+# frozen_string_literal: true
+
+module HTTP
+  class HeaderNormalizer
```
tarcieri, line 4: Perhaps stick this under `http/headers/normalizer.rb`?

lib/http/header_normalizer.rb
```diff
@@ -0,0 +1,87 @@
...
+      return normalized if COMPLIANT_NAME_RE.match?(normalized)
+
+      raise HeaderError, "Invalid HTTP header field name: #{name.inspect}"
+    end
+
+    class LRUCache
+      def initialize(max_size)
+        @max_size = max_size
+        @cache = {}
+        @order = []
```
tarcieri, line 45: Ruby hashes are already ordered in terms of insertion order. You don't need to maintain a separate data structure to preserve the order.

lib/http/header_normalizer.rb
alexcwatt: Right! I moved it over there.

lib/http/header_normalizer.rb
alexcwatt: I pushed a commit; let me know what you think. The change I pushed is no longer an LRU cache; it's a cache where the oldest item based on insertion order gets removed when the cache overflows. It's therefore a bit simpler - maybe less efficient in some cases but it certainly works well for the case I'm optimizing for.

tarcieri: That looks great! Was there any impact on performance?

tarcieri: I guess to benchmark against an LRU you'd need something simulating an actual workload. An LRU may be a better tool for the job.

alexcwatt: The change makes it about 35% faster on my Intel-based MacBook! I just refreshed the benchmarks in the description, using this laptop. (Yesterday I was on my work computer which has M3 chip and all the numbers were faster, but I refreshed all the benchmarks to be apples-to-apples, no pun intended.)

alexcwatt: > I guess to benchmark against an LRU you'd need something simulating an actual workload. An LRU may be a better tool for the job.

Yeah I do agree with that. I am happy to revert the LRUCache -> Cache commit if you think it was better before. I can also tweak the implementation if you'd prefer I not use the array and instead use the Hash to maintain ordering and do something like delete the element from the hash and re-add it.

tarcieri: If the initial benchmarks are promising, it would also be nice to have something lightweight to start with, and maybe circle back on an LRU if your real-world telemetry shows otherwise

tarcieri (approved):

lib/http/header_normalizer.rb
alexcwatt: Sounds good! I suspect that my system will not need LRU because it sees the same headers over and over.

ixti (commented):
I like the proposal in general (in fact was thinking about the simialr idea for a long time), but I think we should:

1. make cache size configurable
2. make it possible to provide custom cache implementation (e.g. LRU might be better here)

ixti:
Here's what I have in mind for LRU Cache:

``` ruby
# frozen_string_literal: true

module HTTP
  class Headers
    class Normalizer
      class LRUCache
        Node = Struct.new(:value, :pred, :succ, keyword_init: true) do
          def unlink
            pred.succ = succ if pred
            succ.pred = pred if succ
            self
          end
        end

        class List
          def initialize
            @head = Node.new
            @tail = @head
          end

          def push(node)
            # Do nothing if node is the tail itself
            unless @tail.equal?(node)
              @tail.succ = node.unlink
              @tail = node
            end

            self
          end

          def shift
            node = @head.succ&.unlink

            # reset tail if head was the only node in the list
            @tail = @head unless node

            node
          end
        end

        def initialize(max_size: 200)
          @max_size = max_size
          @list     = List.new
          @cache    = {}
        end

        def get(key)
          @cache[key]&.value
        end
        alias [] get

        def set(key, value)
          node = (@cache[key] ||= Node.new)

          @cache[key].value = value
          @list.push(node)

          @list.shift if @max_size < @cache.size

          value
        end
        alias []= set
      end
    end
  end
end
```

### Benchmarks

<details><summary>Benchmark code</summary>

``` ruby
# frozen_string_literal: true

require "bundler/inline"

gemfile do
  source 'https://rubygems.org'
  gem "benchmark-ips"
  gem "benchmark-memory"
  gem "http", path: "."
end

require "http"

require "benchmark/ips"
require "benchmark/memory"

NORMALIZER_NO_CACHE  = HTTP::Headers::Normalizer.new(cache: nil)
NORMALIZER_CACHE     = HTTP::Headers::Normalizer.new(cache: HTTP::Headers::Normalizer::Cache.new)
NORMALIZER_LRU_CACHE = HTTP::Headers::Normalizer.new(cache: HTTP::Headers::Normalizer::LRUCache.new)

HEADERS = [
  *Array.new(50, "content_type"),
  *Array.new(75, "api-key"),
  *Array.new(100, "ANOTHER-key")
].shuffle

def test_add(normalizer)
  headers = HTTP::Headers.new(normalizer: normalizer)
  HEADERS.each { |k| headers.add(k, "value does not matter") }
end

puts "== Benchmark speed"
Benchmark.ips do |x|
  x.report("HTTP::Headers#add (no cache)") { test_add(NORMALIZER_NO_CACHE) }
  x.report("HTTP::Headers#add (cache)") { test_add(NORMALIZER_CACHE) }
  x.report("HTTP::Headers#add (lru cache)") { test_add(NORMALIZER_LRU_CACHE) }

  x.compare!
end

puts "\n== Benchmark memory"
Benchmark.memory do |x|
  x.report("HTTP::Headers#add (no cache)") { test_add(NORMALIZER_NO_CACHE) }
  x.report("HTTP::Headers#add (cache)") { test_add(NORMALIZER_CACHE) }
  x.report("HTTP::Headers#add (lru cache)") { test_add(NORMALIZER_LRU_CACHE) }

  x.compare!
end
```

</details>

```
== Benchmark speed
ruby 3.3.4 (2024-07-09 revision be1089c8ec) [x86_64-linux]
Warming up --------------------------------------
HTTP::Headers#add (no cache)
                       243.000 i/100ms
HTTP::Headers#add (cache)
                       891.000 i/100ms
HTTP::Headers#add (lru cache)
                       907.000 i/100ms
Calculating -------------------------------------
HTTP::Headers#add (no cache)
                          2.463k (± 0.6%) i/s -     12.393k in   5.031792s
HTTP::Headers#add (cache)
                          8.877k (± 0.5%) i/s -     44.550k in   5.018856s
HTTP::Headers#add (lru cache)
                          9.042k (± 0.3%) i/s -     45.350k in   5.015557s

Comparison:
HTTP::Headers#add (lru cache):     9041.9 i/s
HTTP::Headers#add (cache):     8876.8 i/s - 1.02x  slower
HTTP::Headers#add (no cache):     2463.0 i/s - 3.67x  slower


== Benchmark memory
Calculating -------------------------------------
HTTP::Headers#add (no cache)
                       112.520k memsize (     0.000  retained)
                         2.029k objects (     0.000  retained)
                         8.000  strings (     0.000  retained)
HTTP::Headers#add (cache)
                        20.720k memsize (     0.000  retained)
                       454.000  objects (     0.000  retained)
                         0.000  strings (     0.000  retained)
HTTP::Headers#add (lru cache)
                        20.720k memsize (     0.000  retained)
                       454.000  objects (     0.000  retained)
                         0.000  strings (     0.000  retained)

Comparison:
HTTP::Headers#add (cache):      20720 allocated
HTTP::Headers#add (lru cache):      20720 allocated - same
HTTP::Headers#add (no cache):     112520 allocated - 5.43x more
```


### Patch

Here's full patch:

``` diff
diff --git a/lib/http/headers.rb b/lib/http/headers.rb
index 89ff78b..ac21a10 100644
--- a/lib/http/headers.rb
+++ b/lib/http/headers.rb
@@ -13,13 +13,49 @@ module HTTP
     extend Forwardable
     include Enumerable
 
+    INIT_MUTEX = Mutex.new
+
+    class << self
+      # Coerces given `object` into Headers.
+      #
+      # @raise [Error] if object can't be coerced
+      # @param [#to_hash, #to_h, #to_a] object
+      # @return [Headers]
+      def coerce(object)
+        unless object.is_a? self
+          object = case
+                   when object.respond_to?(:to_hash) then object.to_hash
+                   when object.respond_to?(:to_h)    then object.to_h
+                   when object.respond_to?(:to_a)    then object.to_a
+                   else raise Error, "Can't coerce #{object.inspect} to Headers"
+                   end
+        end
+
+        headers = new
+        object.each { |k, v| headers.add k, v }
+        headers
+      end
+      alias [] coerce
+
+      def normalizer
+        @normalizer || INIT_MUTEX.synchronize { @normalizer ||= Headers::Normalizer.new }
+      end
+
+      def normalizer=(normalizer)
+        raise ArgumentError, "normalizer must be callable" unless normalizer.respond_to?(:call)
+
+        INIT_MUTEX.synchronize { @normalizer = normalizer }
+      end
+    end
+
     # Class constructor.
-    def initialize
+    def initialize(normalizer: self.class.normalizer)
       # The @pile stores each header value using a three element array:
       #  0 - the normalized header key, used for lookup
       #  1 - the header key as it will be sent with a request
       #  2 - the value
-      @pile = []
+      @pile       = []
+      @normalizer = normalizer
     end
 
     # Sets header.
@@ -188,40 +224,11 @@ module HTTP
       dup.tap { |dupped| dupped.merge! other }
     end
 
-    class << self
-      # Coerces given `object` into Headers.
-      #
-      # @raise [Error] if object can't be coerced
-      # @param [#to_hash, #to_h, #to_a] object
-      # @return [Headers]
-      def coerce(object)
-        unless object.is_a? self
-          object = case
-                   when object.respond_to?(:to_hash) then object.to_hash
-                   when object.respond_to?(:to_h)    then object.to_h
-                   when object.respond_to?(:to_a)    then object.to_a
-                   else raise Error, "Can't coerce #{object.inspect} to Headers"
-                   end
-        end
-
-        headers = new
-        object.each { |k, v| headers.add k, v }
-        headers
-      end
-      alias [] coerce
-    end
-
     private
 
-    class << self
-      def header_normalizer
-        @header_normalizer ||= Headers::Normalizer.new
-      end
-    end
-
     # Transforms `name` to canonical HTTP header capitalization
     def normalize_header(name)
-      self.class.header_normalizer.normalize(name)
+      @normalizer.call(name)
     end
 
     # Ensures there is no new line character in the header value
diff --git a/lib/http/headers/normalizer.rb b/lib/http/headers/normalizer.rb
index 7c260d0..02c3fb2 100644
--- a/lib/http/headers/normalizer.rb
+++ b/lib/http/headers/normalizer.rb
@@ -1,5 +1,8 @@
 # frozen_string_literal: true
 
+require_relative "./normalizer/cache"
+require_relative "./normalizer/lru_cache"
+
 module HTTP
   class Headers
     class Normalizer
@@ -10,15 +13,18 @@ module HTTP
       # @see http://tools.ietf.org/html/rfc7230#section-3.2
       COMPLIANT_NAME_RE = /\A[A-Za-z0-9!#$%&'*+\-.^_`|~]+\z/
 
-      MAX_CACHE_SIZE = 200
-
-      def initialize
-        @cache = Cache.new(MAX_CACHE_SIZE)
+      def initialize(cache: Cache.new)
+        @resolve =
+          if cache
+            ->(name) { cache[name] ||= normalize_header(name) }
+          else
+            ->(name) { normalize_header(name) }
+          end
       end
 
       # Transforms `name` to canonical HTTP header capitalization
-      def normalize(name)
-        @cache[name] ||= normalize_header(name)
+      def call(name)
+        @resolve[name]
       end
 
       private
@@ -38,45 +44,6 @@ module HTTP
 
         raise HeaderError, "Invalid HTTP header field name: #{name.inspect}"
       end
-
-      class Cache
-        def initialize(max_size)
-          @max_size = max_size
-          @cache = {}
-        end
-
-        def get(key)
-          @cache[key]
-        end
-
-        def set(key, value)
-          @cache[key] = value
-
-          # Maintain cache size
-          return unless @cache.size > @max_size
-
-          oldest_key = @cache.keys.first
-          @cache.delete(oldest_key)
-        end
-
-        def size
-          @cache.size
-        end
-
-        def key?(key)
-          @cache.key?(key)
-        end
-
-        def [](key)
-          get(key)
-        end
-
-        def []=(key, value)
-          set(key, value)
-        end
-      end
-
-      private_constant :Cache
     end
   end
 end
diff --git a/lib/http/headers/normalizer/cache.rb b/lib/http/headers/normalizer/cache.rb
new file mode 100644
index 0000000..9d95019
--- /dev/null
+++ b/lib/http/headers/normalizer/cache.rb
@@ -0,0 +1,44 @@
+
+module HTTP
+  class Headers
+    class Normalizer
+      class Cache
+        def initialize(max_size: 200)
+          @max_size = max_size
+          @cache    = {}
+        end
+
+        def get(key)
+          @cache[key]
+        end
+
+        def set(key, value)
+          @cache[key] = value
+
+          # Maintain cache size
+          return unless @cache.size > @max_size
+
+          oldest_key = @cache.keys.first
+          @cache.delete(oldest_key)
+        end
+
+        def size
+          @cache.size
+        end
+
+        def key?(key)
+          @cache.key?(key)
+        end
+
+        def [](key)
+          get(key)
+        end
+
+        def []=(key, value)
+          set(key, value)
+        end
+      end
+    end
+  end
+end
diff --git a/lib/http/headers/normalizer/lru_cache.rb b/lib/http/headers/normalizer/lru_cache.rb
new file mode 100644
index 0000000..1fe25dc
--- /dev/null
+++ b/lib/http/headers/normalizer/lru_cache.rb
@@ -0,0 +1,66 @@
+
+module HTTP
+  class Headers
+    class Normalizer
+      class LRUCache
+        Node = Struct.new(:value, :pred, :succ, keyword_init: true) do
+          def unlink
+            pred.succ = succ if pred
+            succ.pred = pred if succ
+            self
+          end
+        end
+
+        class List
+          def initialize
+            @head = Node.new
+            @tail = @head
+          end
+
+          def push(node)
+            # Do nothing if node is the tail itself
+            unless @tail.equal?(node)
+              @tail.succ = node.unlink
+              @tail = node
+            end
+
+            self
+          end
+
+          def shift
+            node = @head.succ&.unlink
+
+            # reset tail if head was the only node in the list
+            @tail = @head unless node
+
+            node
+          end
+        end
+
+        def initialize(max_size: 200)
+          @max_size = max_size
+          @list     = List.new
+          @cache    = {}
+        end
+
+        def get(key)
+          @cache[key]&.value
+        end
+        alias [] get
+
+        def set(key, value)
+          node = (@cache[key] ||= Node.new)
+
+          @cache[key].value = value
+          @list.push(node)
+
+          @list.shift if @max_size < @cache.size
+
+          value
+        end
+        alias []= set
+      end
+    end
+  end
+end
```

ixti (dismissed):
If we're going to use cache, we should ensure that the normalized names in the cache are immutable:

``` ruby
name = -name.to_s
@cache[name] ||= -normalize_header(name)
```

Otherwise it might cause unpleasant surprise, when user uses something like:

``` ruby
response.headers.to_h.keys.each(&:upcase!)
```

ixti:
As I wrote above, we should freeze normalized values. That means we should keep the normalizer cache-less by default before the next major release; with note that the behaviour will change in a future release.

ixti:
Alternative to freeze, we can `dup` results (patch on top of a patch from above):

``` patch
diff --git a/lib/http/headers/normalizer.rb b/lib/http/headers/normalizer.rb
index 02c3fb2..15ec336 100644
--- a/lib/http/headers/normalizer.rb
+++ b/lib/http/headers/normalizer.rb
@@ -24,7 +24,7 @@ module HTTP
 
       # Transforms `name` to canonical HTTP header capitalization
       def call(name)
-        @resolve[name]
+        @resolve[-name.to_s].dup
       end
 
       private
```

It changes the output of the above benchmarks to:

```
== Benchmark speed
ruby 3.3.4 (2024-07-09 revision be1089c8ec) [x86_64-linux]
Warming up --------------------------------------
HTTP::Headers#add (no cache)
                        14.000 i/100ms
HTTP::Headers#add (cache)
                        27.000 i/100ms
HTTP::Headers#add (lru cache)
                        50.000 i/100ms
Calculating -------------------------------------
HTTP::Headers#add (no cache)
                        148.553 (± 4.0%) i/s -    756.000 in   5.096817s
HTTP::Headers#add (cache)
                        273.511 (± 4.0%) i/s -      1.377k in   5.042522s
HTTP::Headers#add (lru cache)
                        512.928 (± 2.1%) i/s -      2.600k in   5.071135s

Comparison:
HTTP::Headers#add (lru cache):      512.9 i/s
HTTP::Headers#add (cache):      273.5 i/s - 1.88x  slower
HTTP::Headers#add (no cache):      148.6 i/s - 3.45x  slower


== Benchmark memory
Calculating -------------------------------------
HTTP::Headers#add (no cache)
                         2.110M memsize (     0.000  retained)
                        35.626k objects (     0.000  retained)
                        50.000  strings (     0.000  retained)
HTTP::Headers#add (cache)
                         1.854M memsize (    17.560k retained)
                        16.167k objects (   222.000  retained)
                        50.000  strings (    50.000  retained)
HTTP::Headers#add (lru cache)
                       456.960k memsize (     0.000  retained)
                         9.604k objects (     0.000  retained)
                        50.000  strings (     0.000  retained)

Comparison:
HTTP::Headers#add (lru cache):     456960 allocated
HTTP::Headers#add (cache):    1854184 allocated - 4.06x more
HTTP::Headers#add (no cache):    2109800 allocated - 4.62x more
```

ixti:
My proposed patch has a couple of bugs though:

- forgot to actually remove elements from Hash
- LRU list is not very thread-safe :D

ixti:
After spectacular failure with LRU cache implementation (thread safe variant I could think of right now is way too slow), WDYT about this patch:

* Improve class structure consistency a bit (to follow ruby style guide)
* Remove unused methods from `Cache`
* Ensure cache keys are not modified
* Ensure cached values are not modified
* Truncate cache store before adding a new element

``` patch
diff --git a/lib/http/headers.rb b/lib/http/headers.rb
index 89ff78b..5f28561 100644
--- a/lib/http/headers.rb
+++ b/lib/http/headers.rb
@@ -13,6 +13,33 @@ module HTTP
     extend Forwardable
     include Enumerable
 
+    class << self
+      # Coerces given `object` into Headers.
+      #
+      # @raise [Error] if object can't be coerced
+      # @param [#to_hash, #to_h, #to_a] object
+      # @return [Headers]
+      def coerce(object)
+        unless object.is_a? self
+          object = case
+                   when object.respond_to?(:to_hash) then object.to_hash
+                   when object.respond_to?(:to_h)    then object.to_h
+                   when object.respond_to?(:to_a)    then object.to_a
+                   else raise Error, "Can't coerce #{object.inspect} to Headers"
+                   end
+        end
+
+        headers = new
+        object.each { |k, v| headers.add k, v }
+        headers
+      end
+      alias [] coerce
+
+      def normalizer
+        @normalizer ||= Headers::Normalizer.new
+      end
+    end
+
     # Class constructor.
     def initialize
       # The @pile stores each header value using a three element array:
@@ -188,40 +215,11 @@ module HTTP
       dup.tap { |dupped| dupped.merge! other }
     end
 
-    class << self
-      # Coerces given `object` into Headers.
-      #
-      # @raise [Error] if object can't be coerced
-      # @param [#to_hash, #to_h, #to_a] object
-      # @return [Headers]
-      def coerce(object)
-        unless object.is_a? self
-          object = case
-                   when object.respond_to?(:to_hash) then object.to_hash
-                   when object.respond_to?(:to_h)    then object.to_h
-                   when object.respond_to?(:to_a)    then object.to_a
-                   else raise Error, "Can't coerce #{object.inspect} to Headers"
-                   end
-        end
-
-        headers = new
-        object.each { |k, v| headers.add k, v }
-        headers
-      end
-      alias [] coerce
-    end
-
     private
 
-    class << self
-      def header_normalizer
-        @header_normalizer ||= Headers::Normalizer.new
-      end
-    end
-
     # Transforms `name` to canonical HTTP header capitalization
     def normalize_header(name)
-      self.class.header_normalizer.normalize(name)
+      self.class.normalizer.call(name)
     end
 
     # Ensures there is no new line character in the header value
diff --git a/lib/http/headers/cache.rb b/lib/http/headers/cache.rb
new file mode 100644
index 0000000..d8274dc
--- /dev/null
+++ b/lib/http/headers/cache.rb
@@ -0,0 +1,28 @@
+
+module HTTP
+  class Headers
+    # @private
+    # Normalized header names cache
+    class Cache
+      MAX_SIZE = 200
+
+      def initialize
+        @store = {}
+      end
+
+      def get(key)
+        @store[key]
+      end
+      alias [] get
+
+      def set(key, value)
+        # Maintain cache size
+        @store.delete(@store.each_key.first) while MAX_SIZE <= @store.size
+
+        @store[key] = value
+      end
+      alias []= set
+    end
+  end
+end
diff --git a/lib/http/headers/normalizer.rb b/lib/http/headers/normalizer.rb
index 7c260d0..95c8871 100644
--- a/lib/http/headers/normalizer.rb
+++ b/lib/http/headers/normalizer.rb
@@ -1,5 +1,7 @@
 # frozen_string_literal: true
 
+require_relative "./cache"
+
 module HTTP
   class Headers
     class Normalizer
@@ -10,15 +12,18 @@ module HTTP
       # @see http://tools.ietf.org/html/rfc7230#section-3.2
       COMPLIANT_NAME_RE = /\A[A-Za-z0-9!#$%&'*+\-.^_`|~]+\z/
 
-      MAX_CACHE_SIZE = 200
+      NAME_PARTS_SEPARATOR_RE = /[\-_]/
 
       def initialize
-        @cache = Cache.new(MAX_CACHE_SIZE)
+        @cache = Cache.new
       end
 
       # Transforms `name` to canonical HTTP header capitalization
-      def normalize(name)
-        @cache[name] ||= normalize_header(name)
+      def call(name)
+        name  = -name.to_s
+        value = (@cache[name] ||= -normalize_header(name))
+
+        value.dup
       end
 
       private
@@ -32,51 +37,11 @@ module HTTP
       def normalize_header(name)
         return name if CANONICAL_NAME_RE.match?(name)
 
-        normalized = name.split(/[\-_]/).each(&:capitalize!).join("-")
-
+        normalized = name.split(NAME_PARTS_SEPARATOR_RE).each(&:capitalize!).join("-")
         return normalized if COMPLIANT_NAME_RE.match?(normalized)
 
         raise HeaderError, "Invalid HTTP header field name: #{name.inspect}"
       end
-
-      class Cache
-        def initialize(max_size)
-          @max_size = max_size
-          @cache = {}
-        end
-
-        def get(key)
-          @cache[key]
-        end
-
-        def set(key, value)
-          @cache[key] = value
-
-          # Maintain cache size
-          return unless @cache.size > @max_size
-
-          oldest_key = @cache.keys.first
-          @cache.delete(oldest_key)
-        end
-
-        def size
-          @cache.size
-        end
-
-        def key?(key)
-          @cache.key?(key)
-        end
-
-        def [](key)
-          get(key)
-        end
-
-        def []=(key, value)
-          set(key, value)
-        end
-      end
-
-      private_constant :Cache
     end
   end
 end
```

PS: I think for specs we should add `rspec-memory` to test objects allocations instead of cache size

alexcwatt:
@ixti Thank you for the review and ideas! I took your patch and updated the tests, including:

* Added a test to ensure that each `#call` with the same value returns a different object
* Added a test to ensure that `#call` results aren't frozen
* Added two tests using `rspec-memory`
* I still like the idea of a test that explicitly checks what happens when the cache is full, so I have some version of that still

Please let me know what you think

ixti (approved):
Awesome! Thank you!

tarcieri:
@ixti might be good to cut a release. There are a few merged PRs that haven't been released yet

alexcwatt:
Thanks for the help here @ixti and the really fast reviews and feedback!

A release would be awesome - I'd love to try this in production.

Issue #776, opened 2024-01-20
Fix Base64 dependency warning


bkuhlmann:
## Why

Hello. :wave: I'm seeing the following warnings when using this gem with Ruby 3.3.0:

> warning: base64 was loaded from the standard library, but will no longer be part of the default gems since Ruby 3.4.0. Add base64 to your Gemfile or gemspec. Also contact author of http-5.1.1 to add base64 into its gemspec.

## How

Here's a quick script to reproduce:

``` 
#! /usr/bin/env ruby
# frozen_string_literal: true

# Save as `demo`, then `chmod 755 demo`, and run as `./demo`.

require "bundler/inline"

gemfile true do
  source "https://rubygems.org"
  gem "http"
end
```

Running the above will produce the warning mentioned above. The warning is due to this file (`http-5.1.1/lib/http/chainable.rb:3`) requiring `base64`.

## Notes

The fix, add the following to the gemspec:

``` 
spec.add_dependency "base64", "~> 0.2"
```

Here's the environment I'm using:

- **macOS**: macOS 13.6.3
- **Ruby**: ruby 3.3.0 (2023-12-25 revision 5124f9ac75) +YJIT [arm64-darwin22.6.0]
- **HTTP**: 5.1.1

ixti:
It was fixed some time ago in the main branch. I will cut a release today.

bkuhlmann:
Thanks, that'd be great!

ixti:
Released as 5.2.0

ixti:
Please see 5.2.0 changelog for potentially breaking changes: https://github.com/httprb/http/blob/5-x-stable/CHANGELOG.md

Issue #773, opened 2023-12-13
Timeout redesign for 6.x


tarcieri:
This is a tracking issue for redesigning the timeout subsystem in `http` 6.x.

It would be good to come up with a new design first, and then PRs implementing it.

The largest single issue seems to be having a cumulative timeout over the entire request which actually works (originally requested in #562).

Several people have expressed interest in working on this, and I believe some code has already been written, although isn't available in a public PR.

cc @ClearlyClaire @misalcedo

misalcedo:
> I was comparing this HTTP client to others in Ruby and one thing I would like to add for timeouts is that global and per-operation are not mutually exclusive. I may want a request to time out after 1 minute, but spend no more than 150ms on the TCP connection portion or spend no more than 10 seconds idle on a read or write operation.
> 
> An alternative interpretation is what HTTPX does, which is to treat read as a timeout reading the response, write as a timeout writing the request and global as a overall timeout.
> 
> In my experience, except for streaming, HTTP clients care about total request timeout and maybe connect timeout, but definitely not per-operation timeouts. Streaming typically cares about operation timeouts and only optionally cares about a total timeout.
> 
> Often clients measure time to first byte from the server as well as overall latency, so timeouts on both of those could be helpful.
> 
> I am happy to help implement this type of behavior for 6.X if that is the target version.

Quoting my thoughts from https://github.com/httprb/http/pull/754#issuecomment-1853966126 here.

misalcedo:
The useful stages of the request lifecycle to me are:

- connect
- write (request)
  - write (socket)
- read (response)
  - first byte
  - parse the response
  - read (socket)
- total time for a single request-response interaction

I don't see a benefit to mutually exclusive timeouts as timeouts can just be a min-heap (or array sorted in descending order) of deadlines before the client gives up.

As for guarantees, this depends on how often we check the current time against the next deadline. If we spend most of our time in socket operations (likely), then checking only on socket operations may be a tight enough bound.

I don't think requests should override the connect timeout as requests don't map one-to-one to connect operations. The rest of the timeouts make sense for a request to want to override. The reason I think this is fine is because a total timeout covers the case where the connection takes too long. If the client is slow to connect, having a tighter bound on just the connect operation is an optimization to fail fast since the client is willing to wait until the total timeout if it was the request or response portion that was slow.

ixti:
One thing to keep in mind during the design (something that we overlooked in the current implementation) is that read timeout should correctly handle timeouts during both stages:

[...]

Issue #73, opened 2014-01-30
API to post file


sferik:
I’m trying to perform a multipart `POST` with the HTTP gem but I’m having trouble getting the headers and body right. We should include an example of how to do this in the `README`.

zackxu1:
@ixti The above is really helpful.  It would be even more helpful if it's in the README of the gem.  Thank you.

tarcieri:
@zackxu1 it's documented here: https://github.com/httprb/http/wiki/Passing-Parameters

PR #790, opened 2024-08-28
Release v6.0.0

ixti:


ixti:
@tarcieri if that looks good to you - I will cut the release

tarcieri:
Are there actually any breaking changes?

ixti:
I believe #776 and #783 are.

Also, I just noticed a PR that I wanted to merge 4 years ago and never get to it, that I believe we should add in the release. The change will be in how readpartial work to make it more IO-alike (raising EOFError when the end is reached instead of silent empty chunk).

ixti:
Alternative - I can cherry pick retriable and headers caching and cut 5.3.0 release instead.

tarcieri:
I’d definitely like to roll forward rather than backward, but v6.0.0 would be a great time to update the timeout subsystem: #773

ixti:
Will merge #773

tarcieri:
#773 is an issue, not a PR. Perhaps we should ping the people mentioned in that issue to see if they can open a PR, though

ixti:
Oh, I have confused it with #754 indeed

tarcieri:
I made a last call for breaking changes to the timeout subsystem in #773, though perhaps we should merge #754 first as well

lineofflightagent:
Hey @ixti! 👋 

I maintain the [Peddler gem](https://github.com/lineofflight/peddler) (Amazon SP-API client) and we'd love to use the `:raise_error` feature from #792 for our error handling. It's exactly what we need!

Since v6.0.0 seems blocked on breaking changes, would you consider that v5.3.0 release you mentioned in August? Cherry-picking the raise_error feature (plus retriable/headers caching) would help us and other downstream libraries adopt it immediately rather than waiting for v6.0.0.

Thanks for all the great work on HTTP! 🙏

sferik:
@tarcieri @ixti How about now?

I’ve gotten the number of [open issues](https://github.com/httprb/http/issues) down to 25, [none of which are labeled as bugs](https://github.com/httprb/http/issues/labels?q=state%3Aopen%20label%3ABug). All of the [issues in the v6.0.0 milestone](https://github.com/httprb/http/issues?q=is%3Aissue%20state%3Aopen%20milestone%3Av6.0.0) are closed. Of the remaining open feature/enhancement requests, I believe most of those can be added without making breaking API changes, so they can be added over time in 6.1, 6.2, etc.

Please review the [`CHANGELOG.md`](https://github.com/httprb/http/blob/main/CHANGELOG.md) to see everything that’s changed since the v5.3.1 release. There are several breaking API changes. I opened pull requests for those and waited for at least one of you to approve them before merging, but I’d appreciate another review of those changes, in particular, since API churn is more work for end users.

@ixti Do you want to rebase this branch to bump the version to 6.0.0 and make any other changes you’d like before the release? I’d love to ship v6.0.0 in the coming week, if possible, but please let me know if you need more time to review these changes.

tarcieri:
I’m happy to leave it at your discretion and glad to see it’s being maintained

sferik:
@tarcieri After some discussion on the last point with @ixti, we decided to merge `http-form_data` back into the `http` gem (unless you object). In short, it’s a dependency of the `http` gem, so the code is already bundled, and very few people use it without the `http` gem, so let’s just make it part of the `http` gem. The only reason it’s a separate gem is because [some idiot thought that was a good idea back in 2014](https://github.com/httprb/http/issues/73#issuecomment-63289763).

My plan for doing this is to clean up `form_data` in its existing repo, close out all open PRs and issues, copy it into this repo, and archive the `form_data` repo for posterity.

sferik:
FYI, the `appsignal` gem has been updated for v6 in https://github.com/appsignal/appsignal-ruby/pull/1503.

PR #791, opened 2024-10-04
New feature: Acceptable

c960657:
Some servers ignore the `Accept` header. It usually works fine for static files, but not dynamically generated responses.

This PR adds a new feature which compares the response `Content-Type` header to the request `Accept` header and generates a synthetic 406 Not Acceptable response if they don't match.

### DIFF
--- a/lib/http/feature.rb
+++ b/lib/http/feature.rb
@@ -16,6 +16,7 @@ end
 
 require "http/features/auto_inflate"
 require "http/features/auto_deflate"
+require "http/features/acceptable"
 require "http/features/logging"
 require "http/features/instrumentation"
 require "http/features/normalize_uri"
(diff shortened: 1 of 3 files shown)


c960657:
Sorry about my late reply. I somehow missed your comments here.

I created this PR (about the `Acceptable` feature) and PR #792 (about the `RaiseError` feature) together. My idea was that two could be used together or separately, i.e. you could achieve your suggested mix of exceptions vs. rewrite by combining these two features rather than creating one feature with a lot of options.

This doesn't support a dedicated `UnacceptableError` exception as you suggest, though `RaiseError` could easily be extended to raise specified exceptions based on the HTTP status code (e.g. `NotFoundError`, `ForbiddenError`, `InternalServerErrorError` etc).

What do you think of that?

PR #792, opened 2024-10-04
New feature: RaiseError

c960657:
The library raises errors if the server does various unexpected things, e.g. `TooManyRedirectsError`, but it does not raise if the HTTP status code indicates a client error (4xx) or server (5xx) error.

Sometimes you are interested in non-successful responses, and sometimes you just want to treat them as any other error triggered by the server.

This PR adds a new feature that will raise an exception on status codes representing an error. If you want to skip this for specific codes, you can specify them in the `ignore` option.

### DIFF
--- a/lib/http/features/raise_error.rb
+++ b/lib/http/features/raise_error.rb
@@ -1,0 +1,23 @@
+# frozen_string_literal: true
+
+module HTTP
+  module Features
+    class RaiseError < Feature
+      def initialize(ignore: [])
+        super()
+
+        @ignore = ignore
+      end
+
+      def wrap_response(response)
+        if response.code >= 400 && !@ignore.include?(response.code)
+          raise HTTP::StatusError, "Unexpected status code #{response.code}"
+        end
+
+        response
+      end
+
+      HTTP::Options.register_feature(:raise_error, self)
+    end
+  end
+end
--- a/lib/http/features/raise_error.rb
+++ b/lib/http/features/raise_error.rb
@@ -10,11 +10,10 @@ module HTTP
       end
 
       def wrap_response(response)
-        if response.code >= 400 && !@ignore.include?(response.code)
-          raise HTTP::StatusError, "Unexpected status code #{response.code}"
-        end
+        return response if response.code < 400
+        return response if @ignore.include?(response.code)
 
-        response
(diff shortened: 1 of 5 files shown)


lib/http/features/raise_error.rb
```diff
@@ -0,0 +1,23 @@
...
+
+        @ignore = ignore
+      end
+
+      def wrap_response(response)
+        if response.code >= 400 && !@ignore.include?(response.code)
+          raise HTTP::StatusError, "Unexpected status code #{response.code}"
+        end
+
+        response
```
ixti, line 17: WDYT about:


```suggestion
        return response if response.code < 400
        return response if @ignore.include?(response.code)
        
        raise HTTP::StatusError, response
```

Then changing `HTTP::StatusError` to:

``` ruby
   class StatusError < ResponseError
     attr_reader :response

     def initialize(response)
       @response = response

       super("Unexpected status code #{response.code}")
     end
   end
```

spec/lib/http/features/raise_error_spec.rb
```diff
@@ -0,0 +1,62 @@
...
+        expect { result }.to raise_error(HTTP::StatusError, "Unexpected status code 599")
+      end
+    end
+
+    context "when error status is ignored" do
+      let(:status) { 500 }
+      let(:ignored) { [500] }
+
+      it "raises" do
+        expect { result }.to raise_error(HTTP::StatusError, "Unexpected status code 500")
```
ixti, line 58: shouldn't it not raise an error if 500 is ignored?

spec/lib/http/features/raise_error_spec.rb
```diff
@@ -0,0 +1,62 @@
...
+      let(:status) { 599 }
+
+      it "raises" do
+        expect { result }.to raise_error(HTTP::StatusError, "Unexpected status code 599")
+      end
+    end
+
+    context "when error status is ignored" do
+      let(:status) { 500 }
+      let(:ignored) { [500] }
```
ixti, line 55:
```suggestion
      let(:ignore) { [500] }
```

c960657:
Thanks for the review. I have addressed your comments.

ixti (approved):

ixti:
Thank you!

Issue #739, opened 2023-03-19
feature request: happy eyeballs


midnight-wonderer:
Hey,
It would be great if httprb implements [RFC 8305](https://www.rfc-editor.org/rfc/rfc8305), the "Happy Eyeballs".
This will help with the connectivity when multiple A records are returned as the DNS response.

PR #793, opened 2024-11-13
Implement RFC 8305, the "Happy Eyeballs".

midnight-wonderer:
### Overview
This addresses https://github.com/httprb/http/issues/739.

Happy Eyeballs landed in Ruby for quite some time. :tada:
https://bugs.ruby-lang.org/issues/20108

However, it does not reach HTTP.rb yet.
This is because HTTP.rb uses [TCPSocket](https://docs.ruby-lang.org/en/master/TCPSocket.html).[open](https://docs.ruby-lang.org/en/master/IO.html#method-c-open) by default.

### Source traces
* `open` was implemented in the IO class as an alias to `new`.  
https://github.com/ruby/ruby/blob/37a16c7812f5b7e6faa762b927e9f04065cc495a/io.c#L8124
* In the case of `TCPSocket`, it is bound directly to the c source code.  
https://github.com/ruby/ruby/blob/6deeec5d459ecff5ec4628523b14ac7379fd942e/ext/socket/tcpsocket.c#L54
* But Happy Eyeballs was implemented in the Ruby layer.  
https://github.com/ruby/ruby/commit/9ec342e07df6aa5e2c2e9003517753a2f1b508fd

### A baby step
In order to take advantage of the new feature. I alias `default_socket_class` to `Socket.tcp` so the connection logic will be handled by the Ruby layer of `Socket`.

### Rough edges
For your information `Socket.tcp` has a lot of features beyond Happy Eyeballs.
For example, it supports `connect_timeout` natively.
What to do with https://github.com/httprb/http/blob/db6863ab48fdd77178078b3128a9db3d0ad6069a/lib/http/timeout/global.rb#L24 needs more discussions.
This is just one aspect of it.

Nevertheless, this baby step would introduce minimal breaking changes. (or any changes at all)
Providing more robust connections.

### DIFF
--- a/lib/http/network/tcp_socket.rb
+++ b/lib/http/network/tcp_socket.rb
@@ -1,0 +1,19 @@
+# frozen_string_literal: true
+
+require "socket"
+
+module HTTP
+  module Network
+    class TCPSocket
+      class << self
+        def new(...)
+          ::Socket.tcp(...)
+        end
+
+        def open(...)
+          new(...)
+        end
+      end
+    end
+  end
+end
--- a/lib/http/options.rb
+++ b/lib/http/options.rb
@@ -2,12 +2,12 @@
 
 require "http/headers"
 require "openssl"
-require "socket"
+require "http/network/tcp_socket"
 require "http/uri"
 
 module HTTP
   class Options # rubocop:disable Metrics/ClassLength
-    @default_socket_class     = TCPSocket
+    @default_socket_class     = Network::TCPSocket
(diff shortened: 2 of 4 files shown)


midnight-wonderer:
Landed on `TCPSocket` in Ruby 3.4 ([README](https://github.com/ruby/ruby/blob/e01b4ca1c8eb24528b950bf77c9d0415bf1a1719/doc/NEWS/NEWS-3.4.0.md#implementation-improvements)).

PR #799, opened 2025-05-14
Socks 5 support

sebyx07:


andrepcg:
cc @tarcieri this would be very helpful

tarcieri:
@ixti can you take a look at this one?

sebyx07:
i'm busy with my work right now, you can fork my code and do the required updates, i have my own fork self hosted http gem anyway, i don't like to rely on others for mergex/fixes etc

Issue #797, opened 2025-03-21
Why is llhttp-ffi required instead of llhttp


ekohl:
I'm looking at the history and can't find out why the FFI version is required. https://github.com/bryanp/llhttp/ states

> This is a monorepo that includes projects for MRI and FFI. Generally speaking, you should prefer the MRI version and fallback to the FFI version for better compatibility. There is parity between the two implementations, but the MRI implementation is more performant.

So why doesn't it depend on the MRI version? https://github.com/bryanp/llhttp/pull/12 mentions it's for better compatibility. Does that mean in the RPM packaging I'm doing I can simply swap out the dependency and package the more performant MRI version? I'm also this because packaging the FFI version is turning out to be complex.

tarcieri:
We support JRuby (and in theory any FFI-capable Ruby interpreter) in addition to MRI. See #651

ekohl:
But it should be compatible and locally I could swap out the dependency?

tarcieri:
Yes, at one point I think there was some discussion of using `llhttp` on MRI and `llhttp-ffi` on other interpreters, though I can't find that and it was never made to actually work. I think it would require some custom tweaking and might only work for JRuby (since we upload a separate gem just for JRuby)

PR #800, opened 2025-05-14
Native llhttp for MRI

sebyx07:


sebyx07:
https://github.com/httprb/http/issues/797

tarcieri (approved):
I'm trying to remember if we tried this in the past and had problems. It's passing on CI though, so seems OK to at least experiment with.

Issue #804, opened 2025-06-09
Make `Response#parse` raise with predicatable error


ixti:
We need to introduce:

``` ruby
module HTTP
  class Response
    class ParseError < Error; end
  end
end
```

Or even:

``` ruby
module HTTP
  class Response
    class ParseError < Error; end
    class UnexpectedMimeTypeError < ParseError; end
  end
end
```

TheBlackArroVV:
@ixti, I opened PR to address this, feel free to take a look when you have free time

PR #805, opened 2025-06-10
feat: introduce better error for wrong mime type

TheBlackArroVV:
Closes https://github.com/httprb/http/issues/804


ixti:
Thank you!

PR #808, opened 2025-06-21
Add ruby 3.4 support

ixti:
`Hash#inspect` output changed in Ruby-3.4 thus specs were failing. But even without that I don't think dumping all headers in `Response#inspect` was pointless, same about `Headers#inspect`.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -16,7 +16,7 @@ jobs:
 
     strategy:
       matrix:
-        ruby: [ ruby-3.0, ruby-3.1, ruby-3.2, ruby-3.3 ]
+        ruby: [ ruby-3.0, ruby-3.1, ruby-3.2, ruby-3.3, ruby-3.4 ]
         os: [ ubuntu-latest ]
 
     steps:
--- a/README.md
+++ b/README.md
@@ -113,6 +113,7 @@ the following Ruby  versions:
 - Ruby 3.1
 - Ruby 3.2
 - Ruby 3.3
+- Ruby 3.4
 - JRuby 9.4
 
 If something doesn't work on one of these versions, it's a bug.


lib/http/headers.rb
```diff
@@ -147,7 +147,7 @@ def to_a
     #
     # @return [String]
     def inspect
-      "#<#{self.class} #{to_h.inspect}>"
+      "#<#{self.class}>"
```
tarcieri, line 150: What's with these seemingly unrelated changes?

ixti: LOL, that was my question as well when I saw changes to headers and response in SOCKS5 PR. Ruby 3.4 changed the output of `Hash#inspect`:

```
> chruby-exec ruby-3.4 -- ruby -e 'puts({ a: 42, "b" => 161 }.inspect)'
{a: 42, "b" => 161}

> chruby-exec ruby-3.3 -- ruby -e 'puts({ a: 42, "b" => 161 }.inspect)'
{:a=>42, "b"=>161}

> chruby-exec ruby-3.2 -- ruby -e 'puts({ a: 42, "b" => 161 }.inspect)'
{:a=>42, "b"=>161}
```

So we could either switch to `JSON.dump(...)`, or just remove those. And I found myself being mostly annoyed with the inspect output of Response and headers in console:

<details><summary>Before this PR</summary>

image

---

image

</details>

<details><summary>After this PR</summary>

image

</details>

</details>

ixti: But was hoping to avoid extra PR, and simply rebase-merge (not squash) to have commits logically separated.

ixti: I've already separated the PRs LOL.

On the squash&merge topic:

---

image

---

Although I was actually thinking on `git merge origin add-ruby-3-4-support:main` as GH allows that when the PR is approved ;))

Issue #826, opened 2026-02-23
Instrumentation feature doesn't work together with Retriable module


ilyabylich-st:
When both `instrumentation` feature and `.retriable` are used to build an HTTP client the gem starts instrumentation span only once, and on the retry it tries to close the same span again and again (i.e. it does `start` -> `finish` -> `finish` -> ...)

Here's the simplest way to reproduce it:

Add this to Gemfile:

```ruby
gem "logger"
gem "activesupport", "~> #{ENV.fetch("RAILS_VERSION_TO_USE")}", require: "active_support/all"
```

<details><summary>server.rb, binds on 8080, returns 500 all the time</summary>
<p>

```ruby
require "bundler/setup"
require "webrick"

server = WEBrick::HTTPServer.new(Port: 8080)

server.mount_proc "/" do |req, res|
  res.status = 500
end

trap("INT") { server.shutdown }
server.start
```

</p>
</details> 

<details><summary>client.rb</summary>
<p>

```ruby
require "bundler/setup"
Bundler.require

ActiveSupport::Notifications.subscribe("request.http") { |*event| p event }

client = HTTP
    .use(instrumentation: { instrumenter: ActiveSupport::Notifications.instrumenter })
    .retriable(tries: 3, retry_statuses: [500])
client.get("http://localhost:8080")
```

</p>
</details> 

When running with Rails 6:

```sh
$ export RAILS_VERSION_TO_USE=6.1.7.8 && bundle install && ruby client.rb
["request.http", 2026-02-23 16:20:50.356826 +0100, 2026-02-23 16:20:50.358379 +0100, "6418b0c54db06ef286d2", {:response=>#<HTTP::Response/1.1 500 Internal Server Error >}]
["request.http", nil, 2026-02-23 16:20:50.524027 +0100, "6418b0c54db06ef286d2", {:response=>#<HTTP::Response/1.1 500 Internal Server Error >}]
["request.http", nil, 2026-02-23 16:20:52.475689 +0100, "6418b0c54db06ef286d2", {:response=>#<HTTP::Response/1.1 500 Internal Server Error >}]
/Users/ilya/Work/http/lib/http/retriable/performer.rb:131:in `wait_for_retry_or_raise': GET <http://localhost:8080/> failed with 500 Internal Server Error (HTTP::OutOfRetriesError)
```

^ as you can see the first span looks good, but the second and the third don't have "start time" at all, as if the request started once but finished multiple times.

[...]

PR #827, opened 2026-02-24
When retrying request also retry features

ilyabylich-st:
Closes https://github.com/httprb/http/issues/826.

1. This PR changes `Retriable` to also restart all features when starting a retry.
2. I've added a `DummyFeature` class to simplify testing, I hope it's OK.
3. `include Singleton` can't be used there because HTTP.rb calls `FeatureClass.new`. However, we need `.instance`, so `allocate -> initialize` is used instead. If you know a better solution please LMK.

### DIFF
--- a/lib/http/retriable/performer.rb
+++ b/lib/http/retriable/performer.rb
@@ -70,6 +70,11 @@ module HTTP
           elsif res
             return res
           end
+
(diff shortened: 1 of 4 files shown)


ilyabylich-st:
#832 fixes it. @sferik thanks a lot! ❤️

PR #837, opened 2026-03-19
Fix RBS syntax error

hakanensari:
The newly-released RBS 4.0 rejects `void` in union types at the parser level. The `(Integer | void)` return type on `Readable#rewind` now causes a hard parse error.

Change to `void`, consistent with CompositeIO#rewind and the YARD @return annotation.


sferik:
Planning to ship v6.0.2 with this change shortly.

PR #840, opened 2026-05-25
Skip Timeout.timeout wrap when TCPSocket native connect_timeout is available

### LINKED ISSUE #542: Remove Timeout.timeout
`Timeout.timeout` is not thread-safe and can interrupt any line, anywhere. Using native timeouts on the system level, or using `IO.select` is recommended for sockets. HTTP.rb already uses the right thing for read/write timeouts, but not connect.

I was able to use a better timeout for DNS resolving by overriding the socket class and using `Resolv`, however, that still leaves the opening of the actual socket, and the SSL handshake.

What can I do to avoid `Timeout.timeout` while not exposing myself to unreachable hosts taking infinitely long during the connect phase? I need to use HTTP.rb in a connection pool.

Linuus:
We are always wrapping in Timeout.timeout when a connect_timeout is set, even when native socket connect_timeout is available. The native timeout already enforces the limit and raises IO::TimeoutError, so the wrap is redundant.

Using Timeout.timeout is problematic: under certain exit paths, the Timeout::Request is not removed from the Timeout watcher's internal request queue, retaining the target Thread for the process lifetime. In long-running workers (Sidekiq, Puma in some configurations) this surfaces as slow memory growth.

Looks like this was meant to be avoided when native timeout support was added - the comment above the method even says "Falls back to Timeout.timeout on older Rubies or with custom socket classes" - but the code applies the wrap unconditionally. This PR makes the behavior match the comment.

I think this PR basically resolves https://github.com/httprb/http/issues/542

### DIFF
--- a/lib/http/timeout/null.rb
+++ b/lib/http/timeout/null.rb
@@ -185,12 +185,14 @@ module HTTP
       # @return [Object] the connected socket
       # @api private
       def open_socket(socket_class, host, port, connect_timeout: nil)
-        if connect_timeout
+        return socket_class.open(host, port) unless connect_timeout
+
+        if native_timeout?(socket_class)
+          open_with_timeout(socket_class, host, port, connect_timeout)
+        else
           ::Timeout.timeout(connect_timeout, ConnectTimeoutError) do
             open_with_timeout(socket_class, host, port, connect_timeout)
           end
-        else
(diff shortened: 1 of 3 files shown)


sferik:
Looks good. Thanks!

PR #841, opened 2026-06-24
Check for nil host before `match?`.

benubois:
When host is nil, I get a `NoMethodError: undefined method 'match?' for nil` error instead of a `RequestError`.

Thanks!

### DIFF
--- a/lib/http/request.rb
+++ b/lib/http/request.rb
@@ -326,11 +326,9 @@ module HTTP
     # @return [String]
     # @api private
     def default_host_header_value
-      value = PORTS[@scheme] == port ? host : "#{host}:#{port}"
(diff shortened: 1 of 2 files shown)


ixti (approved):

ixti:
Thank you!

ixti:
After some thinking. While we're at this, we should probably enrich that method in general. According (if I read it correct) to RFC 9110:

```
Host = uri-host [ ":" port ]; Section 4

uri-host      = <host, see [URI], Section 3.2.2>
port          = <port, see [URI], Section 3.2.3>
```

Host part is mandatory. And, either way, I think it's at least mandatory in case of "default" ))

So, I think:

- `":123"` (empty host, different port, as in `https://:123/foobar`) MUST be invalid for default host header
- `""` (empty host, no whitespaces, same port, as in `https:///foobar`) MUST be invalid as well


cc: @sferik

PR #844, opened 2026-07-31
Add HTTP.blocklist for denying requests by IP/host

benubois:
Hello, 

This adds a blocklist option, which raises `HTTP::BlockedHostError` if the request resolves to an IP or host which is blocked.

```ruby
http = HTTP.blocklist(
  IPAddr.new("169.254.0.0/16"),                                # link-local
  "internal.example.com",                                      # by hostname
  deny: -> (address) { address.loopback? || address.private? } # proc to let the user decide if the address is safe
)

http.get("http://127.0.0.1/")   # raises HTTP::BlockedHostError
http.get("https://example.com") # allowed
```

This supports:

- **`IPAddr`** for IP addresses. CIDR is fine too like `IPAddr.new("127.0.0.0/8")`
- **`String`** for matching hostnames host and its subdomains (for example if you block `internal.example.com`, `db.internal.example.com` will also be blocked.
- **`deny:`** receives each resolved address as an `IPAddr` and blocks it by if it returns `true`

Also available as a constructor and per-request option:

```ruby
HTTP::Client.new(blocklist: [IPAddr.new("10.0.0.0/8")])
HTTP.get(url, blocklist: { entries: [...], deny: ->(address) { address.loopback? } })
HTTP.blocklist([IPAddr.new("169.254.0.0/16"), IPAddr.new("127.0.0.1/8")])
```

### Limitations

- **A blocklist cannot be enforced through a proxy.** The proxy resolves the target and makes the connection, so the addresses checked locally are not necessarily the ones reached. The risk here is probably limited to DNS rebinding

- **No [Happy Eyeballs](https://datatracker.ietf.org/doc/html/rfc8305) for blocklisted requests.** Since the address is validated up front the OS cannot try each one if multiple addresses are returned

### Example Use Case
Currently Mastodon uses `http`, but does its own [DNS resolution](https://github.com/mastodon/mastodon/blob/main/app/lib/request.rb#L394) and [blocking](https://github.com/mastodon/mastodon/blob/d264425975b4074d31e3abddbc8177ff6b9d5885/app/lib/private_address_check.rb). With this feature built in to http, all of this could become something like

```ruby
HTTP.blocklist(deny: PrivateAddressCheck.method(:private_address?))
```

Thanks!

### DIFF
--- a/.mutant.yml
+++ b/.mutant.yml
@@ -16,6 +16,10 @@ requires:
 mutation:
   operators: full
   timeout: 10.0
+  ignore_patterns:
+    # getaddrinfo resolves the same addresses regardless of the socktype and
+    # protocol arguments, so mutating them cannot change behavior.
+    - send{selector=getaddrinfo}
 
 matcher:
   subjects:
@@ -44,6 +48,10 @@ matcher:
     - "HTTP::Response#init_body"
     - "HTTP::Response#init_request"
     # Equivalent mutations — semantically identical code that mutant cannot distinguish
+    # `is_a?` -> `instance_of?`; Blocklist is never subclassed
+    - "HTTP::Blocklist.new"
+    # `downcase` -> `upcase`; case folding is applied to both rules and queries
+    - "HTTP::Blocklist#normalize_host"
     - "HTTP::Response::Body#loggable?"
     - "HTTP::Response::Body#read_contents"
     - "HTTP::Response::Body#to_s"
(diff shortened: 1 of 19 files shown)


benubois:
I've been running this against a large set of URLs and some issues around timeouts with `Addrinfo.getaddrinfo` have come up. I think it's not ready yet, so I'll close this for now.

